PR #198, opened 2013-05-29
Add subprotocol support

### LINKED ISSUE #29: Implement proper sub-protocol support
The protocol header can include multiple sub-protocol values in the request. In order to properly support sub-protocols, there needs to be some way for the server to select the protocol from the list that it supports (because the response needs to have only one).

This may require a separate callback handler so that the server using 'ws' has the opportunity to select the one it wants to use before the connection is fully established (i.e. on-connection is too).

### LINKED ISSUE #192: WS client: WS subprotocol not processed
When a client connects

```
   var ws = new WebSocket('ws://127.0.0.1:9000', {protocol: ["baz", "foo"]});
```

and the server is willing to speak `baz` (hence sending `Sec-WebSocket-Protocol: baz`), the WebSocket subprotocol selected by the server (from the list supplied by the client) should be available

```
   ws.onopen = function () {
     console.log(ws.protocol);
   }
```

However, it's not there ..

### LINKED ISSUE #193: WS server: incorrect WS subprotocol handling
When a client announces WebSocket subprotocols, the server application should be
- able to select 1 of the client announced subprotocols (if any), and
- set the appropriate HTTP header

However, `ws` currently just sends backs the header as received from the client: https://github.com/einaros/ws/blob/master/lib/WebSocketServer.js#L201

I.e. here is a response:

```
    HTTP/1.1 101 Switching Protocols
    Upgrade: websocket
    Connection: Upgrade
    Sec-WebSocket-Accept: RGs9E/xPwK3Nm00vbK9lpkHLzEs=
    Sec-WebSocket-Protocol: wamp,foobar
```

This is wrong, since the server must select and send _1_ subprotocol.

And as far as I can see the application has no means of controlling which (if any) subprotocol is selected. (I have looked at `verifyClient`, but the info provided to that callback doesn't seem to contain the subprotocols announced by the client).

kanaka:
- The client will set ws.protocol based on the subprotocol returned by
  the server. Fixes #192.
- The server application can select from multiple client subprotocol
  options by specifying a handleProtocols in initialization options.
  This function takes two arguments; the first is an array of
  subprotocols sent by the client (empty array if no subprotocol
  specified by the client), the second is a callback function. The
  callback function takes two arguments; the first indicates if the
  client subprotocols are valid, the second argument (only passed when
  first is true) is the subprotocol string selected to return to the
  client. Fixes #29. Fixes #193.
- The default subprotocol behavior of the server is to select/return
  the first subprotocol in the list.
- Includes 6 tests to test subprotocol support:
  - test default behavior of selecting first subprotocol
  - test custom handleProtocols that selects a subprotocol
  - detect server sending invalid subprotocol
  - detect server sent no subprotocol even though client sent
    subprotocol header.
  - test custom handleProtocols that rejects connection based on
    invalid client subprotocol request.
  - server responds with 501 to client if the handleProtocols
    handler neither selects a protocol nor rejects the connection.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -541,6 +541,25 @@ function initAsClient(address, options) {
       return;
     }
 
+    var serverProt = res.headers['sec-websocket-protocol'];
+    var protList = (options.value.protocol || "").split(/, */); 
+    var protError = null;
+    if (!options.value.protocol && serverProt) {
+        protError = 'server sent a subprotocol even though none requested';
+    } else if (options.value.protocol && !serverProt) {
+        protError = 'server sent no subprotocol even though requested';
+    } else if (serverProt && protList.indexOf(serverProt) === -1) {
+        protError = 'server responded with an invalid protocol';
+    }
+    if (protError) {
+        self.emit('error', protError);
+        removeAllListeners(self);
+        socket.end();
+        return;
+    } else if (serverProt) {
+        self.protocol = serverProt;
+    }
+
     establishConnection.call(self, Receiver, Sender, socket, upgradeHead);
 
     // perform cleanup on http resources
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -24,6 +24,7 @@ function WebSocketServer(options, callback) {
     port: null,
     server: null,
     verifyClient: null,
+    handleProtocols: null,
     path: null,
     noServer: false,
     disableHixie: false,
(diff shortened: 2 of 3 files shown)


Apoorva2405:
@kanaka @einaros Are these changes still present in websockets?
I am trying to use handleProtocols as below:

var wss = new WebSocketServer({
    port: wsPort, handleProtocols: function (protocol,cb) {
        if (protocol[protocol.indexOf("wamp")]) {
            protocolToSet = protocol[protocol.indexOf("wamp")];
           cb(true, protocolToSet);
           return protocolToSet;
        }
          cb(false);
    }
});

But I am getting cb is not defined error in my server code

PR #208, opened 2013-06-25
use wss as location scheme if given as x-forwarded-proto; support 'headers' event in handleHixieUpgrade

machenmusik:
... https


lpinca:
Closing this as we are going to drop support for older versions of the protocol.
Thanks.

PR #209, opened 2013-06-26
fix a request that hangs around after an unexpected response

aktau:
In my setup, sometimes my node websocket server goes away. At that point nginx tells any further clients that there is a Bad Gateway (502). Of course the WebSocket client then errors out, but it forgets to abort the request. For websocket connections, nginx must be configured with a keep-alive value, so the connection can take a long time to time out. This ties up unnecesary resources.


lpinca:
Closing this as it has been addressed by #271.
Thank you!

PR #226, opened 2013-08-23
Server sending broadcast example

stefanocudini:
https://github.com/einaros/ws/issues/216

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,18 @@ wss.on('connection', function(ws) {
 });
 ```
 
+### Server sending broadcast data ###
+
+```js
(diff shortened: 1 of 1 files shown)


zwz:
@einaros 
It is a good example. But what if the clients array is very very long?
As I am using ws for my app with nodejs, I am wondering if this may be an issue:
when the app tries to broadcast to so many clients that it just blocked there, since nodejs is single-threaded. 

If this would be an issue, what  should I do?
It seems that socket.io has a built-in broadcast. How does it address the issue?
Any hints?

Issue #241, opened 2013-09-14
Server upgrade and error listeners not being removed on close


D1plo1d:
I'm attempting to set up a express 3 server that creates and removes websocket servers regularly (ie different routes, not just not client sessions) and I've noticed that when a websocket server is recreated at a URL that was previously used by another websocket (that has since been closed), that previous websocket's events still seem to fire.
### Steps to reproduce:
1. Create a websocket server on express 3
2. Close the websocket server
3. Create a second websocket server on the same server at the same url
4. connection events will now be thrown by both the old and new web socket servers  whenever a client connects

I console.logged my server after the first websocket server was started and after the second and found that in it's _events some listeners were being duplicated.
### After the first ws is opened:

```
listening: { [Function: g] listener: [Function] },
error: [Function],
upgrade: [Function]
```
### After the first ws is closed and the second is opened:

```
listening: [ [Object], [Object] ],
error: [ [Function], [Function] ],
upgrade: [ [Function], [Function]
```
### This is the function being duplicated in upgrade:

I can also post the listening and error fns as well if this is difficult to reproduce

(I ran `wss.options.server._events.upgrade.toString()`)

```
function (req, socket, upgradeHead) {
      //copy upgradeHead to avoid retention of large slab buffers used in node core
      var head = new Buffer(upgradeHead.length);
      upgradeHead.copy(head);

      self.handleUpgrade(req, socket, head, function(client) {
        self.emit('connection'+req.url, client);
        self.emit('connection', client);
      });
}
```

lpinca:
Closing this as I think this issue has been "fixed" by #885. Please comment or reopen if needed.

PR #253, opened 2013-10-13
Cleaning up server event listeners in WebSocketServer close

D1plo1d:
This was causing subsequent websockets on the same url to fail to upgrade properly for me.

There is still another, unrelated issue related to repeatedly creating and destroying websockets on the same URL. Because this could get confusing quickly I'm going to note in big bold letters here: **This Patch Does Not Fix Issue #241**.

Running the included patched test against the current eianros:master should fail whereas running it against this branch should succeed.

In the immortal words of Douglas Adams: "Share and Enjoy"


lpinca:
Closing this as it seems that this change is no longer necessary. Please comment or reopen if needed.

Thank you.

PR #287, opened 2014-02-05
Do not override the "fin" property of the send options

codingphil:
Setting the "fin" porperty to false is needed to send fragmented messages to the client.


lpinca:
@codingphil can you please rebase against master and add a test as suggested by @3rd-Eden?
Thanks.

lpinca:
Merged as 01ab4f7dc0b32b18926cc94502d81906cd70c777.
Thank you!

PR #296, opened 2014-03-18
Update WebSocket.js

Mattijah:
On the line 470 will be thrown an error if `protocolVersion` is not `8` or `13`

`if (options.value.protocolVersion !== 8 && options.value.protocolVersion !== 13) {
    throw new Error('unsupported protocol version');
}`

Below on the line 487 is `this.supports.binary = (this.protocolVersion !== 'hixie-76');` which is completely pointless since `protocolVersion` can be only `8` or `13` anyway.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -484,7 +484,7 @@ function initAsClient(address, protocols, options) {
   this._isServer = false;
   this.url = address;
   this.protocolVersion = options.value.protocolVersion;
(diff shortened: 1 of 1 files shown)


lpinca:
It makes sense. @Mattijah can you please rebase against master?

lpinca:
Merged as e5a2ab4a9e2921597a835c244b7b284f6e269926.
Thank you!

PR #299, opened 2014-03-20
Ability to send a number

Mattijah:
Ability to send a number (Converted to string) otherwise is buffer going to be only allocated and sent as empty OR if is passed bigger number then buffer will throw an exception.


lpinca:
@Mattijah I'm closing this pr as the suggested change is already implemented now. See https://github.com/websockets/ws/blob/0cfa5cc94d50bebd6f85b78ec1094dd9a7a17d62/lib/Sender.js#L182.

Thank you.

PR #308, opened 2014-04-23
wscat: Support sending control frames (ping, pong, close)

simonratner:
Commands are prefixed with a slash "/" character.  The remainder of the
input line is used as command parameters. Also, print out ping and pong
control frames when they are received.

Example:

```
$ ./bin/wscat -c ws://echo.websocket.org
connected (press CTRL+C to quit)
> hi there
< hi there
> /ping
pong
> /ping echo
pong echo
> /close 1001 bye
disconnected
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,10 @@ ws.on('message', function(data, flags) {
     < hi there
     > are you a happy parrot?
     < are you a happy parrot?
+    > /ping echo
+    pong echo
+    > /close
+    disconnected
 
 ### Other examples ###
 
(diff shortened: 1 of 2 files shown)


lpinca:
`wscat` has moved to https://github.com/websockets/wscat. Can you please resubmit there?
Thank you.

simonratner:
It has been a while. I no longer use this library, nor do I have an OSX machine handy to verify that the patch still works.

Feel free to take this code verbatim, or modify and merge as needed, if you think this functionality should be in https://github.com/websockets/wscat.

lpinca:
cc: @nwwells who is helping maintaining `wscat`.

nwwells:
We'll track this on websockets/wscat#32. @lpinca you can close this if you want, and I'll just reference the commits from there.

PR #336, opened 2014-07-22
added small fix to prevent server from crashing on empty this.state

MosNicholas:


petkaantonov:
This bug is affecting us as well, landing this is appreciated

PR #368, opened 2014-10-07
Add code to cleanup any pending http requests when a socket is closed

DamonOehlman:
I have a case where I am testing two url endpoints sequentially (e.g. `ws://test.com/a` and `ws://test.com/b`).  In the event that I haven't received a response from the first endpoint (after a short period of time), I close that websocket connection and proceed to connect to the second endpoint.

This is being done in a client library to handle an change of websocket endpoint for it's server-side counterpart.

I felt that the logic was fairly sound in this, but noticed that when running in a node environment the process was not being cleaned up due to the open HTTP request.  This patch addresses that particular issue, albeit without tests.

Let me know where you'd like me to take it from here for a possible merge :)

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -639,7 +639,8 @@ function initAsClient(address, protocols, options) {
 }
 
 function cleanupRequest() {
-  if (! this._req) return;
+  var sock = this._req && this._req.socket;
+  if (! sock) return;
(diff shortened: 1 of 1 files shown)


lpinca:
@DamonOehlman a lot of time has passed, is this still needed?

PR #379, opened 2014-10-17
Fix documentation, don't emit 'connection' if the socket is closed and improve path behaviour

ibc:
* Fix documentation: `http.ClientRequest` => `http.IncomingMessage` in `ws.Server.options.verifyClient`.
* Documentation: document `ws.WebSocket.upgradeReq`. 
* Don't emit 'connection' event if the socket was closed before the user calls `cb(true)` in the `verifyClient()` callback (note that the user may choose to call `cb(true)` after an asynchronous operation).


sushantdhiman:
+1

PR #409, opened 2014-12-03
Support permessage-deflate extension

nkzawa:
This PR adds [the permessage-deflate extension](https://tools.ietf.org/html/draft-ietf-hybi-permessage-compression-19) support to the module including the following features:
- Extension parameters (widow bits, no context takeover for both client and server)
- Configurations of extension parameters and `memLevel` (zlib)
- Enable extension by default
- All test cases of [Autobahn Testsuite](http://autobahn.ws/testsuite/) pass including WebSocket Compression

Things not included in the PR:
- `x-webkit-frame-deflate` (which Safari supports)
- Comprehensive architecture to add extensions flexibly

If these features also are required, I'm ready to add them.

Related: https://github.com/einaros/ws/issues/34, https://github.com/Automattic/socket.io/issues/1148

### DIFF
--- a/lib/PerMessageDeflate.js
+++ b/lib/PerMessageDeflate.js
@@ -262,7 +262,7 @@ PerMessageDeflate.prototype.compress = function (data, fin, callback) {
     cleanup();
     var data = Buffer.concat(buffers);
     if (fin) {
-      data = data.slice(0, -4);
+      data = data.slice(0, data.length - 4);
     }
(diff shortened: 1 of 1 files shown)


dscherger:
@nkzawa Is there any intention or plan to add support for `x-webkit-deflate-frame` even though it's deprecated, it would be nice for this to work with safari on ios and osx.

PR #427, opened 2015-01-04
Fix error when compress option is true and perMessageDeflate is disabled

nkzawa:
Sorry, I found a bug which happens when setting the `compress` option `true` explicitly.

```js
ws.send(data, { compress: true });

Uncaught TypeError: Cannot call method 'compress' of undefined
```

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -238,7 +238,10 @@ WebSocket.prototype.send = function send(data, options, cb) {
   }
 
   if (typeof options.mask === 'undefined') options.mask = !this._isServer;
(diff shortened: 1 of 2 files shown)


3rd-Eden:
Thanks for the fix and the test.

nkzawa:
thank you for your quick response :)

nkzawa:
@3rd-Eden it seems v0.6.5 is not released to npm. Could you release it please?

3rd-Eden:
Argh, sorry about that. It seems I only did a `git push` and not an `npm publish`. Published 552dddaf9bcc5304c22415b81aa748384d82837c as 0.6.5

nkzawa:
@3rd-Eden thx!

PR #437, opened 2015-01-18
Make it work on nodejs v0.8 too

pjump:
Do use 'extern "C"' for the module's init function if nodejs version is less than 0.10.0.
This fixes the travis test for nodejs v0.8.

###########################
Apologies for breaking the travis-ci test for nodejs v0.8 with my last pull request.
This one conditionally keeps the 'extern "C"' for older node releases.
Tests for both versions are green https://travis-ci.org/pjump/ws

### DIFF
--- a/src/bufferutil.cc
+++ b/src/bufferutil.cc
@@ -6,6 +6,7 @@
 
 #include <v8.h>
 #include <node.h>
+#include <node_version.h>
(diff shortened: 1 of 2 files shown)


3rd-Eden:
That's odd, Travis-Ci said all was fine =/

PR #447, opened 2015-02-05
Fix for cleanupWebsocketResources using the wrong 'this'

fillrate:
When the WebSocket's HTTP request gets either an error or a response event, the cleanupWebsocketResources is called on this, which is the HTTP request, rather than self which prevents a clean close.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -665,7 +665,7 @@ function initAsClient(address, protocols, options) {
 
   req.on('error', function onerror(error) {
     self.emit('error', error);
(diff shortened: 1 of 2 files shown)


3rd-Eden:
Is there any way you can write a test against this as obviously all our tests passed making it believe that the old way was the correct way.

fillrate:
I added the test 'can handle error before request is upgraded', which attempts to connect to a non-existent local server, then ensures that the socket is cleaned up (readyState is set to CLOSED)

3rd-Eden:

PR #453, opened 2015-02-17
update WebSocket client docs

zemirco:
sync docs with code - https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L46


3rd-Eden:
Thanks for the pull <3

PR #468, opened 2015-03-11
Throw TypeError when class constructor called without new

LinusU:
I had the strangest bug where it would throw an error instead of emitting `connected`. Turns out that I wasn't using the `new` keyword when creating my `WebSocketServer`. This pull requests throws an error when creating the instance as to alert the user of the misstake as soon as possible. It follows closely how it's done in [ES6 (9.2.2 - [[call]])](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-ecmascript-function-objects-call-thisargument-argumentslist) with the new `class` keyword.

I have added new tests and all test passes.

### DIFF
--- a/lib/BufferPool.js
+++ b/lib/BufferPool.js
@@ -7,6 +7,10 @@
 var util = require('util');
 
 function BufferPool(initialSize, growStrategy, shrinkStrategy) {
+  if (this instanceof BufferPool === false) {
+    throw new TypeError("Classes can't be function-called");
+  }
+
   if (typeof initialSize === 'function') {
     shrinkStrategy = growStrategy;
     growStrategy = initialSize;
--- a/lib/PerMessageDeflate.js
+++ b/lib/PerMessageDeflate.js
@@ -12,6 +12,10 @@ PerMessageDeflate.extensionName = 'permessage-deflate';
  */
 
 function PerMessageDeflate(options, isServer) {
+  if (this instanceof PerMessageDeflate === false) {
+    throw new TypeError("Classes can't be function-called");
+  }
+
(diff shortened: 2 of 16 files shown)


3rd-Eden:
Not really a fan of throwing errors for this. I think it makes more sense to just return a new instance. E.g.

```js
function BufferPool(initialSize, growStrategy, shrinkStrategy) {
if (!(this instanceof BufferPool) {
   return new BufferPool(initialSize, growStrategy, shrinkStrategy);
}
```

LinusU:
I've been a bit back and forth myself on that matter but after the ES6 spec throws error for their classes I've become set on that approach.

The problem with not requiring `new` is that it there is no way for the user to know if it will work or not, unless it's specifically mentioned in the docs.

From the users (of this library) perspective, omitting the `new` keyword is a bad pattern since it only works when the library architect has specifically engineered it in. In the cases it dosen't work, it often produces really weird bugs that's not always apparent what caused it. E.g. in my case it was `undefined is not a function` at the line `self.emit('connected');`.

Just my 2 cents.

LinusU:
This is also how the `WebSocket` class in the browsers works (tested in Safari and Chrome).

It is also possible to change from requiring it to not at any point in the future, but going the other way breaks compatibility. If we allowed usage without `new` we couldn't rewrite this using the ES6 `class` keyword in the future...

LinusU:
@3rd-Eden Any comments? :)

3rd-Eden:
Thanks for the correction. Will be in the next minor release.

LinusU:
Thank you :raised_hands:

PR #479, opened 2015-04-03
Correctly switch the socket data handler

lpinca:
I think that the `realHandler` function was designed to push data efficiently to the receiver after the first chunk of data, but it seems that it never worked as intended.

If you look at the commit where it was intruduced (1faa9525) you'll see that the socket data handler is set to be `dataHandler` which points to `firstHandler` when the listener is registered. When `firstHandler` is invoked the value of `dataHandler` is changed to `realHandler`, but this doesn't replace the listener which will keep to be `firstHandler`.

With this patch the listener is correclty switched.

P.S. In #212 `socket.on('data', dataHandler);` was moved at the end of the function to ensure that the `readyState` was set to `WebSocket.OPEN` before receiving any data, but this is not needed because a net stream data event does not fire in the same tick.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -744,13 +744,15 @@ function initAsClient(address, protocols, options) {
 }
 
 function establishConnection(ReceiverClass, SenderClass, socket, upgradeHead) {
-  var ultron = this._ultron = new Ultron(socket);
-  this._socket = socket;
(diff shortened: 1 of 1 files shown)


lpinca:
I just noticed that there is another [pull request](https://github.com/websockets/ws/pull/325) for this.

3rd-Eden:
Thanks for your pull request!

Issue #472, opened 2015-03-20
Attempting to access ws server using browser gives status code 200, not an error


deltaflux:
When creating a ws server by specifying a port, then a default handler is used which returns the message 'Not implemented' and a status code of 200 OK. Wouldn't it make more sense to return 404?

See WebSocketServer.js line 46 for the function I'm talking about.

3rd-Eden:
Yes, it might make more sense to do so.

PR #480, opened 2015-04-04
Return 426 for non-Upgrade requests

lpinca:
This patch will fix the status coded returned for non-Upgrade requests, changing it from `200` to `426`.
See #472.


3rd-Eden:
Thanks for your pull request!

PR #481, opened 2015-04-09
Enable to use regexp instead string path

yarax:
Hi!
I had to listen different ws paths in my code by mask, so I think it would be useful for others

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -141,7 +141,7 @@ WebSocketServer.prototype.handleUpgrade = function(req, socket, upgradeHead, cb)
   if (this.options.path) {
     var u = url.parse(req.url);
     if (this.options.path instanceof RegExp) {
-      if (u && !u.pathname.match(this.options.path)) return;
+      if (u && u.pathname && !u.pathname.match(this.options.path)) return;
     } else {
       if (u && u.pathname !== this.options.path) return;
     }
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -140,7 +140,12 @@ WebSocketServer.prototype.handleUpgrade = function(req, socket, upgradeHead, cb)
   // check for wrong path
   if (this.options.path) {
     var u = url.parse(req.url);
-    if (u && u.pathname !== this.options.path) return;
+    if (this.options.path instanceof RegExp) {
+      if (u && !u.pathname.match(this.options.path)) return;
+    } else {
+      if (u && u.pathname !== this.options.path) return;
+    }
+
   }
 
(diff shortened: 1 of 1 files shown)


nkzawa:
I'm curious when you need this.

3rd-Eden:
Also, pull requests without tests are accepted.

yarax:
We balance ws connections among servers, using nginx and key in path (/ws/room1 should go to server1, /ws/room2 to server2)  and in code I just want to know, that it it something like /ws/* comes

If you approve feature I'll make tests

yarax:
Another case I'd like to pass some data on handshake in path
And the another reason why I ask pull request is no ability to patch `handleUpgrade` method with my own one, because of using global functions in module

sequoiar:
@yarax it should be 

-    if (u && u.pathname !== this.options.path) return;
+    if (this.options.path instanceof RegExp) {
+      if (u && u.pathname && !u.pathname.match(this.options.path)) return;
+    } else {
+      if (u && u.pathname !== this.options.path) return;
+    }
+
   }

yarax:
@sequoiar sure, thank you

nkzawa:
Can't we use query parameters for that? I think it's more appropriate than using path.

Btw, it seems `path` option is used in other places too.

https://github.com/websockets/ws/blob/master/lib/WebSocketServer.js#L57

yarax:
@nkzawa nginx `location` directive doesn't match query string at all. So it's impossible.
And the another point is that I shouldn't synchronise balancing logic and code. I just should know that the coming request is suitable for the certain server. But coming url path may occur different

michaelmcmillan:
> Also, pull requests without tests are accepted. Especially when they introduce new functionality.

Did you mean "unaccepted"?

auchenberg:
+1 for this FR.

pukhalski:
+1 for this PR.

lib/WebSocketServer.js
```diff
@@ -140,7 +140,12 @@ WebSocketServer.prototype.handleUpgrade = function(req, socket, upgradeHead, cb)
   // check for wrong path
   if (this.options.path) {
     var u = url.parse(req.url);
-    if (u && u.pathname !== this.options.path) return;
+    if (this.options.path instanceof RegExp) {
+      if (u && u.pathname && !u.pathname.match(this.options.path)) return;
```
lpinca: Can you use `RegExp.prototype.test()` instead of `String.prototype.match()`?

lpinca:
@yarax can you add a test?
Thank you.

lpinca:
@yarax sorry I've seen #483 only now. That may be a better way to handle this.

lpinca:
Closing in favor of #483.

PR #483, opened 2015-04-14
Expose logic for what a WSS should handle

yaworsw:
I thought it might be a good idea to expose the logic that checks whether a WSS should handle a request.  This would allow developers to override this logic if they **really** want to.

Imo this would effectively solve #481 without any downsides.

### DIFF
--- a/test/WebSocketServer.test.js
+++ b/test/WebSocketServer.test.js
@@ -293,6 +293,22 @@ describe('WebSocketServer', function() {
     });
   });
 
+  describe('#shouldHandle', function() {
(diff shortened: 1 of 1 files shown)


headlessme:
+1

lpinca:
Can you please rebase against master?
Thank you.

lpinca:
Merged as ebf86b58f1fb1e96ab96ebf95782bad7bc5ab8ca...d83117d5fbaf2ece7e4ffe575efd486469abc29d.

Thanks!

Issue #452, opened 2015-02-16
Browserify does not work


zemirco:
My `server.js`:

```js
var WebSocketServer = require('ws').Server;
var wss = new WebSocketServer({
  port: 8080
});

wss.on('connection', function connection(ws) {

  ws.on('message', function incoming(message) {
    console.log(Date.now() + ': ' + message);
    ws.send('pong');
  });

});
```

My `client.js`:

```js
var WebSocket = require('ws');
var ws = new WebSocket('ws://localhost:8080');

// init connection from client
ws.on('open', function open() {
  ws.send('ping');
});

ws.on('message', function(data) {
  console.log(Date.now() + ': ' + data);
  setTimeout(function() {
    ws.send('ping');
  }, 1000);
});
```

Running `$ node server.js` and `$ node client.js` in two separate terminals works fine as expected. 

I'd like to use my `client.js` in a browser and create a `bundle.js` with browserify:

`$ ./node_modules/.bin/browserify client.js > bundle.js`

Here is my super simple `index.html`:

```html
<!doctype html>
<html>
  <head>
      <meta charset="utf-8">
      <meta http-equiv="X-UA-Compatible" content="IE=edge">
  </head>
  <body>
    <p>Hello world!</p>
    <script src="bundle.js"></script>
  </body>
</html>
```

Unfortunately this does not work
> TypeError: ws.on is not a function (bundle.js:7)

I think the problem is that ws uses `websocket.on('message', callback)` and browsers use `websocket.onmessage = function() {}`. Can I somehow use my `client.js` code with browserify in a browser? Thanks a lot :)

zemirco:
Thanks @3rd-Eden - 

<blockquote class="twitter-tweet" lang="en"><p><a href="https://twitter.com/zemirco">@zemirco</a> use addEventListener instead? Also on the serverside we also support onmessage etc</p>&mdash; Arnout Kazemier (@3rdEden) <a href="https://twitter.com/3rdEden/status/567340446129864704">February 16, 2015</a></blockquote>
<script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

I'm going to change my code and use `websocket.onmessage` or `addEventListener` inside my `client.js`.

maxleiko:
@zemirco 

You can also use a trick in your browser to ensure that browserified module that uses `ws` with the `EventEmitter`-like API will work:

```js
// this is an ugly hack to prevent browseryfied github.com/einaros/ws module to throw errors at runtime
[...]

PR #493, opened 2015-05-06
fix https://github.com/websockets/ws/issues/452

maxleiko:
This monkey-ish patch will allow the use of `.on` in browser environment when you use `einaros/ws` with **browserify**  

It will also add `.once` and `.off`


3rd-Eden:
I'm not going to allow extending of native/build-in objects. So I have to decline this pull request as the side affects are worse than the benefits.

PR #272, opened 2013-12-08
fix Origin request options.

talrasha007:
some server will return 403 if Origin header is not a url.

here's the sample code that have the issue.

var WebSocket = require('ws'),
    ws = new WebSocket('wss://websocket.mtgox.com/mtgox?Currency=USD');

ws.on('message', function() { console.log(arguments); });

PR #494, opened 2015-05-06
adhere to RFC by removing default Origin header

tuukka:
This fixes the long-running issue https://github.com/websockets/ws/pull/272 by removing the Origin header by default.

The default value "host:port" was not syntactically valid according to RFCs, which caused (some) servers to reject such connections. According to the WebSocket RFC 6455 it should be one of 1) missing 2) "null" 3) "scheme://host:port" (defined by the Origin RFC 6454).

Semantically, an origin header is not required for non-browser clients, and if present it should describe the source of the resource which caused this request. If the caller of ws has such origin, it can set it in the options (test case provided).

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -578,7 +578,7 @@ function initAsClient(address, protocols, options) {
   var agent = options.value.agent;
 
   var headerHost = serverUrl.hostname;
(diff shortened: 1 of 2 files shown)


3rd-Eden:
Thank you for your pull request.

Issue #31, opened 2012-02-26
Missing 'close' event when using WebSocket as a client


youurayy:
Based on the WebSocket spec, section [Feedback from the protocol](http://dev.w3.org/html5/websockets/#feedback-from-the-protocol), the last part reads:

> "When the WebSocket connection is closed, ... 
> 2. ...  fire a simple event named error ...
> 3. Create an event ... with the event named close ...  and dispatch the event"

Now, when using the WebSocket object to initiate a connection to a remote WebSocket server, in multiple cases only the `error` event is emitted, but no `close` event.

I have mapped the cases and the relevant source locations. So, this happens:
1. When the server closes the connection, the listeners are cleaned from the WebSocket too early, and the `close` event is not delivered, [see here, line 485](https://github.com/einaros/ws/blob/1243c3fc3ad8cd0184e47f5b6a3b5721e97b0199/lib/WebSocket.js#L485) (perhaps it's OK to leave the listeners there, for either the user or the GC to clean them)
2. When the client is connecting to a server which is not listening, only the error event is emitted, [see here, line 451](https://github.com/einaros/ws/blob/1243c3fc3ad8cd0184e47f5b6a3b5721e97b0199/lib/WebSocket.js#L451) (this is ambiguous in the spec, because it says _when the connection is closed_ and here the connection wasn't established, but indeed all browsers emit a `close` here)
3. And the last one I haven't experienced, but by the looks it's the same case, when the server key is invalid, [see here, line 462](https://github.com/einaros/ws/blob/1243c3fc3ad8cd0184e47f5b6a3b5721e97b0199/lib/WebSocket.js#L462) (again, I think this should be consistent, e.g. when a user reconnection code waits for a `close` event, it should get it)

PR #496, opened 2015-05-07
Server socket does not close after calling .close()

michaelmcmillan:
This test shows that the socket on the WebSocketServer does not close after calling .close().

As others have suggested, the .close() method should be an async function that accepts a callback
for when it is finished with terminating the clients and closing up the socket.

https://github.com/websockets/ws/blob/71ff077a6b2c6aff05022521cdcd745e76cf5446/lib/WebSocketServer.js#L101-L131

Related issue: #31

### DIFF
--- a/test/WebSocketServer.test.js
+++ b/test/WebSocketServer.test.js
@@ -3,7 +3,9 @@ var http = require('http')
   , WebSocket = require('../')
   , WebSocketServer = WebSocket.Server
   , fs = require('fs')
-  , should = require('should');
+  , net = require('net')
+  , should = require('should')
+  , assert = require('assert');
 
 var port = 8000;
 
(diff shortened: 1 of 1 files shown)


michaelmcmillan:
Any advice on how I can implement this? I'll happily code it myself.

lpinca:
Hi,
I'm not sure why but the same also happens with a plain HTTP server:
```js
'use strict';

const http = require('http');
const net = require('net');

const server = http.createServer((req, res) => {
  res.statusCode = 200;
  res.setHeader('Content-Type', 'text/plain');
  res.end('Hello World\n');
});

server.listen(3000, '0.0.0.0', () => console.log('listening on *:3000'));
server.close();

const socket = net.connect(3000, () => {
  throw new Error('connected after calling close');
});

```

lpinca:
If you use

```js
server.listen(3000, '::1', () => console.log('listening on *:3000'));
```

it works as expected, but if you use

```js
server.listen(3000, '127.0.0.1', () => console.log('listening on *:3000'));
```

it doesn't and the connection is established.

Anyway this is not a `ws` issue. Feel free to file a bug in the Node.js repository.

PR #500, opened 2015-05-15
fix typo in README

nyx:
encypted -> encrypted


3rd-Eden:
Thanks for the doc fix!

PR #504, opened 2015-05-19
fix Host header port inequality check

davidburhans:
url.parse returns an object with the port as a string. This allows interoperablity with servers with a more strict interpretation of https://tools.ietf.org/html/rfc6455#page-17

The way the RFC is worded could be interpreted to mean the port on the Host header is optional except if it is a default port, in which case it should not be included.

This will change will leave the port off of the Host header if it is a default port, even if it was explicitly specified.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -581,7 +581,7 @@ function initAsClient(address, protocols, options) {
   // Append port number to Host header, only if specified in the url
   // and non-default
   if (serverUrl.port) {
-    if ((isSecure && (port !== 443)) || (!isSecure && (port !== 80))){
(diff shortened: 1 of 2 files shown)


3rd-Eden:
Could you add an additional test against this so it doesn't break in the future?

davidburhans:
Tests added. Not a big fan of the way I needed to expose that method to tests due to permissions concerns with running a server on port 80 and 443, but it works.

Kosta-Github:
Ran into the same issue. Can this PR be merged, please?

3rd-Eden:
@Kosta-Github Yes, sorry. This will be in the next release which i'll try to do today.

PR #512, opened 2015-06-11
Make websocket events be in line with browser

### LINKED ISSUE #511: addEventListener should include event type
I currently use a single event listener for all my websocket events and do different actions based on `event.type`. I notice this is missing from the node version so my code actually fails :(

**Edit**: Sorry, let me try to clarify what is happening:

I am using `ws.addEventListener()` because my library is isomorphic and needs to run in both node and the browser. Using the `addEventListener()` api in the browser exposes what event called the handler via `event.type`. This property is missing from the event object in the node version of `.addEventListener()`. It should echo back the method parameter when passing back the event object.

englercj:
Fixes #511.

The `type` property of the event should refer to the event type that is being fired. Unfortunately `type` for message events was being used to pass Binary/Text which was not only non-standard but broken applications that relied on the browser behavior.

This PR fixes events to have the proper type, and adds a non-standard `binary` property that has the value of `flags.binary`. If people were relying on `event.type === 'Binary'` or `event.type === 'Text'`, then this is a breaking change.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -411,7 +411,7 @@ WebSocket.prototype.addEventListener = function(method, listener) {
   var target = this;
 
   function onMessage (data, flags) {
-    listener.call(target, new MessageEvent(data, flags.binary ? 'Binary' : 'Text', target));
+    listener.call(target, new MessageEvent(data, !!flags.binary, target));
   }
 
   function onClose (code, message) {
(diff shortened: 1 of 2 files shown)


3rd-Eden:
Could you include a test?

Arnout Kazemier

englercj:
Will do!

englercj:
Some random test on node 0.8 failed on that second commit, but not sure how adding a test made that happen O.o

englercj:
Ping! Curious if there is anything else you need from me for this.

3rd-Eden:
No, sorry for the delay and thanks for your pull request.

PR #514, opened 2015-06-22
Adding proxy support

shuky19:


bigbagboom:
hi, can you tell how to use proxy server with your code? a sample please?

atian25:
hi @shuky19  any example?  see also #559

shuky19:
Ok, so I kind of put it here out of the blue, sorry for that...

Mainly this allows you to reuse an existing socket instead of creating one for the websocket connection.

The reason it helps connect a websocket server throug a proxy, is that you can open a socket, send the proxy a CONNECT message and then use that socket later on to negotiate a websocket connection.

The final flow will like this:

Open connection agains the proxy server
Negotiate with it using HTTP CONNECT
Receive 200 ok
Use that socket to transfer websocket data that will pass through the proxy to the websocket server.

Now, after doing the same fix for socket.io (localy) I understood the protocol creates multiple sockets in the habdshake period, so we may want to convert this param into a callback so we can create socket on demand.

I hope this clears things out,
Shuky

lpinca:
Closing this in favor of https://github.com/TooTallNate/node-https-proxy-agent or similar alternatives.

Thanks.

Issue #515, opened 2015-06-22
 Is there a way to limit the maximum size of a message?


Deathspike:
Please see: https://github.com/websockets/ws/issues/513

I'm not sure why _my issue report_ is closed for improper wording/opinion _of another person_?

I'd like to keep this issue going. Preferably in the original issue report as it contains the discussion.

Dillybob92:
Because, this is a not a 'forum', remember? It's a issue tracker where we cannot quote or make discussions with other people! That logic though... 

 I told you, the intellegence here is lacking and it needs to be posted on SO or another site. They don't want to admit they have a vulnerability and they will most likely end up saying: 'It's not a WS problem, it's a `nodejs` issue itself', please contact them! That's what they always do :)

Inb4 that happens.

3rd-Eden:
There was absolutely no reason for such discussion to take place in a bug report. But lets get back on topic here.

There is indeed no way to limit the maximum size of a message. The closed thing to message size moderation are:

- Looking at the bufferedAmount property of the connection and actively start dropping or blocking connections based on that.
- Attach an additional data listener to the socket that `ws` wraps. 
- Add functionality for this directly in to the project, which is probably the `received.js` file.

Deathspike:
I'd be in complete favor for the latter option. The first two options are not exactly straight to the point and therefore will be harder to reason about and/or maintain in future. On top of that, the WebSocket protocol describes the way a message is framed and a message is always framed with the length. Upon receiving this header, we know the size of the payload and whether or not it exceeds our maximum. This is the perfect place to drop the connection when the incoming payload is violation; we don't have to waste any further resources on something that we know is in violation of our constraints.

From a quick glance, I'd say that rule would have to be checked in the `getData` functions of the binary and text opcode handlers. That is https://github.com/websockets/ws/blob/master/lib/Receiver.js#L423 and https://github.com/websockets/ws/blob/master/lib/Receiver.js#L488. That said, I'm not too familiar with the `ws` internals and cannot possibly reason about the entire code base from a quick glance. But, that's what make sense to me.

Should I take a quick stab at this later and see what I can come up with? Or is someone else preferring to do this instead, or has alternatives/better locations and/or issues with the suggested checks at the suggest location?

Dillybob92:
Wait, I thought this wasn't a forum, why are you asking questions?

apsillers:
[...]

PR #520, opened 2015-06-27
Added maximumPayload option for non-hixie protocols

Deathspike:
See https://github.com/websockets/ws/issues/515

maximumPayload operates on the length of a packet. Fragments are
supported. Essentially it checks the size of a packet against the
configured maximumPayload and throws an error if the client sends a
packet that exceeds the maximum size. This way your server won't have to
allocate resources for it. Do note that with compression enabled, the
deflated resulting buffer can be larger than the maximum payload.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -17,6 +17,7 @@ This class is a WebSocket server. It is an `EventEmitter`.
   * `disableHixie` Boolean
   * `clientTracking` Boolean
   * `perMessageDeflate` Boolean|Object
+  * `maximumPayload` Number (requires `disableHixie`)
 * `callback` Function
 
 Construct a new server object.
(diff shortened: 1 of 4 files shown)


3rd-Eden:
Can't find anything wrong against this patch, so lgtm. The only thing that is missing is a test to ensure that the connection is properly aborted.

avesus:
Tests have failed because the API is changed. ```Receiver (extensions)``` becomes ```Receiver (maximumPayload, extensions)```. What is most adequate method to fix that?

renatoc:
Credits are for Deathspike, just changed code for pass in tests. 
#574

galambalazs:
any update on this or #574?

jssuttles:
Can we get this? It looks pretty good.

lpinca:
Closing this as duplicate of #650.
Thank you.

Issue #523, opened 2015-06-30
Binary messages give TypeError: Invalid non-string/buffer chunk


kamstrup:
I am unable to get any of the examples with binary messages working. Here's my example project:

index.js

``` js
var http = require('http')
  , express = require('express')
  , app = express()
  , WebSocket = require('ws')
;

app.use(express.static(__dirname + '/public'));

var server = http.createServer(app);
server.listen(8080);

var wss = new WebSocket.Server({ server: server })

wss.on('connection', function(sock) {
    console.log("CONN: " + sock);
    sock.on('error', function(err) { console.log("CONNERR: " + sock + " - " + err); });
    sock.on('close', function() { console.log("CONNECLS " + sock); });
    sock.on('message', function(msg) { console.log("msg: %s", msg); });

    var binData = new Float32Array(4);
    binData[0] = 0.0;
    binData[1] = 1.1;
    binData[2] = 2.2;
    binData[3] = 3.3;
    sock.send(binData,
              {binary: true, mask: false},
              function(err) { if (err) console.log("sock send err: " + err); })
})
```

public/index.html

``` html
<!DOCTYPE html>

<html lang="en">
<head>
    <meta charset="utf-8" />
    <title>WS TEST</title>
    <script>
      var ws = new WebSocket('ws://localhost:8080');
      ws.onmessage = function(event) {
        console.log("MSG: " + event)
      };
    </script>
</head>
<body>
WS TEST
</body>
</html>
```

Terminal session:

```
$ node index.js
CONN: [object Object]
sock send err: TypeError: Invalid non-string/buffer chunk
```

package.json:

``` json
{
  "name": "wshacks",
  "version": "1.0.0",
  "description": "hack around with ws",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "express": "^4.13.0",
    "ws": "^0.7.2"
  }
}
```

Node version 0.12.5

kamstrup:
Stack trace from the error:
```
TypeError: Invalid non-string/buffer chunk
    at validChunk (_stream_writable.js:186:14)
    at DeflateRaw.Writable.write (_stream_writable.js:215:12)
    at PerMessageDeflate.compress (/Users/kamstrup/wshacks/node_modules/ws/lib/PerMessageDeflate.js:260:17)
[...]

PR #524, opened 2015-07-01
Fix for sending typed arrays or ArrayBuffer when perMessageDeflate is enabled

nkzawa:
Related: https://github.com/websockets/ws/issues/523


3rd-Eden:
You, are amazing. Thanks

PR #528, opened 2015-07-08
websocket: allow path to be specified with unix domain sockets

varmalho:
This change allows URL path with params to be specified with
unix domain socket URLs.

And example URL with unix domain socket looks like:
  ws+unix:///absolule/path/to/uds_socket:/pathname?query_params

Note, ':' is the separator between socket path and URL path.

Existing URLs like the following also work:
  ws+unix:///absolule/path/to/uds_socket

In the above case, path gets set to '/'.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -652,7 +652,16 @@ function initAsClient(address, protocols, options) {
   }
 
   if (isUnixSocket) {
-    requestOptions.socketPath = serverUrl.pathname;
+    // An example URL with unix domain socket (UDS) would like:
+    //   ws+unix:///absolule/path/to/uds_socket:/pathname?query_params
+    // At this point, serverUrl.pathname is set to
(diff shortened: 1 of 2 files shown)


alandoherty:
This would be really useful, can't specify a path right now so having to fork and publish a separate version of ws for now.

alandoherty:
Last minute note, got everything working now.

```js
var udsParts = serverUrl.pathname.split(':'),
```
needs to be

```js
var udsParts = serverUrl.path.split(':'),
```

Otherwise it neglects the query parameter.

varmalho:
Thanks @alandoherty, that makes sense. I will apply your suggested fix in my local fork :) and push the change to this pull request as well.

lpinca:
Merged as 060b275d978ff7d095a92863fc3a1f7bcea878ed.
Thanks!

PR #533, opened 2015-07-21
add express example

coolaj86:
a simple expressjs example addressing a common use case


3rd-Eden:

PR #534, opened 2015-07-23
Issue path validation

fatelei:
When a client visits a server with a wrong path, the connection should be rejected.


JoshuaWise:
I second the importance of this change.

fatelei:
The old way: If the client connects with a wrong path, the connection will hang. Both the server and the client can know what hanppened.

lpinca:
Can you please add a test for this? Thank you.

fatelei:
ok

fatelei:
@lpinca I have added a test

lpinca:
@fatelei thank you. LGTM

PR #540, opened 2015-08-05
Allow pending CLOSE_WAIT connections to close

mkamioner:
HTTP server sets `allowHalfOpen` to true, but we want that to be false in websockets otherwise we get lots of connections with `CLOSE_WAIT` status


3rd-Eden:
Thanks for the pull request.

mkamioner:
My pleasure! When do you think this new fix will be released?

Issue #543, opened 2015-08-07
Inconsistent error/close behavior between client running in Node.js vs browser 


jondubois:
I noticed a few inconsistencies between how the client works when running in the browser vs in Node.js.

When the WS server crashes/is killed, if the client is running in the browser, the code is `1006` (which indicates an error - This is the desired behavior) but when running the client on Node.js, the code is `1000` (which indicates a normal disconnection) - This is a problem because a reconnection will not happen if the code is 1000 - I only want to reconnect if the connection was lost due to a failure (so the Node.js client will fail and not reconnect).

Another difference is that when running the client in the browser, when I try to connect to a non-existing server, it triggers the onclose handler with an error code (which is nice). When running the client in Node.js - The onclose handler is never called - Instead, the onerror handler will be called.

I'm using v0.7.2

jondubois:
To reproduce the issue, you should:

1. Create a WebSocketServer in its own Node.js process
2. Then connect to it using the WebSocket client (running in a separate Node.js process).
3. Now kill the server process

The Socket 'close' event will be emitted but the status code will be 1000 - This is different from the behavior in the browser where the status code will be 1006.

The WebSocket client appears to get an error as expected inside its cleanupWebsocketResources() method, but it incorrectly reports the status code as 1000 (normal).

jtblin:
Same problem. This is a major issue for us as we cannot handle reconnection properly, see https://tools.ietf.org/html/rfc6455#section-7.4.1 for exit codes.

It looks like `0.8.0` is trying to fix the issue by [returning 1006 in case of error](https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L924) but from my testing there is no error captured when the server crashes so it doesn't actually fix the problem.

The `cleanupWebsocketResources` is called from the `close` listener on https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L778 but doesn't seem to have any relevant information.

jondubois:
@jtblin Yes, I think a second problem which my solution doesn't solve is that the error is not always raised when it should be.

Also my previous patch doesn't fix this issue (which I mentioned in my initial post):

"... When running the client in Node.js - The onclose handler is never called - Instead, the onerror handler will be called."

I think in this case both the onerror AND the onclose handlers should be called. This issue was simple for me to work around though, it doesn't seem as bad as the first one.

jondubois:
[...]

PR #546, opened 2015-08-11
Status code should default to 1006 when socket closes abnormally with…

jondubois:
This is a fix for issue https://github.com/websockets/ws/issues/543

The new behavior is more consistent with that of the native browser WebSocket API.
According to RFC 6455, the status code 1006 is meant to be used to signify that the connection was closed abnormally (without receiving a close packet). So it looks like it's intended to be set from the client-side as I have done when an error occurs. Please confirm that I made the change in the correct place.

I tried to add a test for this, but it's difficult to do so (the socket's 'close' event did not trigger after calling srv.close() as I thought it would - I would have to launch the server in a separate process... Maybe there is another way).

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -912,6 +912,11 @@ function cleanupWebsocketResources(error) {
   this._closeTimer = null;
 
   if (emitClose) {
+    // If the connection was closed abnormally (with an error), 
+    // then the close code must default to 1006.
+    if (error) {
+      this._closeCode = 1006;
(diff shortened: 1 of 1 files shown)


3rd-Eden:
Thanks for the patch, these edge case are really hard to test but the pull request looks good to me, so I insta merged it.

bguiz:
@3rd-Eden +1

jondubois:
@3rd-Eden Thanks!

PR #550, opened 2015-08-13
Add io.js 3 to the testing matrix and test only major versions

ide:
Adds io.js to the Travis tests. Also removes tests for minor versions because versions like io.js 1.8.x and 2.5.x weren't being tested.

Also adds `sudo: false` which runs the tests on Travis' docker infrastructure, which launches tests more quickly: http://docs.travis-ci.com/user/migrating-from-legacy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+sudo: false
 npm_args: --ws:native
 node_js:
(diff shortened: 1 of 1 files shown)


3rd-Eden:
Thanks for the pull request <3

PR #551, opened 2015-08-14
Pass localAddress option to http request object.

aklaswad:
Hi

I know this doesn't have major usage, but I found that there is no way to specify which network interface ws would use, during I've been trying to compare the responsibility of multiple networks. Am happy if you consider to merge this change. thanks!


3rd-Eden:
Would be nice to have test against this.

aklaswad:
added some simple test, and fixed a bug.

aklaswad:
Grr, looks the test doesn't work on travis-ci, though it's fine for my osx.
And each versions of nodejs are throwing different exception...

3rd-Eden:
Thanks for your contribution

PR #557, opened 2015-08-20
Update gcc/g++ during Travis tests so that we get C++11

ide:
Building the native modules (utf-8-validate and bufferutil) with io.js 3.0 requires C++11. Travis uses an ancient version of gcc (4.6) which doesn't support C++11, so this commit tells Travis to install gcc/g++ 4.9 and use it with node-gyp.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,26 @@ language: node_js
 sudo: false
 npm_args: --ws:native
 node_js:
+  - "iojs-v3"
+  - "iojs-v2"
+  - "iojs-v1"
   - "0.12"
   - "0.11"
   - "0.10"
   - "0.9"
   - "0.8"
-  - "iojs-v1"
-  - "iojs-v2"
-  - "iojs-v3"
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - gcc-4.9
+      - g++-4.9
 before_install:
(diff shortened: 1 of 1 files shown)


ide:
The native modules build on io.js 3.x but now there is a core dump at runtime...
```
    #send
  ✓ very long binary data can be sent and received (with echoing server)
  ✓ can send and receive text data
  ✓ send and receive binary data as an array
  ✓ binary data can be sent and received as buffer
  ✓ ArrayBuffer is auto-detected without binary flag
  ✓ Buffer is auto-detected without binary flag
  ✓ before connect should fail
  ✓ before connect should pass error through callback, if present
  ✓ without data should be successful
  ✓ calls optional callback when flushed
  ✓ with unencoded message is successfully transmitted to the server
  ✓ with encoded message is successfully transmitted to the server
  ✓ with unencoded binary message is successfully transmitted to the server
  ✓ with encoded binary message is successfully transmitted to the server
  ✓ with binary stream will send fragmented data
  ✓ with text stream will send fragmented data
make[1]: *** [run-tests] Segmentation fault (core dumped)
make[1]: Leaving directory `/home/travis/build/websockets/ws'
make: *** [test] Error 2
```

3rd-Eden:
Shouldn't node-gyp automatically detect this?

ide:
It doesn't :(
This was the bare minimum that I could use to compile utf-8-validator and bufferutils.

PR #565, opened 2015-08-28
ws.sendv(array of buffers) iovecs support

guymguym:


lib/Sender.js
```diff
@@ -150,7 +151,16 @@ Sender.prototype.frameAndSend = function(opcode, data, finalFragment, maskData,
     return;
   }
 
-  if (!Buffer.isBuffer(data)) {
+  // iovecs support to allow callers to avoid bufer copying for single frame
+  if (data instanceof Array) {
```
3rd-Eden: Array.isArray() is prefered here over instanceof checks

3rd-Eden:
Would love to see tests being added for this. It will be impossible to track regressions otherwise.

lib/Sender.js
guymguym: np. is it also preferable over `util.isArray()` ?

guymguym: ignore the question, found the answer -
http://stackoverflow.com/questions/22798826/in-node-js-why-is-there-a-util-isarray-and-an-array-isarray

guymguym:
np about tests.

actually I see now that I could have implemented iovecs in upper layer in `lib/WebSocket.js` using `options.fin=false` and finally `true` similar to how it's done in `sendStream()`. the benefits would be that it would be handled once for both `Sender.js` and `Sender.hixie.js`, and also that it will support all types of buffers/strings mixed together. what do you think @3rd-Eden ?

guymguym:
added tests, and moved iovecs handling up from Sender to WebSocket (much shorter).

see below the updated benchmark results. 
the benefit is very noticeable with large buffers.

```
$ node bench/iovecs.js 
 
Sending messages with 1 buffers of 8 kB (options {"compress":false,"mask":false})
without iovecs:	9s	14.2 MB/ss
with    iovecs:	9s	13.59 MB/s
 
Sending messages with 8 buffers of 1 kB (options {"compress":false,"mask":false})
without iovecs:	9s	14.89 MB/s
with    iovecs:	9s	13.21 MB/s
 
Sending messages with 1 buffers of 128 kB (options {"compress":false,"mask":false})
without iovecs:	9s	95.99 MB/s
with    iovecs:	9s	135.52 MB/s
 
Sending messages with 8 buffers of 16 kB (options {"compress":false,"mask":false})
without iovecs:	9s	97.09 MB/s
with    iovecs:	9s	100.48 MB/s
 
Sending messages with 1 buffers of 1 MB (options {"compress":false,"mask":false})
without iovecs:	9.1s    205.29 MB/s
with    iovecs:	9s	316.14 MB/s
 
Sending messages with 8 buffers of 128 kB (options {"compress":false,"mask":false})
without iovecs:	9s	212.24 MB/s
with    iovecs:	9s	292.58 MB/s
 
Sending messages with 1 buffers of 128 kB (options {"compress":false,"mask":true})
without iovecs:	9s	77.77 MB/s
with    iovecs:	9s	96.98 MB/s
 
Sending messages with 1 buffers of 128 kB (options {"compress":true,"mask":true})
without iovecs:	9.1s    13.57 MB/s
with    iovecs:	9s	13.71 MB/s
done.
```

guymguym:
@3rd-Eden it would be great to get your review and fix any issues to get this into a release. thanks

guymguym:
@3rd-Eden Hi, any idea when is the next planned release and if this PR can be included? Thanks!

lib/WebSocket.js
```diff
@@ -262,6 +265,14 @@ WebSocket.prototype.send = function send(data, options, cb) {
 
       if (typeof cb === 'function') cb(error);
     });
+  } else if (Array.isArray(data)) {
+    // send array as iovecs
+    // using fin just like sending a stream but there's no need to
+    // wait for the stream data to be ready, so push all to the sender.
+    for (var i = 0; i < data.length; ++i) {
+      options.fin = (i + 1 === data.length);
+      this._sender.send(data[i], options, options.fin ? cb : undefined);
```
ronkorving: Should `options.binary` not be decided for each element in the array instead?

guymguym: Hi @ronkorving Thanks for the comment.

you mean in the case that `typeof(options.binary)==='undefined'`? 
sure, that can be done of course.

but there's a weird thing about this case that on the receiver side the message always appears with `flags.binary===true` even when all iovecs are strings and `options.binary===false` so I'm not sure how much it makes sense to support iovecs send with 'options.binary=false at all? or is that somehow expected that the receiver might be flagged that the message is binary although was sent as non binary?

ronkorving:
I was really going by function logic here, not so much use-case. But an array of strings... unthinkable?

lib/WebSocket.js
ronkorving: I was really going by function logic here, not so much use-case. But an array of strings... unthinkable?

guymguym: in my opinion it's thinkable indeed.
I'm just not sure if the fact that this is always received as binary makes it a confusing usage...

guymguym: I updated the test to run cases with strings and binary=false.
note these things:
* when you send mix of strings and buffers you must not use binary=true
* when you send an array of all strings you can use either binary=false or true, and the default is assumed to be true for the entire message.

does that make sense?

ronkorving: It does :+1:
Awesome job on this PR!

guymguym: :+1: Great! and thanks! I hope it makes it smoother for merging...

guymguym:
would it make more sense if this functionality will be added in a `ws.sendv(iovec_array)` instead of overloading the (already quite overloaded) `ws.send()` function?

ronkorving:
@guymguym actually, that could make a lot of sense imho

guymguym:
ok, changed to `sendv()`

guymguym:
feedback anyone?

README.md
```diff
@@ -59,6 +59,43 @@ Setting `mask`, as done for the send options above, will cause the data to be
 masked according to the WebSocket protocol. The same option applies for text
 data.
 
+### Sending iovecs
```
ronkorving: I think the "iovecs" lingo is not particularly user friendly...

lib/WebSocket.js
```diff
@@ -268,6 +270,60 @@ WebSocket.prototype.send = function send(data, options, cb) {
...
+ *
+ * @param {Array} the elements will be sent as a single message to the server
+ * @param {Object} Members - mask: boolean, binary: boolean, compress: boolean
+ * @param {function} Optional callback which is executed after the send completes
+ * @api public
+ */
+
+WebSocket.prototype.sendv = function sendv(data, options, cb) {
+  if (!Array.isArray(data)) {
+    throw new Error('array expected for sendv (WebSocket)');
```
ronkorving: You might wanna throw a TypeError instead

ronkorving:
Overall very happy, but left a few comments. Nothing big.

README.md
guymguym: I agree. any suggestions?
I could only come up with a super explicit sentence -
 "Sending a message composed of multiple buffers".

guymguym:
by the number of open PR's on this project, I probably shouldn't expect a merge soon :(

guymguym:
bump :8ball:

guymguym:
@3rd-Eden that's a small change for code, but one giant leap in buffers sendv performance...
would you consider for next release? any comments?

lpinca:
@guymguym thank you for the effort you put here.

I have one question: Isn't already possible to achieve this with the existing `send` method?
There are some redundant checks but If I'm not wrong it should be possible to do the same by calling `send` in a loop.

The reason I ask this is that I like this feature and I understand its importance but I would like to keep code base as small as possible. It is already bloated imo.

guymguym:
Hi @lpinca, Thanks for reviewing.

The thing is that calling `ws.send()` in a loop will send multiple ws framed messages and will require additional framing to identify in the application level. 

In `ws.sendv()` I actually did a similar thing but in a single ws frame by calling `this._sender.send()` multiple times with `fin=false` option until the last buffer which is called with `fin=true`.

guymguym:
@lpinca The reason for `sendv` is that in many cases when you construct large ws messages you end up doing `Buffer.concat(multiple buffers)` which is copyful and wasteful. `sendv` is meant to avoid this concat which is only needed since ws requires a single buffer, and not the tcp socket.

lpinca:
@guymguym you can use the `fin` option when calling `send`. For example,

```js
const bufs = [buf1, bu2, bufn];
const opts = { binary: true, fin: false }

bufs.forEach((buf, i) => {
  if (i === bufs.length - 1) opts.fin = true;
  ws.send(buf, opts);
});
```

sends a single message where each buffer is a fragment of the same message.

This doesn't work in `1.1.1` as there is bug where the `fin` option is unconditionally set to `true`, but it should be fixed in [master](ea50be7ab12ad028b41a315e924aa8c442c095d7).

guymguym:
@lpinca Thanks, sounds like a good option. Is it documented?

lpinca:
@guymguym no, but it should be.

guymguym:
@lpinca Small suggestion - how about making the API accept the fin option without a buffer so that it would be simpler to write like this:
```
const bufs = [...];
// append message buffers without finalizing
bufs.forEach(buf => ws.send(buf, { binary: true, fin: false }));
// finalize the message
ws.send(null, { binary: true, fin: true })
```

lpinca:
@guymguym I think that already works not 100% sure though.

lpinca:
200db3354fed6b97ee0c5bfbf4d37586783fc1c6 slightly improves the documentation for the `send` method.

lpinca:
@guymguym I'm going to close this, sorry. I think that using the `fin` option is better than adding a new method.

guymguym:
I agree completely. Thanks.

guymguym:
@lpinca Hey. I am using `fin` and it works great for sender side.
However the receiver still concats fragments before emitting the message.
See - https://github.com/websockets/ws/blob/master/lib/Receiver.js#L355-L359

In my tests I changed the receiver code locally to emit the fragments array directly instead of concatenation which is copyful, and this makes a very big performance improvement - in my case throughput increased by 50%.

The reason is that for binary protocol that is sending fragments it also uses the same fragments when parsing the protocol on the receiver side, and by eliminating the memory copy we get more CPU cycles doing send/receive.

Would you consider this as a new option?
One way is to add a new option like `binaryType` called `receiveType='fragments' or 'messages'`
Second way is to leave the current events intact and just do in addition `emit('fragments', fragments_array)` so that only callers who intentionally want fragments will register.
What do you think?

lpinca:
It makes sense if the change is not too invasive and does not affect performance.

guymguym:
OK, I think it will be pretty easy to keep zero effect when not used.
Should I will suggest a PR and you can review?

lpinca:
Sure, thanks!

Issue #553, opened 2015-08-16
Cannot assign to read only property 'target' of invalid compressed data


binarykitchen:
Seeing this weird error few times and have no clue what the real cause is:

```
TypeError: Cannot assign to read only property 'target' of invalid compressed data
    at WebSocket.onError (/var/www/videomail/production/source/node_modules/ws/lib/WebSocket.js:422:18)
    at emitTwo (events.js:87:13)
    at WebSocket.emit (events.js:172:7)
    at Receiver.onerror (/var/www/videomail/production/source/node_modules/ws/lib/WebSocket.js:832:10)
    at Receiver.error (/var/www/videomail/production/source/node_modules/ws/lib/Receiver.js:317:8)
    at /var/www/videomail/production/source/node_modules/ws/lib/Receiver.js:528:32
    at /var/www/videomail/production/source/node_modules/ws/lib/Receiver.js:354:9
    at InflateRaw.onError (/var/www/videomail/production/source/node_modules/ws/lib/PerMessageDeflate.js:222:5)
    at emitOne (events.js:77:13)
    at InflateRaw.emit (events.js:169:7)
    at Zlib._handle.onerror (zlib.js:366:10)
```

psorowka:
I also observe this type of error, when using iojs v3. 

The error occurs here: https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L425-L429 

From a perspective of ws this happens because the emitted error is a `string` and not an object of type `Error`. 

A workaround seems inserting:

```
if(typeof event == 'string') {

  event = new Error(event);
}
```

in the `onError` function. However, I also haven't figured out yet why the error happens in the first place.

3rd-Eden:
@psorowka Do you have any trace on where the string originates from because I'm pretty sure we're all using Error instances.

psorowka:
@3rd-Eden I see the same trace as @binarykitchen is showing.

I cannot strictly reproduce the error. For me it happens in test cases of the mosca mqtt broker. In about 30% of the cases, the tests run without error, the rest raises either the above error or the similar variant: 

```
 Uncaught TypeError: Cannot assign to read only property 'target' of continuation frame cannot follow current opcode
      at WebSocket.onError (/app/node_modules/websocket-stream/node_modules/ws/lib/WebSocket.js:422:18)
      at emitTwo (events.js:87:13)
      at WebSocket.emit (events.js:172:7)
[...]

PR #569, opened 2015-08-31
Fixed error emission on receiver error

psorowka:
As discussed in https://github.com/websockets/ws/issues/553#issuecomment-136507574

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -848,7 +848,7 @@ function establishConnection(ReceiverClass, SenderClass, socket, upgradeHead) {
   self._receiver.onerror = function onerror(reason, errorCode) {
     // close the connection when the receiver reports a HyBi error code
     self.close(typeof errorCode !== 'undefined' ? errorCode : 1002, '');
-    self.emit('error', reason, errorCode);
+    self.emit('error', new Error(reason));
   };
 
   // finalize the client
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -848,7 +848,7 @@ function establishConnection(ReceiverClass, SenderClass, socket, upgradeHead) {
   self._receiver.onerror = function onerror(reason, errorCode) {
     // close the connection when the receiver reports a HyBi error code
     self.close(typeof errorCode !== 'undefined' ? errorCode : 1002, '');
-    self.emit('error', new Error(reason));
+    self.emit('error', (reason instanceof Error) ? reason : (new Error(reason)));
   };
 
   // finalize the client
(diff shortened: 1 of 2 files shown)


lib/WebSocket.js
```diff
@@ -848,7 +848,7 @@ function establishConnection(ReceiverClass, SenderClass, socket, upgradeHead) {
   self._receiver.onerror = function onerror(reason, errorCode) {
     // close the connection when the receiver reports a HyBi error code
     self.close(typeof errorCode !== 'undefined' ? errorCode : 1002, '');
-    self.emit('error', reason, errorCode);
+    self.emit('error', new Error(reason));
```
binarykitchen: to be safe i would add a type check here for `reason`. it might be already an instance of error. only do `new Error()` when it's a string

psorowka:
@binarykitchen valid point. although I think that this function is tied closely to the implementation of the `Receiver` class (which obviously expects a string rather than an error object in its error interface), this is a bit of convenience for future extensions :)

lib/WebSocket.js
psorowka: @binarykitchen valid point. although I think that this function is tied closely to the implementation of the Receiver class (which obviously expects a string rather than an error object in its error interface), this is a bit of convenience for future extensions :)

binarykitchen:
looking good now

binarykitchen:
but i think unit tests are missing and crucial here

psorowka:
here you go :)

binarykitchen:
good job! but probably you will want to remove that `console.log(error);`

psorowka:
hehe do you still see that? I have already pushed an updated version without that :)

binarykitchen:
great - all good like that. many thanks again

psorowka:
you are welcome. thanks for quality enforcement

stoffeastrom:
Just upgraded to nodejs v4.0.0 and got this error. I copy pasted this solution and it solves the problem.

peol:
It seems to break compatibility with node.js 4.0, also verified that it seems to fix the compatibility issue.

stoffeastrom:
@binarykitchen any ETA on merging this?

binarykitchen:
@stoffeastrom i am not the maintainer of this package - @3rd-Eden is.

3rd-Eden:
*contributor

binarykitchen:
oh, @einaros it is then. can you merge this?

stoffeastrom:
:shipit:

ekristen:
+1

rupamkhaitan:
when is the plan to merge this to main branch ?

3rd-Eden:
Thanks for your contirubtion.

PR #572, opened 2015-09-02
Return new instance if new keyword is missing.

tabone:
When `WebSocket` and `WebSocketServer` are invoked as a function (i.e. without the `new` keyword), instead of throwing an error (change introduced in PR https://github.com/websockets/ws/pull/468/), it should create and return a new instance.

In addition to this I also removed the related test cases.


3rd-Eden:
Thanks for your contribution :trophy:

tabone:
:+1:  my pleasure

PR #582, opened 2015-09-14
Remove 'new' from docs.

tabone:
Remove `new` keyword from examples and docs. Forgot to include this with PR https://github.com/websockets/ws/pull/572.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ npm install --save ws
 
 ```js
 var WebSocket = require('ws');
-var ws = new WebSocket('ws://www.host.com/path');
+var ws = WebSocket('ws://www.host.com/path');
 
 ws.on('open', function open() {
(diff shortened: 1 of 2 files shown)


3rd-Eden:
The `new` keyword is still required if you want to run the same code in browsers. So I cannot merge this.

tabone:
Correct me if I'm wrong but with the changes introduced in PR https://github.com/websockets/ws/pull/572  isn't the `new` keyword optional?

3rd-Eden:
Yes, but people still use this module to provide WebSocket constructor in Node so the code you write is compatible with browser & node. The WebSocket constructor in the browser still requires the new keyword so keeping it is safer than removing it.

tabone:
Makes sense :+1: . Thanks for your time. I'll close this PR then.

PR #587, opened 2015-09-17
Fix binary mode sending of strings

vmarkovtsev:
See http://stackoverflow.com/a/7094957

When using fs.createReadStream with binary encoding, passing data into send() leads to corruption.


lpinca:
I think the best way to work around this is to not specify an encoding when using `fs.createReadStream()`. In this way each chunk of data is a `Buffer` which can be passed directly into send. The added benefit is that your data is not converted again into a `Buffer` inside `send()` leading to better performance.

Passing a latin1 (alias for binary) string into `send()` and checking for the opcode will not have the desired effect as the opcode for a text frame is 1 so you'll end up creating a buffer with utf8 encoding from a latin1 string again which I think is not what you want.

lpinca:
Closing this, please comment or reopen if needed.

Issue #594, opened 2015-10-01
permessage-deflate does not work with Chrome.


yutakahirano:
Please see https://code.google.com/p/chromium/issues/detail?id=536627 for the minimized test case.

In "finish" functions in Receiver.js, incoming data is stored to a variable |data|, a.k.a. |packet|. The data may be modified after the function returns (I assume calling |endPacket| is the trigger but I'm not sure), but permessage-deflate module uses the data after that. As a result it sees modified data and that leads to deflate errors.

yutakahirano:
This may be related to #585.

nkzawa:
Thanks for your report!  Will fix asap.

PR #595, opened 2015-10-02
Fix error when permessage-deflate is enabled

nkzawa:
Fix https://github.com/websockets/ws/issues/594

This PR forces to create a new buffer on each packet process so that buffers would not be affected each other.
As far as I benchmarked this by `bench/speed.js`, there is no performance problem.


greggman:
Any schedule to push this to a release?

nkzawa:
@3rd-Eden can you merge this too?

Issue #527, opened 2015-07-08
Server port stays in use after `close()` method call


lonelyelk:
It seems there is no way to tell, when server address and port are stopped being used:

``` javascript
var WebSocketServer = require('ws').Server;
var wss = new WebSocketServer({port:9000});
wss.close();
wss = new WebSocketServer({port:9000});
```

This gives EADDRINUSE on both node v0.12.6 and iojs v2.3.3. If this is an asynchronous method, than there should be a callback but `wss.on('close', fn)` is never called.

The problem appears when trying to test something that requires creation of the same server in `beforeEach`.

Eeems:
+1 on this. I'm racking my brain currently attempting to figure out a way to wait until it's free. It's causing problems with some hot-reloading I'm trying to get working.

woverton:
+1

woverton:
Right, I did a little hunting and found this https://github.com/joyent/node/issues/55

It turns out that ws is doing nothing wrong.

```this_server``` is a simple node http server, and then ws calls close on it.

https://github.com/websockets/ws/blob/master/lib/WebSocketServer.js#L59

**Comment by Eeems** [comment]
@woverton the problem is that there is no exposed way to pass a callback for after the connection finally closes. WS closes it fine, but you run into timing issues due to t knowing if its closed yet or not.

**Comment by woverton** [comment]
I know, It's just the error exists in node http not the lib. I cant see that anything can be done unless you want to switch to another port or something I'm afraid...

(I'm not a ws dev, just looked at the code and then did some digging)

**Comment by Eeems** [comment]
Nodes http server has a callback at fires after the port is released. All ws needs to do is expose it. https://nodejs.org/api/http.html#http_event_close

**Comment by woverton** [comment]
I tried making the change. It didn't seem to work however...

https://github.com/woverton/ws/commit/131c5e96dabf4318e1981564bc82058c985e4d56

**Comment by Eeems** [comment]
Interesting. I've not had problems with it when I'm us in it with http servers manually. What version of nodejs are you using?

**Comment by woverton** [comment]
With just http servers it works fine, I tested that once confused.

**Comment by woverton** [comment]
I think it may be todo with this line here: https://github.com/woverton/ws/blob/131c5e96dabf4318e1981564bc82058c985e4d56/lib/WebSocketServer.js#L137

**Comment by woverton** [comment]
We might be deleting the server before it fully finishes closing

**Comment by woverton** [comment]
Interestingly:

```
[...]

PR #600, opened 2015-10-12
Fix issue with server not releasing port

woverton:
Please see https://github.com/websockets/ws/issues/527 for the discussion about this issue.

Let me know if you need anything.

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -130,9 +130,11 @@ WebSocketServer.prototype.close = function(cb) {
   // close the http server if it was internally created
   cb = cb || function(){};
 
+  var self = this;
+
   this._server.once('close', function(){
     cb()
-    delete this._server;
+    delete self._server;
   });
 
   this._closeServer();
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -128,15 +128,15 @@ WebSocketServer.prototype.close = function(cb) {
   }
 
   // close the http server if it was internally created
-  try {
-    if (typeof this._closeServer !== 'undefined') {
-      this._closeServer(cb || function(){});
-    }
-  }
-  finally {
+  cb = cb || function(){};
+
+  this._server.once('close', function(){
+    cb()
     delete this._server;
-  }
-  
+  });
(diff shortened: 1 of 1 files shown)


Gray1989:
You don't need to rewrite it like you did. These changes would work fine:

```javascript
this._closeServer = function(cb) { self._server?self._server.close(cb):cb(); };

WebSocketServer.prototype.close = function(cb) {
  // terminate all associated clients
  var error = null;
  try {
    for (var i = 0, l = this.clients.length; i < l; ++i) {
      this.clients[i].terminate();
    }
  }
  catch (e) {
    error = e;
  }

  // remove path descriptor, if any
  if (this.path && this._server._webSocketPaths) {
    delete this._server._webSocketPaths[this.path];
    if (Object.keys(this._server._webSocketPaths).length == 0) {
      delete this._server._webSocketPaths;
    }
  }

  // close the http server if it was internally created
  try {
    if (typeof this._closeServer !== 'undefined') // checking if it matches the string "undefined", possible logic error?
      this._closeServer(cb);
    else if(cb !== undefined) cb();
  }
  finally {
    delete this._server;
  }
  if (error) throw error;
}
```

Ignoring the possible logic error already implemented... The if/else was simply shortened and the callback was passed as-is through to the method. Also not sure why you created the closure for simply calling the method and deleting the _server object. Not deleting the _server object when it was originally designed to be deleted could lead to more logic errors further on in the application or in the way people use it so it's always best to try to change as little as possible. Adding functionality is always good but changing the way things work needs to be done with extreme caution.

woverton:
Yeah you're right. I should have remade my fix after the time I spent tracking it down. Finding the fix took a little longer than it should have. Feel free to drop this and create a PR with your fix in.

rimmartin:
Hi,

Since this was in conflict with current master branch I tried applying @Gray1989 but now thee callback gets called twice because another callback was put in at:
https://github.com/websockets/ws/blob/master/lib/WebSocketServer.js#L156

Best way around this?  and I'd like to see it get pulled in to the master

lpinca:
Closing as this should be addressed in master.

PR #602, opened 2015-10-14
Add duplex stream functionality

Gray1989:
Complete with full flow-control support for piping streams, built with
both usage and efficiency in mind.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -3,6 +3,7 @@
 /*!
  * ws: a node.js websocket client
  * Copyright(c) 2011 Einar Otto Stangvik <einaros@gmail.com>
+ * Stream functionality added by Jonathan Gray [Gray1989] (Oct-14-2015)
  * MIT Licensed
  */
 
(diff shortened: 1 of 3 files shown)


Gray1989:
Actually maybe hold off on this for a bit. The stream will have its own pause and resume methods which is why the ones included aren't necessary. However those will only pause and resume the actual stream's data event and will not modify the state of the underlying socket's pause/resume methods until the internal buffer of the stream is either full or depleted. This should be a simple fix with by overwriting the pause and resume methods that are inherited to call stream.Readable.prototype.pause.call(this) and stream.Readable.prototype.resume.call(this) before calling this._socket.pause() and this._socket.resume(). That should complete the update. However I have more to do as-well and I'm pretty tired. I'll finish it off within 12 hours

Gray1989:
Sorry for the late fix but everything should be working now. Probably needs some testing but if anyone has any issues pertaining to the added streaming functionality or anything caused by it please just let me know.

lib/WebSocket.js
```diff
@@ -48,7 +48,7 @@ function WebSocket(address, protocols, options) {
     return new WebSocket(address, protocols, options);
   }
 
-  EventEmitter.call(this);
+  stream.Duplex.call(this); // [Gray1989] Changed from EventEmitter
```
ronkorving: Are all these comments necessary?

ronkorving:
Out of curiosity, will this allow the mixing of strings and buffers over the stream?

JoshuaWise:
In my opinion, this is the responsibility of application modules or modules that wrap `ws`. What do others think?

@3rd-Eden 
@JacksonTian

JacksonTian:
-1 from me

PR #605, opened 2015-10-26
Clean event handlers on WebSocketServer's internal http server when closing

sebpiq:
This gave me quite a headache to debug in a test suite where I had an express server running on the same port as the `WebSocketServer`. I had no need to close the express server, but I needed to close the `WebSocketServer` to do some clean-up between tests. And I had weird error messages popping-up at second test. Turns-out all event handlers (e.g. `"upgrade"`) where still called on the old `WebSocketServer` even though I had closed it after end of first test.

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -72,13 +72,15 @@ function WebSocketServer(options, callback) {
       this._server._webSocketPaths[options.value.path] = 1;
     }
   }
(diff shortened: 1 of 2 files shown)


131:
i came to the same bug & same solution

131:
Please merge :+1:

3rd-Eden:
Thanks for your contribution, it will end up in the new minor release.

Issue #497, opened 2015-05-08
Memory leak when 'perMessageDeflate' is enabled


feross:
With the default WebSocketServer options (i.e. `perMessageDeflate` enabled), I see a memory leak.

I'm using iojs v1.8.1. 

Setting `{ perMessageDeflate: false }` solves the issue. Has anyone seen this before?

V1c70r:
I have the same problem too. perMessageDeflate: false helps.

PR #608, opened 2015-11-01
Release memory used by PerMessageDeflate extension

darrachequesne:
Hi! It seemed the memory allocated by perMessageDeflate is not properly freed, thus causing memory leaks when this feature is enabled (as reported here https://github.com/websockets/ws/issues/497)

As documented [here](http://www.zlib.net/manual.html) and implemented [here](https://github.com/nodejs/node/blob/master/src/node_zlib.cc#L79), calling `.close()` on DeflateRaw and InflateRaw objects should (at least partially) fix these leaks.

Related: https://github.com/socketio/socket.io/issues/2251

*(please tell me this is it)*

### DIFF
--- a/lib/PerMessageDeflate.js
+++ b/lib/PerMessageDeflate.js
@@ -69,6 +69,23 @@ PerMessageDeflate.prototype.accept = function(paramsList) {
 };
 
 /**
+ * Releases all resources used by the extension
+ *
+ * @api public
+ */
+
+PerMessageDeflate.prototype.cleanup = function() {
+  if (this._inflate) {
+    this._inflate.close();
+    this._inflate = null;
+  }
+  if (this._deflate) {
+    this._deflate.close();
+    this._deflate = null;
+  }
+};
+
+/**
  * Accept extension offer from client
  *
  * @api private
@@ -231,9 +248,11 @@ PerMessageDeflate.prototype.decompress = function (data, fin, callback) {
   }
 
   function cleanup() {
+    if (!self._inflate) return;
     self._inflate.removeListener('error', onError);
     self._inflate.removeListener('data', onData);
     if (fin && self.params[endpoint + '_no_context_takeover']) {
+      self._inflate.close();
       self._inflate = null;
     }
   }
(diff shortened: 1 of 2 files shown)


darrachequesne:
(just to be sure :sunny:) ping @3rd-Eden @nkzawa

nkzawa:
:+1: awesome

But I'm not sure if this is the root cause, since zlib objects should be released with socket instances when no reference exists.
Anyway explicit cleaning up would be nice.

darrachequesne:
@nkzawa hi! that's what I thought too, but running the following [gist](https://gist.github.com/darrachequesne/630c3f14154e4489e647) seems to show a different behaviour

* with `perMessageDeflate: false`, RSS tends to stabilize around 70 Mo
* with `perMessageDeflate: true` (without this commit), RSS tends to stabilize around 180 Mo
* with `perMessageDeflate: true` (with this commit), RSS tends to stabilize around 100 Mo

Meanwhile, the HeapSize goes up and down (which is normal, as GC runs). 

Are you able to reproduce my results? Or is there a problem with my test?

I mean, the extra memory is normal, as it comes as a cost for compression (as said [here](https://www.igvita.com/2013/11/27/configuring-and-optimizing-websocket-compression/)). But as said [here](http://www.zlib.net/zlib_how.html) (that's a lot of links :smile:), memory has to be deallocated by calling `deflateEnd()` (or `inflateEnd()`), which is what `.close()` does here.

But it may not be the root cause (and that's where I need your help).

Note: `node -v` returns `v0.12.7`

nkzawa:
Hi, sorry for the delay. I could reproduce similar results with your scripts, but maybe this means it's not leak? Though I have no idea about the difference of RSS.

darrachequesne:
You're right, the term 'leak' was probably wrongly used here, please excuse me.

I found this https://github.com/nodejs/node-v0.x-archive/issues/4172. Could this be related explain the difference in RSS? (as we are now explicitly calling .close())

Source: https://github.com/nodejs/node/blob/master/src/node_zlib.cc#L97

rauchg:
@3rd-Eden can you merge please?

3rd-Eden:
PR will be adressed this week. Sorry for the hold up.

darrachequesne:
Great, thanks!

Yuyi:
Maybe this issue is still at #804 .

PR #615, opened 2015-11-10
Fix global context detect in strict mode

NeZeD:
`var global = (function() { return this; })()` - isnt great solution, it should return `undefined` when ws defined in scope with 'use strict', or when node_modules transpiled by babel, which add 'use strict' by default.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Securing_JavaScript

### DIFF
--- a/lib/browser.js
+++ b/lib/browser.js
@@ -3,7 +3,7 @@
  * Module dependencies.
  */
 
-var global = (function() { return this; })();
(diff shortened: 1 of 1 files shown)


3rd-Eden:
Closing this as in the 1.0 the client code has been removed from this project.

PR #627, opened 2015-12-03
Change read state values to unique strings

feifanzhou:
Currently, the ready state values are low integers (0 – 3 inclusive). This could very likely conflict with other constants or cause subtle errors else. Replacing the constants with more unique strings helps prevent conflicts.


3rd-Eden:
This convention is inline with the browser implementation:

```js
console.log(WebSocket.CONNECTING) // 0
```

Issue #628, opened 2015-12-05
Uncaught "zlib binding closed" errors


DanReyLop:
I'm setting up a simple WebSockets server, and this error (apparently randomly) happened:

```
Error: zlib binding closed
    at DeflateRaw.Zlib._transform (zlib.js:485:15)
    at DeflateRaw.Transform._read (_stream_transform.js:167:10)
    at DeflateRaw.Transform._write (_stream_transform.js:155:12)
    at doWrite (_stream_writable.js:292:12)
    at clearBuffer (_stream_writable.js:396:7)
    at onwrite (_stream_writable.js:331:7)
    at WritableState.onwrite (_stream_writable.js:89:5)
    at afterTransform (_stream_transform.js:79:5)
    at TransformState.afterTransform (_stream_transform.js:54:12)
    at Zlib.callback (zlib.js:614:5)
```

It happened in a moment where the client was connecting and disconnecting continuously, so maybe it's caused by a race condition when the connection is closed abruptly. I wasn't able to reproduce it again. I had the PerMessageDeflate extension enabled (the WS server was configured as default).

It looks like it's the same issue as this: https://github.com/faye/permessage-deflate-node/issues/1

Edit: I forgot, I'm using version 0.8.1 of ws.

Niceman35:
Same Error here.
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: zlib binding closed
    at DeflateRaw.Zlib._transform (zlib.js:487:15)
    at DeflateRaw.Transform._read (_stream_transform.js:179:10)
    at DeflateRaw.Transform._write (_stream_transform.js:167:12)
    at doWrite (_stream_writable.js:301:12)
    at clearBuffer (_stream_writable.js:411:7)
    at onwrite (_stream_writable.js:344:7)
    at WritableState.onwrite (_stream_writable.js:105:5)
    at afterTransform (_stream_transform.js:99:5)
    at TransformState.afterTransform (_stream_transform.js:74:12)
    at Zlib.callback (zlib.js:611:5)

NodeJS version v0.12.7
WS version 0.8.1 (even with latest commit)

darrachequesne:
Hi! It seems to be caused by the use of `this._deflate.close` but I don't understand why the error is not caught by the listener:

```javascript
var PerMessageDeflate = require('./lib/PerMessageDeflate');
var extension = new PerMessageDeflate();

extension.accept([{}]);
[...]

PR #632, opened 2015-12-15
Fix "zlib binding closed" errors

darrachequesne:
This should fix https://github.com/websockets/ws/issues/628, which I think is caused by:

- deflate is closed
- `deflate.write()` then triggers an 'error' event
- `cleanup()` removes the listener attached to the 'error' event
- `deflate.flush()` triggers the uncaught error 'zlib binding closed' 

Does this make any sense?

### DIFF
--- a/lib/PerMessageDeflate.js
+++ b/lib/PerMessageDeflate.js
@@ -76,12 +76,20 @@ PerMessageDeflate.prototype.accept = function(paramsList) {
 
 PerMessageDeflate.prototype.cleanup = function() {
   if (this._inflate) {
-    if (this._inflate.close) this._inflate.close();
-    this._inflate = null;
+    if (this._inflate.writeInProgress) {
+      this._inflate.pendingClose = true;
+    } else {
+      if (this._inflate.close) this._inflate.close();
+      this._inflate = null;
+    }
   }
   if (this._deflate) {
-    if (this._deflate.close) this._deflate.close();
-    this._deflate = null;
+    if (this._deflate.writeInProgress) {
+      this._deflate.pendingClose = true;
+    } else {
+      if (this._deflate.close) this._deflate.close();
+      this._deflate = null;
+    }
   }
 };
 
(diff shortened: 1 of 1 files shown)


nkzawa:
:+1: It might be better to add a guard to `compress` and `decompress` for checking if the instance is already cleaned up, like https://github.com/websockets/ws/blob/master/lib/Receiver.js#L269 .

lib/PerMessageDeflate.js
```diff
@@ -76,10 +76,18 @@ PerMessageDeflate.prototype.accept = function(paramsList) {
 
 PerMessageDeflate.prototype.cleanup = function() {
   if (this._inflate) {
+    if (this._inflate.writeInProgress) {
+      this._inflate.pendingClose = true;
+      return;
```
DanReyLop: I don't know this codebase, but... I think this means that if there is an ```_inflate.writeInProgress``` when the connection is closed, the ```_deflate``` will never be closed. Instead of returning, this should be rewritten as:
```
if (this._inflate.writeInProgress) {
      this._inflate.pendingClose = true;
} else {
      if (this._inflate.close) this._inflate.close();
     this._inflate = null;
}
```

darrachequesne: You're absolutely right! I'll patch this.

rauchg:
:+1: 

Tested & works perfectly. No more

```
node: ../src/node_zlib.cc:132: static v8::Handle<v8::Value> node::ZCtx::Write(const v8::Arguments&): Assertion `!ctx->pending_close_ && "close is pending"' failed.
Aborted (core dumped)
```

3rd-Eden:
Adressing this tomorrow 

Arnout Kazemier

darrachequesne:
Thanks! That should fix https://github.com/websockets/ws/issues/633 too.

darrachequesne:
Hi! @3rd-Eden could you please release this as version 0.8.2 before the big 1.0? :pray: (the current 0.8.1 is not really usable due to that issue).

3rd-Eden:
Release of 1.0 goes out tomorrow. No 0.8.2 -- if you have issues, turning off deflate by default

Arnout Kazemier

Issue #459, opened 2015-02-27
new public method setKeepAlive for WebSocket


humingchun:
setKeepAlive is a great way to detect dead TCP connection and Node.js net module has a method called `socket.setKeepAlive([enable][, initialDelay])`.

To set keepalive in ws, we have to use `ws._socket` which is OK but not elegant. I would suggest adding setKeepAlive to ws so it clearly shows ws has the way to maintain a connection or detect dead connection.

The code could be
```
WebSocket.prototype.setKeepAlive = function setKeepAlive(enable, initialDelay) {
  return this._socket.setKeepAlive(enable, initialDelay);
};
```

nkzawa:
It seems `keepAlive` doesn't mean always works well.

http://stackoverflow.com/questions/23238319/websockets-ping-pong-why-not-tcp-keepalive

So I think we usually should use `ping/pong`.

humingchun:
Great source for compare WebSocket ping/pong and TCP keepalive :+1: 

So it's clear WebSocket ping/pong could do much more than TCP keepalive, and even more reliable. But in controlled network (like no HTTP proxy), use TCP keepalive to do dead-pear-detection is much more efficient than WebSocket ping/pong.

I used this method when I am developing a IoT project which need to show the online status of real device. Using TCP keepalive intead of WebSocket ping/pong adds no coding to exising project and reduces the workload on device side.

3rd-Eden:
@humingchun if you are feeling strongly about this you can create a pull request for it and I have no problem with accepting it (if it's properly tested and documented)

glennschler:
Is it better to instantiate the Websocket with this keepAlive option instead of adding a new public method?

For example:
### new ws.WebSocket(address, [protocols], [options])

* `options` Object
  * `keepAlive` Object
    * `enable` Boolean 
    * `initialDelay` Number

https://github.com/websockets/ws/blob/master/doc/ws.md#new-wswebsocketaddress-protocols-options

humingchun:
@glennschler I think keepAlive is more useful at WebSocket Server side, and I don't find any method to pass options to ws instance returned by WebSocket server' `connection` event.

nkzawa:
I think it might be nice to implement keep-alive using websocket ping/pong instead of just setting tcp keep-alive.

JustinTArthur:
If you're like me and prefer the lighter-weight TCP keep-alive for your use-case, ws provides some options for configuring the underlying network layer:
```javascript
[...]

PR #635, opened 2015-12-18
KeepAlive implementation

FlorianBELLAZOUZ:
Hi ! @nkzawa @humingchun @3rd-Eden
As discussed in the #459 issue.

I implement a Keep Alive protocol using websocket ping/pong.
When the websocket don't reply to a ping with a pong, the serveur close the dead connection .

As it say in the documentation you can active it with a function : 
`websocket.setKeepAlive([enabled], [interval],  [timeout])`

or as recommend @glennschler you can configure a server to enable it for all websockets:
`new ws.Server([options], [callback])`
- `options` Object
  - `keepAlive` Boolean|Object
    - `interval` Number
    - `timeout` Number

This new feacture is documented and unit tested.

Aside from that, I remain open to all suggestions.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -192,6 +192,61 @@ WebSocket.prototype.pong = function(data, options, dontFailWhenClosed) {
 };
 
 /**
+ * Enabled a keepAlive based on ping pong process
+ *
+ * /!\ Warning : The websocket client must respond to the ping request with a pong message.
+ *               That is not necessarily the case if you use an other websocket client than WS.
+ *
+ * @param {boolean} enabled - enable or disable the keepAlive protocol.
+ * @param {number} interval - the interval of time in milliseconds between a ping request and a pong response. default: 2000.
+ * @param {number} timeout - the maximum waiting time between a ping request and the pong response before closing the socket. default: 10000.
+ * @api public
+ */
+WebSocket.prototype.setKeepAlive = function(enabled,interval,timeout){
+  var self = this;
+  if (enabled === false) return remove();
+
+  if (typeof interval !== "number" && interval !== undefined) throw new TypeError('interval must be a number')
+  if (typeof timeout !== "number" && timeout !== undefined) throw new TypeError('timeout must be a number')
+
+  this._keepAliveInterval = interval || 2000;
+  this._keepAliveTimeout = timeout || 10000;
+  this._keepAliveEnabled = enabled || true;
+  this._keepAliveIdTimeout = pingRequest(this,this._keepAliveTimeout);
+  this._keepAliveIdInterval = null;
+
+  this.on('pong',pongResponse);
+  this.on('close',remove);
+
+  function pongResponse(){
+    clearTimeout(self._keepAliveIdTimeout);
+    self._keepAliveIdInterval = setTimeout(function(){
+      self._keepAliveIdTimeout = pingRequest(self,self._keepAliveTimeout);
+    },self._keepAliveInterval);
+  }
+
+  function pingRequest(ws,timeout){
+    ws.ping();
+    return setTimeout(ws.close.bind(ws),timeout);
+  }
+
+  function remove(){
+    if(typeof self._keepAliveIdTimeout == "object"){
+      clearTimeout(self._keepAliveIdTimeout);
+    }
+    if(typeof self._keepAliveIdInterval == "object"){
+      clearTimeout(self._keepAliveIdInterval);
+    }
+    if(typeof self._keepAliveIntervalFunction == "function"){
+      self.removeListener("pong",self._keepAliveIntervalFunction);
+    }
+
+    self.removeListener("close",remove);
+    self._keepAliveEnabled = false;
+  };
+};
+
+/**
  * Resume the client stream
  *
  * @api public
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -17,6 +17,7 @@ This class is a WebSocket server. It is an `EventEmitter`.
   * `disableHixie` Boolean
   * `clientTracking` Boolean
   * `perMessageDeflate` Boolean|Object
+  * `keepAlive` Boolean|Object
 * `callback` Function
 
 Construct a new server object.
@@ -66,7 +67,14 @@ If `handleProtocols` is not set then the handshake is accepted regardless the va
 
 If a property is empty then either an offered configuration or a default value is used.
 
-### server.close([callback])
+### options.keepAlive
+
+`keepAlive` Makes available a ping pong protocol which allows to check if the connection is alive for all websockets. The extension is disable when false (default value).The extension is enable when an object is instead provided in parameters:
+
+* `interval` Number: the interval of time in milliseconds between a ping request and a pong response. default: 2000.
+* `timeout`  Number: the maximum waiting time between a ping request and the pong response before close the socket. default: 10000.
+
+### server.close()
 
 Close the server and terminate all clients, calls callback when done with an error if one occured.
 
@@ -107,7 +115,7 @@ This class represents a WebSocket connection. It is an `EventEmitter`.
   * `protocol` String
   * `agent` Agent
   * `headers` Object
-  * `protocolVersion` Number|String  
+  * `protocolVersion` Number|String
     -- the following only apply if `address` is a String
   * `host` String
   * `origin` String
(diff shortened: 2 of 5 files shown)


3rd-Eden:
It looks to me at a first glance.

FlorianBELLAZOUZ:
Hi ! 
So what do you think about this PR ?
This need any modifications ?

LordMajestros:
@FlorianBELLAZOUZ you should rebase this

pavan4:
Is this included in master ?

arjenvanderende:
@3rd-Eden You're probably quite busy, since I see multiple pull requests on this repository without a response. I'd really like this feature to be merged and released. If I rebase this commit and resubmit it as a new PR, would you have time to look at it and get it merged?

pdefreitas:
Seriously why this PR isn't merged?!?! +1

qristin:
+1

lib/WebSocket.js
```diff
@@ -188,6 +188,61 @@ WebSocket.prototype.pong = function(data, options, dontFailWhenClosed) {
...
+  }
+
+  function remove(){
+    if(typeof self._keepAliveIdTimeout == "object"){
+      clearTimeout(self._keepAliveIdTimeout);
+    }
+    if(typeof self._keepAliveIdInterval == "object"){
+      clearTimeout(self._keepAliveIdInterval);
+    }
+    if(typeof self._keepAliveIntervalFunction == "function"){
```
artemio: unnecessary?

ladytellur:
@FlorianBELLAZOUZ could you please please rebase this PR?

FlorianBELLAZOUZ:
Rebase Done !

sebblanc:
Great PR. I tried this on ubuntu 14.0.4, it works but as @pavan4 I can't go below ~30/40 seconds no matter the timeout or interval. Do someone have an idea why ? Probably something to configure on ubuntu but I have no idea what.

ladytellur:
Is there something wrong with this PR? What should be improved if so? @3rd-Eden @LordMajestros

john-doherty:
What happened to this? Was it ever merged?

thermogl:
This PR would be really handy, is there any update on this?

lpinca:
This PR offers great flexibility as the keep-alive can be enabled and customized per connection.

If the goal is to only find and close no longer active connections, a very simple alternative would be something like this:

```js
'use strict';

  ws.isAlive = true;
  ws.on('pong', () => ws.isAlive = true);
});

const interval = setInterval(() => {
  wss.clients.forEach((ws) => {
    if (ws.isAlive === false) return ws.terminate();

    ws.isAlive = false;
    ws.ping('', false, true);
  });
}, 30000);
```

The main advantage of this approach is that a single timer is run instead of one per connection.

john-doherty:
@lpinca have you tried this?

lpinca:
@john-doherty no but I'm confident that it works, the only gotcha is that `ws.terminate()` does not make the `close` event to be emitted immediately (#891). A possible **hacky** workaround for that is to call `socket.destroy()`.

```js
if (ws.isAlive === false) return ws._socket.destroy();
```

lpinca:
@websockets/admin what should we do with this PR? Do we want this functionality per connection?

I would prefer to add something much simpler, similar to the example I posted above, and leave this kind of customization to the user or a wrapper module.

The reason I didn't add it yet is because `clientTracking` can be disabled.

Thoughts?

iongion:
Can't this be good for both cases ? Both for sweeping inactive connections and for keep-alive purposes, while the main one being the keep alive ?
Then, `clientTracking` - _The **clientTracking** option provides access to collection of active WebSocket clients_

But is it ?  The `active` is not really **active** in the sense of this PR's keep-alive functionality.

👍  to this PR!

elithrar:
@lpinca - I can likely pick up this PR and resolve any outstanding {merge, design} issues, although the [example you posted above](https://github.com/websockets/ws/pull/635#issuecomment-281026825) is arguably a better (simpler) implementation.

Making it trivial to maintain keep-alives/send pings to the Set() of connected clients would definitely benefit package users.

Thoughts?

lpinca:
@elithrar I don't know, I agree that having a built-in keep-alive mechanism would be useful for all users but I'm not sure if it makes sense to customize/enable this per connection.

I think this level of customization is better suited for a wrapper user land class.

I prefer to have something simpler which is all or none with the added benefit of running only one timer instead of one per connection.

elithrar:
To clarify, I'm proposing a mechanism that:

* is set up once on server initialization
* automatically loops over the connected clients
* pings them every X seconds
* removes clients from the set when `ws.ping` throws

This would likely just be implicitly provided when `clientTracking: true`, but I'm open to making it an additional opt-in (e.g. `keepAlive: true` or similar).

lpinca:
I'm fine with that.

elithrar:
OK, great (and thanks for the quick replies!). I'll put together a PR shortly.

elithrar:
Working through this and trying to test (manually, at first) whether `ws.ping` throws an exception - 

e.g. inside the Server constructor itself -

```js
        this.heartbeatInterval = 20

        setInterval(() => {
            this.wss.clients.forEach(ws => {
                try {
                    console.log("heartbeat event fired")
                    ws.ping("", false, true)
                } catch (e) {
                    console.log("terminating dead client")
                    ws.terminate()
                }
            })
        }, 1000 * this.heartbeatInterval)
```

* Even if I abruptly cut off a (Node) client or call `ws.terminate()` I never see an error thrown
* The heartbeat events do stop, so the client is being cleaned up correctly

What am I missing here? Note that this implementation is more naive than the PR here, in the sake of simplicity, as we don't do the ping <-> pong dance with setTimeout/clearTimeout per-client.

lpinca:
@elithrar if the `failSilently` argument is `true`, `ws.ping()` does not throw.
Moreover, if you pull the cord on the client nothing changes on the server, technically the connection is still established on the server. There is no `readyState` change or anything like that so even if `failSilently` is `false`, `ws.ping()` does not throw.

This is why the additional flag, `isAlive`, is needed in the above example. This is set to `false` before sending a ping and to `true` when a pong is received. If `isAlive` is `false` before sending a ping, it means that a pong has not been received and the connection should be closed.

bnielsen1965:
Are the plans for a ping/pong keepAlive implementation dead?

I would like to see a ping/pong based keep alive option but I think it should  be available in both server and client WebSockets.

I am using ws for both server and client applications where the client subscribes to an event stream from the server. After a connection and subscription are established the client waits for events from the server. If the connection fails silently then the client is unaware of the need for a reconnect. The server can clean up dead connections that are discovered via the ping/pong keep alive method but this does nothing for the client which established the connection. So the client as well needs a keep alive method.

While this can be implemented at the application level it can be problematic if an application is designed to run in both a browser and node.js. Browsers are not exposing the ping() method and instead seem to implement the TCP keep alive method in the browser client. It would be nice if ws would also implement a keep alive as an option so the application can leave the connection handling up to the connection library.

lpinca:
Closing as per https://github.com/websockets/ws/pull/635#issuecomment-291424986.

corporatepiyush:
Is PING/PONG after some default interval is implemented or need to be implemented explicitly after starting WS server ?

lpinca:
@corporatepiyush it needs to be implemented explicitly.

kedicesur:
I guess @lpinca's suggesstion is very nice. I just would like add one thing. Depending on the number of open sockets, the workload of the server or whatnot one may need to be able to change the pinging interval dynamically. I suppose using a pseudo recursive `setTimeout` pattern is better than `setInterval` for this purpose. A `pingInterwal` property attached to `wss` (WebSockets server) could be used to alter the pinging interval dynamically.

```
function keepAliveSockets(wss){
  setTimeout(wsSet => { wsSet.forEach(ws => ws.isAlive ? (ws.isAlive = false, ws.ping("", false))
                                                       : ws.terminate());
                        keepAliveSockets(wss);
                      }, wss.pingInterval, wss.clients);
}

var wss = new ws.Server(Object.assign({},opts,{server: srv, clientTracking: true}));

wss.pingInterval = 10000;
keepAliveSockets(wss);
```

jacobbogers:
is this already implemented, the idea is from 2016, is this part of the codebase?

gitcatrat:
Has anyone tested `forEach` ping/pong in production?

Relatively mid-tier AWS EC2 m5.large can comfortably support 75-84,000 connections according to few Netflix talks I've seen (they're using Java). I feel like the loop of this size every 30-60 seconds could cause some issues and even if loop itself is relatively fast, there's probably a database request or something you have to do for each ghost connection.

Yeah I know, "could". Just a gut feeling.

Issue #634, opened 2015-12-16
Masking key not random per-frame


SEAPUNK:
RFC:

```
   The masking key is a 32-bit value chosen at random by the client.
   When preparing a masked frame, the client MUST pick a fresh masking
   key from the set of allowed 32-bit values.
```

The code generates a masking key, and then reuses it for each subsequent frame: https://github.com/websockets/ws/blob/master/lib/Sender.js#L192

Is this by design?

3rd-Eden:
Isn't the sender instance create per connection and therefor generating a new masked frame for each client?

SEAPUNK:
Per *connection*, it generates a new masking key. The RFC states that it should create a new masking key per frame, not connection.

3rd-Eden:
I guess thats a bug then.

Arnout Kazemier

SEAPUNK:
Alright, I'll create a PR for it.

rupamkhaitan:
Will this create any problem in terms of socket close or any packet loss for long running connection in AWS hosted machine with ELB?

SEAPUNK:
@rupamkhaitan I wouldn't think so. The Mocha tests for my PR worked (and I'll write that test I need to write to get the PR merged), and I'll run the Autobahn test suite later.

PR #636, opened 2015-12-18
Sender: Don't reuse masking key

SEAPUNK:
This creates a new masking key for each frame.

Ref: #634


3rd-Eden:
@SEAPUNK Could you add a test for this so it's not broken in future updates?

SEAPUNK:
I'll see what I can do.

PR #641, opened 2015-12-21
Provide original request object for initial connection

ivo-:
I could be useful if we have access to the request object that initiated the connection. In my use case this is a way to provide initial parameters for the client.


FlorianBELLAZOUZ:
Nice ! :+1:

FlorianBELLAZOUZ:
Document it, and test it !

3rd-Eden:
It already exists on the received client object/connection; https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L534

3rd-Eden:
Which is also, already documented; https://github.com/websockets/ws/blob/master/doc/ws.md#websocketupgradereq

PR #654, opened 2016-01-05
Use Buffer.concat() instead of private concatBuffers method

JacksonTian:
Buffer.concat(buffers) works fine.


3rd-Eden:
Thanks for your contribution, this will end up in the minor release.

PR #657, opened 2016-01-07
moved utf-8-validate to normal dep

wesleytodd:
This module is installed as a `devDependency` but is required at runtime.  This just moves it over so it is installed when run with `NODE_ENV=production npm install`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,18 +21,18 @@
     "test": "make test"
   },
   "dependencies": {
+    "bufferutil": "1.2.x",
     "options": ">=0.0.5",
-    "ultron": "1.0.x"
+    "ultron": "1.0.x",
+    "utf-8-validate": "1.2.x"
   },
   "devDependencies": {
     "ansi": "0.3.x",
     "benchmark": "0.3.x",
-    "bufferutil": "1.2.x",
     "expect.js": "0.3.x",
     "mocha": "2.3.x",
     "should": "8.0.x",
-    "tinycolor": "0.0.x",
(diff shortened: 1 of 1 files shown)


wesleytodd:
See here for where it is included: https://github.com/websockets/ws/blob/7debd827d39925650bba5a270811ed82616f0eb1/lib/Validation.js#L10

JacksonTian:
:-1:

wesleytodd:
@JacksonTian Why?

wesleytodd:
Also, just got an email from my team, `bufferutil` is also causing build failures for us when we npm install in production.  Moved that one over to.

JacksonTian:
The C++ addon modules caused a lot install issues.

JacksonTian:
You can see more detail : https://github.com/websockets/ws/commit/49b11093e9a009e5305dcde7003d3a896b2811dc

wesleytodd:
That makes a lot more sense.  Thanks for linking that.

The problem is that with `browserify` bundling `socket.io-client -> engine.io-client -> ws` we end up with a build error from this module.  And since this is ACTUALLY a run time dependency it SHOULD be specified as a dependency, despite the try/catch around it.

wesleytodd:
Ok, so there is clearly no way you should support the possible build tool combinations that might arise.  So I am retracting my statement from before.  The more I think about it this is `engine.io`'s issue for not bumping their major version.  I am going to close this because I can see where you are coming form with this and your solution to it.  Sorry for bothering you!  Good job on the package even though I never knew I was using it :)

PR #663, opened 2016-01-11
package.json: add bufferutil and utf-8-validate

rstacruz:
It seems this library needs [bufferutil][] and [utf-8-validate][].

> Error: Cannot find module 'bufferutil' from 'xxx/node_modules/ws/lib'
> Error: Cannot find module 'utf-8-validate' from 'xxx/node_modules/ws/lib'

They're defined in package.json as `devDependencies`, not `dependencies`. This fixes that.

[bufferutil]: https://www.npmjs.com/package/bufferutil
[utf-8-validate]: https://github.com/websockets/utf-8-validate

As seen in [browserify-hmr](https://www.npmjs.com/package/browserify-hmr) (cc @AgentME).

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,18 +21,18 @@
     "test": "make test"
   },
   "dependencies": {
+    "bufferutil": "1.2.x",
     "options": ">=0.0.5",
-    "ultron": "1.0.x"
+    "ultron": "1.0.x",
+    "utf-8-validate": "1.2.x"
   },
   "devDependencies": {
     "ansi": "0.3.x",
     "benchmark": "0.3.x",
-    "bufferutil": "1.2.x",
     "expect.js": "0.3.x",
     "mocha": "2.3.x",
     "should": "8.0.x",
-    "tinycolor": "0.0.x",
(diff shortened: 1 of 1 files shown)


rstacruz:
okay, hmm... i see there are fallbacks in case they fail to require. maybe the better solution is to define `browser` fields in bufferutil and utf-8-validate?

(I'm using ws with browserify.)

STRML:
This problem is caused by https://github.com/websockets/ws/commit/e54d45fbab3b19c2940e9057ce1e7b8f105873e0, which I've attempted to revert in #661, which will hopefully be reopened and merged.

halcwb:
I installed today, but got had to install both `utf-8-validate` as `bufferutil` to got things running.

rstacruz:
don't install those. the solution is to not use `ws` with browserify (or any of the modules that depend on `ws`).

This was a very long debate here: https://github.com/websockets/ws/commit/e54d45fbab3b19c2940e9057ce1e7b8f105873e0

halcwb:
Got, thanks

PR #650, opened 2015-12-31
Add maxpayload limit to hybi clients

LordMajestros:
Limit the amount of data that can be sent to a websocketserver from a
hybi client.
Raise a 1009 - message too big error.
Tests included.

Borrows some ideas and code from https://github.com/websockets/ws/pull/574/files

PR #668, opened 2016-01-16
[Cleanup] Added maximumPayload option for non-hixie protocols

jondubois:
This is a cleanup of https://github.com/websockets/ws/pull/520 - It seems that patch was almost ready except that it changed the ```Receiver``` constructor signature in such a way that it broke backwards compatibility (and the tests ;p).

I just changed the order of arguments to get the tests working again and then rebased the branch on the latest master to resolve conflicts.


jondubois:
I just noticed this PR: https://github.com/websockets/ws/pull/650 - It seems to be well tested - l'll leave someone else to decide which is better. Feel free to ignore this PR.

PR #671, opened 2016-01-17
Minor tweaks to tests for robustness

infinity0:
- Allow use of a custom/system installation of mocha for tests
- Use higher ports for tests, less likely to conflict with actual services

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 ALL_TESTS = $(shell find test/ -name '*.test.js')
 ALL_INTEGRATION = $(shell find test/ -name '*.integration.js')
+MOCHA = ./node_modules/.bin/mocha
 
 all:
 	node-gyp configure build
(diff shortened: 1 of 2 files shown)


infinity0:
The tests are passing on node 4 and 5, fail on node 0.12 for some reason. Possibly node 0.12 is using ports 58000-5800x? The ideal solution would be to pick a port dynamically.

3rd-Eden:
It makes no sense to merge this if starts breaking test suite for no reason, while the current test suite and ports work just fine.

infinity0:
It *doesn't* work fine on certain setups, i.e. if you have a service running on port 8008 already (e.g. Apple CalendarServer). I was going to re-patch it to use port 48008 instead of 58008 but had better things to do in the meantime.

lpinca:
Closing this as we dropped support for Node.js < 4.

PR #673, opened 2016-01-22
Added broadcast

Accesse:
Included broadcast in WebSocketServer.js

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -102,6 +102,26 @@ function WebSocketServer(options, callback) {
 
 util.inherits(WebSocketServer, events.EventEmitter);
 
+
+/**
+ * Broadcast to all Clients of Webserver
+ */
+ 
+ WebSocketServer.prototype.broadcast = function(msg) {
+  // broadcast to all clients
+  var error = null;
+  try {
+    for (var i = 0, l = this.clients.length; i < l; ++i) {
+      this.clients[i].send(msg);
+    }
+  }
+  catch (e) {
+    error = e;
+  }
+}
+
+
+
 /**
  * Immediately shuts down the connection.
  *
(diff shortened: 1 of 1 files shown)


lib/WebSocketServer.js
```diff
@@ -102,6 +102,26 @@ function WebSocketServer(options, callback) {
...
+ 
+ WebSocketServer.prototype.broadcast = function(msg) {
+  // broadcast to all clients
+  var error = null;
+  try {
+    for (var i = 0, l = this.clients.length; i < l; ++i) {
+      this.clients[i].send(msg);
+    }
+  }
+  catch (e) {
```
JacksonTian: one error will break other clients

Accesse: @JacksonTian That's right. what if we remove the try/catch?

JacksonTian: try catch the `this.clients[i].send(msg);`

Accesse: @JacksonTian  Like that?
```
WebSocketServer.prototype.broadcast = function(msg) {
 // broadcast to all clients
 var error = null;
   
   for (var i = 0, l = this.clients.length; i < l; ++i) {
       try {
           this.clients[i].send(msg);
       }
       catch (e) {
              error = e;
       }
   }

}
```

lib/WebSocketServer.js
```diff
@@ -102,6 +102,27 @@ function WebSocketServer(options, callback) {
...
+  var error = null;
+  for (var i = 0, l = this.clients.length; i < l; ++i) {
+    try {
+      this.clients[i].send(msg);
+    }
+    catch (e) {
+      error = e;
+    }
+  }
+
```
JacksonTian: how do you process the error?

3y3:
Why we expect this in `ws` lib?
There is no access to private fields, so each user can implement this by himself.
`ws` should not to be `socket.io`

lpinca:
I agree with @3y3.

3rd-Eden:
This does not belong in a core library like ws, there are 1000's of ways to implement broadcasting and I don't want to force this implementation or any other upon users. The README already includes a simple broadcasting example as well.

PR #682, opened 2016-02-13
Fix DoS vulnerable, fix #679

### LINKED ISSUE #679: DoS exploit that crashes server
Hi.
27 days ago i created issue #669 but no action there, so i making new one with new title and one bug in it.
More info you can find in #669 `crash 1`

server.js
```javascript
var WebSocketServer = require('ws').Server
    , wss = new WebSocketServer({ port: 8080 });

    ws.onerror = function(error) {
        console.log('onerror called');
    };
});
```

client.js
```javascript
var socket = new WebSocket('ws://127.0.0.1:8080/');

socket.onopen = function() {
    console.log('onopen called');
    socket._socket.write(new Buffer([5]));
    socket.send('');
};

socket.onclose = function(event) {
    console.log('onclose called');
};

socket.onerror = function(error) {
    console.log('onerror called');
};
```
This code will crash server and the problem is that there is no API to prevent this crash.
In #669 i have same problem with client, so can we also have fix for client-side crash?
Thank you.

pulviscriptor:
Please check if i did it correctly, because i am not sure.

### DIFF
--- a/lib/Receiver.hixie.js
+++ b/lib/Receiver.hixie.js
@@ -154,7 +154,7 @@ Receiver.prototype.parse = function() {
 
 Receiver.prototype.error = function (reason, terminate) {
   this.reset();
-  this.onerror(reason, terminate);
+  this.onerror(new Error(reason), terminate);
   return this;
 };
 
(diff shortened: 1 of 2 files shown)


LordMajestros:
Nice, this is what I did. Add a test perhaps?

pulviscriptor:
@LordMajestros 
Oops. Didn't saw your changes. 
Can you also add fix to `lib/Receiver.hixie.js`

LordMajestros:
I always disable hixie clients but I've added the fix to Receiver.hixie all the same.

Issue #687, opened 2016-02-19
Hixie-76 doesn't work behind a proxy (e.g. HAProxy)


agilethomas:
This issue has been reported and fixed on other web socket implementations, but not this one.

When the socket server is behind a proxy (in my case HAProxy), the server waits for the nonce before replying with the response header, but HAProxy won't send the extra data until the server responds.

Same issue reported on other libraries:

https://github.com/ninenines/cowboy/issues/73

https://github.com/faye/faye-websocket-node/issues/4

lpinca:
Closing as Hixie-76 is no longer supported.

PR #688, opened 2016-02-19
Fix for #687 Hixie-76 doesn't work behind HAProxy

agilethomas:
- Solution is to send headers before receiving the nonce


3rd-Eden:
Could you add a test for this?

agilethomas:
I'm not sure exactly how to test for the scenario in which an HAProxy server sits between the client and the server. The existing test for Hixie-76 connection passed, so I didn't break anything. Do you have any suggestions on how to test for the issue that I fixed?

3rd-Eden:
I was just thinking of a test that just confirms that headers are send before the nonce. But I'm just going to merge this in as I want it in the next release.

Issue #441, opened 2015-01-22
Add support for binaryType = "arraybuffer"


gigadude:
It would help browser/node code reuse if binary transactions through ArrayBuffers were supported.

At least checking for `binaryType` being set to `'arraybuffer'` and flagging it as unsupported would be nice.

JoshuaWise:
+1

PR #698, opened 2016-03-15
lib/WebSocket.js: expose binaryType property

rib:
The W3C WebSocket interface includes a binaryType attributes for a
socket and a corresponding BinaryType { 'blob', 'arraybuffer' } enum.
This enables support for the 'arraybuffer' BinaryType, but avoiding
changing the existing behaviour the default binaryType is set as
'nodebuffer'. The 'blob' type is left unsupported.

This improves ws compatibility with some existing W3C WebSocket based
client code.

Fixes: #441

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -71,6 +71,7 @@ function WebSocket(address, protocols, options) {
   this.readyState = null;
   this.supports = {};
   this.extensions = {};
+  this._binaryType = 'nodebuffer';
 
   if (Array.isArray(address)) {
     initAsServerClient.apply(this, address.concat(options));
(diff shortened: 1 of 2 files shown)


3rd-Eden:
LGTM. Thanks a lot for your contribution.

lpinca:
This makes primus tests fail.

3rd-Eden:
@lpinca Why would it make it fail though?

lpinca:
The issue is actually in Primus as we are setting [`arraybuffer`](https://github.com/primus/primus/blob/1bc071f2fdc1ac2229338630e2148cf991c64b36/transformers/websockets/client.js#L70) also for the Node client. I guess we should fix this in Primus.

Issue #672, opened 2016-01-17
Error event called without Close event


pulviscriptor:
Code i use:
```javascript
var WebSocket = require('ws'); // remove this for browser run
var socket = new WebSocket('ws://1.2.3.4:5/');

socket.onopen = function() {
    console.log('onopen called');
};

socket.onclose = function(event) {
    console.log('onclose called');
};

socket.onerror = function(error) {
    console.log('onerror called');
};
```
- Node.JS v0.12.2 + ws 1.0.1

Result:
```
onerror called
```
- Firefox ESR 38.5.2
- SRWare Iron 46.0.2450.0 (chromium based)
- Internet Explorer 11

Result:
```
onerror called
onclose called
```
I assume change this behavior is very "painful" with so many users of library.
But sometimes <b>ws</b> do call `onerror` and then `onclose`. Is there some "algorithm" when `onclose` called after `onerror` and when not? 
I want to emit `disconnect` events in my library even when connection failed, like browsers do.
I tried to find places in source code, but its too hard for me (for my experience and knowledge).

misobytes:
Hi! We're noticing the same behavior using our node.js app.

I can see there is a similar outstanding issue, opened from last year: #505 

According to [https://www.w3.org/TR/websockets/](url) it seems the expected behavior should include the close event:

> If the establish a WebSocket connection algorithm fails, it triggers the fail the WebSocket connection algorithm, which then invokes the close the WebSocket connection algorithm, which then establishes that the WebSocket connection is closed, which fires the close event

I can see in WebSocket.js:
```
var req = httpObj.request(requestOptions);

  req.on('error', function onerror(error) {
    self.emit('error', error);
    cleanupWebsocketResources.call(self, error);
  });
```
... that there is an error listener for the http request responsible for establishing the socket connection.

However, since the ```readyState === WebSocket.CONNECTING``` a close event is intentionally not emitted:

``` var emitClose = this.readyState !== WebSocket.CONNECTING; ```

I'm guessing the above statement exists for a reason. Can anyone explain the rationale behind it?

It would be possible to emit the close event when an error is detected during connection establishment if we:
* Removed the conditional check ```this.readyState !== WebSocket.CONNECTING```.
* Checked the type of error within the request error handler to see if it contains ENOTFOUND or similar. If it does contain such an error then explicitly emit a close event or set the ```readyState``` to something other than ```WebSocket.CONNECTING```.

I would be happy to open a PR once I have a little more background information. 

Thanks!

[...]

Issue #505, opened 2015-05-21
'close' event isn't fired when a connection failed


flowersinthesand:
How to reproduce the issue:
```javascript
var ws = new (require('ws'))('ws://no.where'); // 0.7.2
ws.onerror = function() {
    console.log('error');
};
ws.onclose = function() {
    console.log('closed');
};
```

Output:
```
error
```

Expected:
```
error
closed
```

As the [API doc](https://github.com/websockets/ws/blob/master/doc/ws.md#websocketonclose) says, `websocket.onclose` emulates the W3C Browser based WebSocket interface so it would be nice if it follows other browsers' behavior where it fires `error` and `close` event together in such cases.

How to reproduce the issue in browser:
```javascript
var ws = new WebSocket('ws://no.where');
ws.onerror = function() {
    console.log('error');
};
ws.onclose = function() {
    console.log('closed');
};
```

Chrome output:
```
WebSocket connection to 'ws://no.where/' failed: Error in connection establishment: net::ERR_NAME_NOT_RESOLVED
error
closed
```

Firefox output:
```
Firefox can't establish a connection to the server at ws://no.where/.
"error"
"closed"
```

IE output:
```
SCRIPT12007: WebSocket Error: Network Error 12007, The server name or address could not be resolved
error
closed
```

I would like to contribute if it's right that it should fire `error` and `close` event in such cases. Let me know that.

3rd-Eden:
Fixed in master.

PR #706, opened 2016-04-06
Emit both an error and close event when connection establishment fails

misobytes:
Hi,

It would be foolish of me to think that the following statement in [WebSocket.js](https://github.com/websockets/ws/blob/master/lib/WebSocket.js) existed for no reason:

```var emitClose = this.readyState !== WebSocket.CONNECTING;```

But this is preventing the ```close``` event from being emitted when the initial HTTP handshake fails. It also creates a discrepancy in behavior between the browser vs a node.js app.

According to [WHATWG](https://html.spec.whatwg.org/multipage/comms.html#the-websocket-interface) and [W3C](https://www.w3.org/TR/websockets/):

> If the establish a WebSocket connection algorithm fails, it triggers the fail the WebSocket connection algorithm, which then invokes the close the WebSocket connection algorithm, which then establishes that the WebSocket connection is closed, which fires the close event

As an FYI, there are a couple of open issues regarding this matter: #672 #505 

I created this PR to find out the rationale behind the decision to suppress the ```close``` event and to address the difference in behavior between the browser and node.js apps.

Let me know your thoughts.

Thanks!

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -936,21 +936,18 @@ function sendStream(instance, stream, options, cb) {
 function cleanupWebsocketResources(error) {
   if (this.readyState === WebSocket.CLOSED) return;
 
-  var emitClose = this.readyState !== WebSocket.CONNECTING;
   this.readyState = WebSocket.CLOSED;
 
   clearTimeout(this._closeTimer);
   this._closeTimer = null;
 
-  if (emitClose) {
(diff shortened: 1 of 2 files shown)


3rd-Eden:
Thanks for your contribution. I'll make a new minor release with this included as this could break code user code who do not expect both events to be emitted.

misobytes:
Thanks @3rd-Eden ! This is great news. :)

PR #712, opened 2016-04-21
fix WebSocketServer options' judgement

JiaJiaJiang:
there is a logic mistake to judge if port or server is provided
I change it as the following logic:
if noServer not true,look for port or server in the options.If both of them are not in the options,throw TypeError

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -40,7 +40,7 @@ function WebSocketServer(options, callback) {
     maxPayload: null
   }).merge(options);
 
-  if (!options.isDefinedAndNonNull('port') && !options.isDefinedAndNonNull('server') && !options.value.noServer) {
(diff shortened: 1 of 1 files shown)


lib/WebSocketServer.js
```diff
@@ -40,7 +40,7 @@ function WebSocketServer(options, callback) {
     maxPayload: null
   }).merge(options);
 
-  if (!options.isDefinedAndNonNull('port') && !options.isDefinedAndNonNull('server') && !options.value.noServer) {
+  if (!options.value.noServer && (!options.isDefinedAndNonNull('port') || !options.isDefinedAndNonNull('server'))) {
```
lpinca: It doesn't seem correct to me. For example:
```js
const http = require('http');
const ws = require('ws');

const wss = new ws.Server({ server });

...
```
throws with this change but it shouldn't.

JiaJiaJiang: This means we should not make the Server which is listing on a port as the server option?

lpinca: No problem.

Issue #612, opened 2015-11-08
wscat: unsupported protocol version


thauer:
I just installed ws on ubuntu 15.04 (with `npm install -g ws`) and when running `$ wscat -c ws://echo.websocket.org -p 8`,  I get an exception from 

```
  if (options.value.protocolVersion !== 8 && options.value.protocolVersion !== 13) {
    throw new Error('unsupported protocol version');
  }
```

The reason being that at this point, when specified on the command line, `options.value.protocolVersion` seems to be a `String`, rather than a `Number`.

janpieper:
I have the same problem.

`options.value.protocolVersion` is a string, not a number!

Ugarz:
Right, thanks to @thauer and @janpieper this worked for me !
In file located at `/usr/local/lib/node_modules/wscat/node_modules/ws/lib/WebSocket.js:568:11` (i'm using a mac).
Replace the if statement line 567 by this (just add `"`around numbers):
```javascript
if (options.value.protocolVersion !== "8" && options.value.protocolVersion !== "13") {
    throw new Error('unsupported protocol version');
  }
```

oblalex:
I also confirm this.

lpinca:
Fixed in https://github.com/websockets/wscat/commit/4de45ae56f817a37af5a2a689b7e1c59f748e441.

PR #713, opened 2016-04-21
fix wscat protocol related to issue #612

Ugarz:
Right, thanks to @thauer and @janpieper I can now use wscat !
In file located at `/lib/WebSocket.js:568:11`, replace the if statement line 567 by this :

``` javascript
if (options.value.protocolVersion !== "8" && options.value.protocolVersion !== "13") {
    throw new Error('unsupported protocol version');
  }
```

By adding `"` around numbers, you are now waiting for a `string` and not an `number`.

PS : it is my first pull request, so please let me know if I'm doing something wrong :)

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -589,7 +589,7 @@ function initAsClient(address, protocols, options) {
     localAddress: null
   }).merge(options);
 
-  if (options.value.protocolVersion !== 8 && options.value.protocolVersion !== 13) {
+  if (options.value.protocolVersion !== "8" && options.value.protocolVersion !== "13") {
     throw new Error('unsupported protocol version');
   }
 


lpinca:
It's probably better to fix this in `wscat`.
Changing this line https://github.com/websockets/wscat/blob/2a66bd6eebde1b7e6b6803e05fe6e0edc2bb1432/bin/wscat#L178 to

```js
if (program.protocol) options.protocolVersion = +program.protocol;
```

Should be sufficient. Would you mind opening a PR on `wscat` [repo](https://github.com/websockets/wscat/) with this change?

Thank you.

lpinca:
Fixed in https://github.com/websockets/wscat/commit/4de45ae56f817a37af5a2a689b7e1c59f748e441.

Issue #465, opened 2015-03-07
Send message to everyone except sender


cra0kalo:
Hey guys just started using ws how would I only send or broadcast a message from a client to all other clients except the sender. Theres very little example code I couldn't find anything.

		//when the client sends a message  
        ws.on("message", function(message) {
                console.log("Received: %s", message);
                server.broadcast(message);

my broadcast method has a for each loop currently

        server.clients.forEach(function(client) {		
                client.send(data);
        }

glennschler:
You may want to look at Socket.io which enhances WebSockets with an
implementation which allows for broadcasting.

glennschler:
You seem to be doing it correctly. Though, Socket.io already enhances this
WebSocket implementation with additional support for broadcasting.

3rd-Eden:
@glennschler are you seriously advising socket.io because there a single if statement missing in the `forEach` code? I don't even.. 

```js
ws.on("message", function(message) {
  server.clients.forEach(function(client) {       
    if (client !== ws) client.send(data);
  });
});
```

@cra0kalo The only thing that your code was missing was a simple `if` statement to check if the client that you received from the forEach was not the same as the `ws` client that received the message.

glennschler:
@3rd-Eden no, I thought that this project is a pure RFC-6455 implementation of Websocket. So if a user implements broadcasting themselves, then this project is the best option. If someone is asking if broadcasting is an implemented feature of this API, then I think that is outside of the spec and is one of the reasons why engine.io/socket.io exists. I realize now this issue was not directly asking to modify the API, so please excuse my suggestion.

cra0kalo:
Thankyou 3rd-Eden got it working sir

PR #729, opened 2016-05-12
added broadcast to everyone else

dirkk0:
https://github.com/websockets/ws/issues/465

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,18 @@ wss.broadcast = function broadcast(data) {
   });
 };
 ```
+### Server sending broadcast data to everyone else
+
+```js
+var WebSocketServer = require('ws').Server
+  , wss = new WebSocketServer({ port: 8080 });
+
+wss.broadcast = function broadcast(data) {
+  wss.clients.forEach(function each(client) {
+    if (client !== ws) client.send(data);
+  });
+};
+```
 
 ### Error handling best practices
 
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ var WebSocketServer = require('ws').Server
 
 wss.broadcast = function broadcast(data) {
   wss.clients.forEach(function each(client) {
-    // use this, if you want to send to everyone else
+    // use this, if you want to send to everyone except the current webcocket (ws)
     // if (client !== ws) client.send(data);
     client.send(data);
   });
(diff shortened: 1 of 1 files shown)


lpinca:
@dirkk0 thank you for this patch.
I find the example a bit redundant as there is a broadcast to all example just above it.

What do you think about adding a sentence to the broadcast to all example explaining that if you want to exclude the sender you have to filter it out from the clients loop?

dirkk0:
Hi, thanks for your response! Sure, we could do this also this way, but in
most of my work the exclusion is rather the default, no?
So maybe just one example, with the filter in, commented out with a comment
saying 'Uncomment this to exclude the sender'?

lpinca:
It makes sense. 👍

dirkk0:
ok, it's a micro patch, but may help.

README.md
```diff
@@ -130,6 +130,8 @@ var WebSocketServer = require('ws').Server
 
 wss.broadcast = function broadcast(data) {
   wss.clients.forEach(function each(client) {
+    // use this, if you want to send to everyone else
+    // if (client !== ws) client.send(data);
```
lpinca: The only thing I'm not sure is that the reader may wonder what `ws` is as it isn't defined in this example, otherwise looks good.

dirkk0: good point.

dirkk0:
I changed it accordingly.

lpinca:
@dirkk0 sorry to pester but what do you think about changing it like this?
```js
var WebSocketServer = require('ws').Server
  , wss = new WebSocketServer({ port: 8080 });

// Broadcast to all.
wss.broadcast = function broadcast(data) {
    client.send(data);
  });
};

  ws.on('message', function message(data) {
    // Broadcast to to everyone else.
      if (client !== ws) client.send(data);
    });
  })
});
```

dirkk0:
You are not pestering me at all -  it's me who is wasting your time! :-)
This is a great solution, please go for it.
Thanks!

lpinca:
Update the PR when you have time, so we can this merged :)
Thanks.

lpinca:
Merged as 78425d0c3360acd67280f33cab89c84a46d3eb0d.
I hope you don't mind not having "purple merged status" on this PR.
Thanks!

dirkk0:
not at all, thx!

PR #733, opened 2016-05-30
Permessage-deflate calls zlib flush with wrong flush level

joux3:
I started wondering why very similar packets didn't compress better. Turns out the Node.js API for `zlib` is a tad bit misleading: the kind of flush needs to be specified when calling `flush` even if it's been specified in the options (see https://nodejs.org/api/zlib.html#zlib_zlib_flush_kind_callback). 

This pull requests calls `flush` with the correct level and adds a test that checks that the compression actually makes similar packets smaller.

Note: I gave this only a very preliminary testing. Please do testing across more browsers before even considering merging.

### DIFF
--- a/lib/PerMessageDeflate.js
+++ b/lib/PerMessageDeflate.js
@@ -7,6 +7,7 @@ const DEFAULT_WINDOW_BITS = 15;
 const DEFAULT_MEM_LEVEL = 8;
 const TRAILER = Buffer.from([0x00, 0x00, 0xff, 0xff]);
 const EMPTY_BLOCK = Buffer.from([0x00]);
+const FLUSH_LEVEL = zlib.Z_SYNC_FLUSH;
 
 /**
  * Per-message Compression Extensions implementation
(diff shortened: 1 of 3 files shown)


JacksonTian:
It seems OK.

BenV:
Would be great to get this merged, we also noticed the poor compression ratio.

3rd-Eden:
It should really be customizable through configuration instead of setting arbitrary "better" levels. Better compression doesn't always mean better performance as it will take longer to compress the data.

3rd-Eden:
ignore that.

BenV:
I'm not against that, but in this case I think `Z_SYNC_FLUSH` is "correct", it doesn't quite make sense to restart the compression for each packet, if that is the desired behavior it can be controlled with the server/client context takeover options, right? 

Note that the flush level (https://nodejs.org/api/zlib.html#zlib_zlib_flush_kind_callback) is not the same as the compression level. I was actually thinking of doing a pull for configuring the compression level this weekend, as we would like to change ours to fastest.

More on the flush modes can be found here: http://www.bolet.org/~pornin/deflate-flush.html, in particular: `A full flush degrades the compression efficiency since it removes sequence sharing opportunities for the next 32 kB of data`

BenV:
Consider it ignored! :smile: 

Will still try to find time to do a PR for the configurable compression level in the near future, though.

lpinca:
@joux3 can you please rebase this?

joux3:
Done @lpinca. Further review for the `Sender.test.js` is needed, I just changed the asserts as the actual byte counts changed after my fix. I'm not familiar enough with the codebase and WebSocket framing to verify if that is indeed OK.

lpinca:
I merge this, it should be easy to revert/fix if we find that something isn't working as expected.

lpinca:
Thanks!

PR #734, opened 2016-06-01
Avoiding memory leaks

andrija-hers:
Here is a massive refactor that minimizes memory leaks.

There are 2 important notes here:
1. This PR will not work until https://github.com/einaros/options.js/pull/14 is accepted (the `options` module leaks memory massively)
2. One of the test still fails - just like the original code does (due to timeout of 5000 ms)

I hope that @einaros will take care of the `options` repo so that this PR may be accepted.

The code changes are pretty much self explanatory, but I'll just note that almost all the work is based on removing the notorious `var self = this` paradigm, that _always_ leads to memory leaks (via closures that never get collected as garbage).

### DIFF
--- a/lib/BufferPool.js
+++ b/lib/BufferPool.js
@@ -6,6 +6,12 @@
 
 var util = require('util');
 
+function noop () {}
+
+function defaultGrowStrategy (db, size) {
+  return db.used + size;
+}
+
 function BufferPool(initialSize, growStrategy, shrinkStrategy) {
   if (this instanceof BufferPool === false) {
     throw new TypeError("Classes can't be function-called");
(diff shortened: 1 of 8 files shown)


3rd-Eden:
Might make more sense to dump the options module completely.

3rd-Eden:
As for the `var self = this` -- We're changing to fat arrow syntax so that should be resolved in master.

einaros:
Regarding the Options pullreq, I can accept it, but at first glance it seems unnecessary. When references to an Options instance are released, everything - including the closures - should be gc'd.

JacksonTian:
@andrija-hers could you rebase it with master

andrija-hers:
@JacksonTian I'll do my best, asap

lpinca:
Is this still needed? We no longer depend on `options` and are no longer using the `var self = this;` paradigm.

lpinca:
I'm going to close this as there would be a pretty big amount of work only to rebase against master.
Thanks a lot for the effort you put here @andrija-hers.

PR #739, opened 2016-06-12
Clean up code

JacksonTian:


lib/WebSocket.js
```diff
@@ -89,7 +89,7 @@ util.inherits(WebSocket, EventEmitter);
  * Ready States
  */
 ["CONNECTING", "OPEN", "CLOSING", "CLOSED"].forEach(function each(state, index) {
-    WebSocket.prototype[state] = WebSocket[state] = index;
+    WebSocket[state] = index;
```
ericmdantas: Wouldn't this break those who use the `wsInstance.CLOSE`, etc?

JacksonTian: Maybe deprecation should be better?

ericmdantas: Well, I wouldn't deprecate it if it was really needed for whatever reason. I guess there's a few people out there who test if a client is ready to go by doing:

```
if (client.readyState === client.OPEN) {
  // send data
} 
```

I currently use the above.

JacksonTian: roll backed.

3rd-Eden:
Looks sane to me, thanks for the contribution.

PR #740, opened 2016-06-12
use bufferutil & utf-8-validate instead

JacksonTian:


3rd-Eden:
Nope, sorry can't do. 

We cannot depend on buffer util and utf-u validation directly because there environments where binary addons fail to compile causing this module's installation to fail while it would still work as there are JS fallbacks. 

That's why there's the decision to have users install bufferutil and utf-8 validation manually when they want to have binary addons for their own project.

JacksonTian:
I found the bufferutil and utf-8 validation has fallback implementations in itself's package.

3rd-Eden:
That's indeed true, but if the compilation of the binary addon fails, the whole module will not be installed. So even the fallback code will be not be installed. Thats why it's bundled in ws instead.

PR #744, opened 2016-06-13
fix PerMessageDeflate options pass issue

JacksonTian:
The `typeof options.value.perMessageDeflate !== true` always `false`.


3rd-Eden:
Thanks for the correction.

PR #745, opened 2016-06-13
refine code via eslint

JacksonTian:


3rd-Eden:
+1 on this, Node 0.12 is failing but in all honesty, i don't even want to support that anymore. So i'm gonna nuke it from travis.

Issue #716, opened 2016-04-29
hixie 76 is broken in node 6


ArtemGovorov:
There's [this breaking change in node 6](https://github.com/nodejs/node/pull/5522), so it's required to pass the `binary` encoding as a second argument to all `hash.update` calls at:
- https://github.com/websockets/ws/blob/master/lib/WebSocketServer.js#L421, 
- https://github.com/websockets/ws/blob/master/lib/WebSocketServer.js#L427,
- and also ideally at https://github.com/websockets/ws/blob/master/lib/WebSocketServer.js#L221

lpinca:
Closing as Hixie-76 is no longer supported.

PR #746, opened 2016-06-13
specify data encoding type 'binary' at hash#update

tejasmanohar:
Solves https://github.com/websockets/ws/issues/716, which originated in Node v6 due to [this breaking change](https://github.com/nodejs/node/pull/5522#discussion_r54774150).


3rd-Eden:
Thanks for the correction.

tejasmanohar:
No problem. Lmk when patch has been pushed to NPM.

PR #755, opened 2016-06-15
reduce usage of self = this

JacksonTian:


ericmdantas:
This would kill the compatibility with the versions of Node that don't support arrow functions yet, though.

JacksonTian:
We don't support the version less than v4 in the future.

ericmdantas:
Sounds good, then!

PR #786, opened 2016-07-17
Remove unnecessary assignment

lpinca:
The `messageBuffer` property is not used anywhere, so this patch removes the whole assignment.


JacksonTian:
Thank you.

PR #787, opened 2016-07-17
add another one emit event headers with path, like event connection

andreevWork:
I needed to share connections and i find it:

```
self.emit('connection' + req.url, client); 
self.emit('connection', client);
```

But i set cookie in 'headers' event and i needed same behavior with 'headers' event.

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -256,6 +256,7 @@ function handleHybiUpgrade(req, socket, upgradeHead, cb) {
     }
 
     // allows external modification/inspection of handshake headers
+    self.emit('headers' + req.url, headers, req.headers);
     self.emit('headers', headers, req.headers);
 
     socket.setTimeout(0);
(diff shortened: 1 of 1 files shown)


lpinca:
Instead of using two different events, why not using `req` as second argument for the existing one?

```js
this.emit('headers', headers, req);
```

so you can get all the required info (including the URL) from the `req` object if needed.

lib/WebSocketServer.js
```diff
@@ -256,7 +256,8 @@ function handleHybiUpgrade(req, socket, upgradeHead, cb) {
     }
 
     // allows external modification/inspection of handshake headers
-    self.emit('headers', headers);
+    self.emit('headers' + req.url, headers, req.headers);
+    self.emit('headers', headers, req.headers);
```
lpinca: Instead of using two different events, why not using `req` as 3rd argument?

```js
this.emit('headers', headers, req);
```

so you can get all the required info (including the URL) from the `req` object if needed.

PR #788, opened 2016-07-18
Remove some dead code

lpinca:
This patch removes some unreachable branches and some unnecessary function calls.
See inline comments.

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -357,7 +357,7 @@ class Receiver {
       extension.decompress(messageBuffer, fin, (err, buffer) => {
         if (this.dead) return;
         if (err) {
-          callback(new Error('invalid compressed data'));
+          callback(err.closeCode === 1009 ? err : new Error('invalid compressed data'));
           return;
         }
         callback(null, buffer);
@@ -448,19 +448,13 @@ var opcodes = {
       // decode length
       var firstLength = data[1] & 0x7f;
       if (firstLength < 126) {
-        if (this.maxPayloadExceeded(firstLength)){
-          this.error('Maximumpayload exceeded in compressed text message. Aborting...', 1009);
-          return;
-        }
+        if (this.maxPayloadExceeded(firstLength)) return;
         opcodes['1'].getData.call(this, firstLength);
       }
       else if (firstLength == 126) {
         this.expectHeader(2, (data) => {
           var length = readUInt16BE.call(data, 0);
-          if (this.maxPayloadExceeded(length)){
-            this.error('Maximumpayload exceeded in compressed text message. Aborting...', 1009);
-            return;
-          }
+          if (this.maxPayloadExceeded(length)) return;
           opcodes['1'].getData.call(this, length);
         });
       }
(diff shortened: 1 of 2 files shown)


lib/Receiver.js
```diff
@@ -471,11 +465,8 @@ var opcodes = {
             return;
           }
           var length = readUInt32BE.call(data, 4);
-          if (this.maxPayloadExceeded(length)){
-            this.error('Maximumpayload exceeded in compressed text message. Aborting...', 1009);
```
LordMajestros: Why do you remove the error calls here? This means that if the max payload is exceeded the message is ignored. Is this safe? This could be an attempt at a buffer overflow attack.

lib/Receiver.js
```diff
@@ -448,19 +448,13 @@ var opcodes = {
       // decode length
       var firstLength = data[1] & 0x7f;
       if (firstLength < 126) {
-        if (this.maxPayloadExceeded(firstLength)){
-          this.error('Maximumpayload exceeded in compressed text message. Aborting...', 1009);
```
LordMajestros: Here too

lib/Receiver.js
```diff
@@ -448,19 +448,13 @@ var opcodes = {
...
-          return;
-        }
+        if (this.maxPayloadExceeded(firstLength)) return;
         opcodes['1'].getData.call(this, firstLength);
       }
       else if (firstLength == 126) {
         this.expectHeader(2, (data) => {
           var length = readUInt16BE.call(data, 0);
-          if (this.maxPayloadExceeded(length)){
-            this.error('Maximumpayload exceeded in compressed text message. Aborting...', 1009);
```
LordMajestros: Here too

lib/Receiver.js
lpinca: Because the check is already made on line 468 in the `maxPayloadExceeded`.

lib/Receiver.js
lpinca: Same reason.

lib/Receiver.js
LordMajestros: True, I missed that. Good catch.

lib/Receiver.js
```diff
@@ -590,37 +569,33 @@ var opcodes = {
...
     finish: function(mask, data) {
-      var self = this;
       var packet = this.unmask(mask, data, true) || new Buffer(0);
       var state = clone(this.state);
-      this.messageHandlers.push(function(callback) {
-        self.applyExtensions(packet, state.lastFragment, state.compressed, function(err, buffer) {
+      this.messageHandlers.push((callback) => {
+        this.applyExtensions(packet, state.lastFragment, state.compressed, (err, buffer) => {
           if (err) {
-            if (err.type === 1009){
```
lpinca: This branch was unreachable because the error here is the one from line 360.

lib/Receiver.js
```diff
@@ -590,37 +569,33 @@ var opcodes = {
...
-            return self.error(err.message, 1007);
+            this.error(err.message, err.closeCode === 1009 ? 1009 : 1007);
+            return;
           }
+
           if (buffer != null) {
-            var length = (self.currentMessageLength + buffer.length);
-            if (self.maxPayload == 0 || (self.maxPayload > 0 && length < self.maxPayload)) {
-              self.currentMessage.push(buffer);
-              self.currentMessageLength += length;
```
lpinca: There was a bug here as length was already the new length so it shouldn't have be summed but only assigned.

LordMajestros: 👍🏽

3rd-Eden:
Thanks for your contribution.

Issue #680, opened 2016-02-13
maxPayload crash: "this.onerror is not a function" in Receiver


rkaw92:
The server code below can easily be crashed by sending a message that is over the size limit of 100 bytes (on master):

```js
var ws = require('ws');

var websocketServer = new ws.Server({
	port: 7717,
	disableHixie: true,
	clientTracking: false,
	perMessageDeflate: false,
	maxPayload: 100
});

websocketServer.on('connection', function(socket) {
	setInterval(function() {
		socket.send('some notification');
	}, 1000);
	
	socket.on('error', function() {});
});
```

The output:
```
/home/thewanderer/Devel/Node/WebSocket/bugs/maxPayload/node_modules/ws/lib/Receiver.js:331
  this.onerror(reason, protocolErrorCode);
       ^

TypeError: this.onerror is not a function
    at Receiver.error (/home/thewanderer/Devel/Node/WebSocket/bugs/maxPayload/node_modules/ws/lib/Receiver.js:331:8)
    at Receiver.<anonymous> (/home/thewanderer/Devel/Node/WebSocket/bugs/maxPayload/node_modules/ws/lib/Receiver.js:483:18)
    at Receiver.add (/home/thewanderer/Devel/Node/WebSocket/bugs/maxPayload/node_modules/ws/lib/Receiver.js:102:24)
    at Socket.realHandler (/home/thewanderer/Devel/Node/WebSocket/bugs/maxPayload/node_modules/ws/lib/WebSocket.js:801:20)
    at emitOne (events.js:90:13)
    at Socket.emit (events.js:182:7)
    at readableAddChunk (_stream_readable.js:147:16)
    at Socket.Readable.push (_stream_readable.js:111:10)
    at TCP.onread (net.js:525:20)
```

Specifically, I am using a [very talkative test client](https://gist.github.com/rkaw92/24e2ae81edcbd7e33681), which keeps sending a 100KB-long "lorem ipsum" as fast as possible.

My guess is that lib/Receiver.js:122 is the part that deletes the function.

LordMajestros:
Hello @rkaw92, dos this crash occur if maxPayload is set to 0 (zero) or omitted.
Could you please post the contents of the file ipsum.txt

LordMajestros:
[...]

PR #789, opened 2016-07-20
Make sure that no string is emitted as an error

lpinca:
This patch makes sure that the `error` listener of the `WebSocket` object always receives an error. All errors now have a meaningful stack trace.

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -191,12 +191,12 @@ class Receiver {
   processPacket (data) {
     if (this.extensions[PerMessageDeflate.extensionName]) {
       if ((data[0] & 0x30) != 0) {
-        this.error('reserved fields (2, 3) must be empty', 1002);
+        this.error(new Error('reserved fields (2, 3) must be empty'), 1002);
         return;
       }
     } else {
       if ((data[0] & 0x70) != 0) {
-        this.error('reserved fields must be empty', 1002);
+        this.error(new Error('reserved fields must be empty'), 1002);
         return;
       }
     }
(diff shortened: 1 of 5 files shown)


lib/Receiver.js
```diff
@@ -310,18 +312,9 @@ class Receiver {
    * @api private
    */
 
-  error (reason, protocolErrorCode) {
-    if (this.dead) return;
```
lpinca: This was added here https://github.com/websockets/ws/commit/913afa3de0fb9b46aab215c40e177d98988c2098#diff-ee39beab4c8223a58e34e180ba59f345 but the issue it attempted to fix (#680) was actually caused by [reentrance](https://github.com/websockets/ws/blob/913afa3de0fb9b46aab215c40e177d98988c2098/lib/Receiver.js#L462) after [cleanup](https://github.com/websockets/ws/blob/913afa3de0fb9b46aab215c40e177d98988c2098/lib/Receiver.js#L396).
This is no longer the case in master so it has been removed.

lib/WebSocket.js
```diff
@@ -870,10 +870,10 @@ function establishConnection(ReceiverClass, SenderClass, socket, upgradeHead) {
     self.close(code, data);
   };
 
-  self._receiver.onerror = function onerror(reason, errorCode) {
+  self._receiver.onerror = function onerror(error, errorCode) {
     // close the connection when the receiver reports a HyBi error code
-    self.close(typeof errorCode !== 'undefined' ? errorCode : 1002, '');
-    self.emit('error', (reason instanceof Error) ? reason : (new Error(reason)));
+    self.close(errorCode, '');
```
lpinca: `errorCode` is always specified.

lib/WebSocket.js
```diff
@@ -870,10 +870,10 @@ function establishConnection(ReceiverClass, SenderClass, socket, upgradeHead) {
...
+    self.emit('error', error);
```
lpinca: `error` is always an `Error` instance.

test/WebSocket.test.js
```diff
@@ -46,40 +46,6 @@ describe('WebSocket', function() {
       ws.should.be.an.instanceOf(WebSocket);
       done();
     });
-
-    it('should emit an error object when the receiver throws an error string', function(done) {
```
lpinca: No longer needed tests.

3rd-Eden:

PR #795, opened 2016-07-27
Allow 'backlog' option for listening socket

### LINKED ISSUE #794: Unable to set TCP backlog for WebSocketServer
As node sets the backlog limit to 511 (see https://nodejs.org/api/net.html#net_server_listen_port_hostname_backlog_callback), the net.server.listen() call lets you set the backlog parameter. Unfortunately the ws module hides this functionality.

So, in lib/WebSocketServer.js you can change this

this._server.listen(options.port, options.host, callback);

to this

this._server.listen(options.port, options.host, options.backlog, callback);

and add this to the options

    backlog: 511 (or whatever you want)

In my heavy load env, the backlog was full and this solved my issue.

It would be nice to fix this in the next ws version.

pavkriz:
This commit fixes #794

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -41,7 +41,8 @@ function WebSocketServer(options, callback) {
     disableHixie: false,
     clientTracking: true,
     perMessageDeflate: true,
-    maxPayload: 100 * 1024 * 1024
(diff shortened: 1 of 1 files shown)


JacksonTian:
Thanks for your contribution.

PR #805, opened 2016-08-06
Added method removeEventListener to WebSocket interface

hpherzog:
Hey Guys,

added a new method "removeEventListener" to the WebSocket interface, because I think it's common to have the counterpart to addEventListener.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ build
 coverage
 
 builderror.log
+
+*.iml
+.idea
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -481,6 +481,25 @@ WebSocket.prototype.addEventListener = function(method, listener) {
   }
 };
 
+WebSocket.prototype.removeEventListener = function(method, listener) {
+  var listeners = this.listeners(method);
+  for(var i = 0; i < listeners.length; i++) {
+    if(listeners[i]._listener === listener) {
+      this.removeListener(method, listeners[i]);
+    }
+  }
+};
+
+WebSocket.prototype.hasEventListener = function(method, listener) {
+  var listeners = this.listeners(method);
+  for(var i = 0; i < listeners.length; i++) {
+    if(listeners[i]._listener === listener) {
+      return true;
+    }
+  }
+  return false;
+};
+
 module.exports = WebSocket;
 module.exports.buildHostHeader = buildHostHeader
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -6,3 +6,6 @@ build
 coverage
 
 builderror.log
+
+*.iml
```
lpinca: Can you please remove these changes to `.gitignore`? Ideally these should go in your `.gitignore_global`.

lib/WebSocket.js
```diff
@@ -481,6 +481,25 @@ WebSocket.prototype.addEventListener = function(method, listener) {
   }
 };
```
lpinca: Please add JSDoc for this method.

lib/WebSocket.js
```diff
@@ -481,6 +481,25 @@ WebSocket.prototype.addEventListener = function(method, listener) {
...
+WebSocket.prototype.removeEventListener = function(method, listener) {
+  var listeners = this.listeners(method);
+  for (var i = 0; i < listeners.length; i++) {
+    if (listeners[i]._listener === listener) {
+      this.removeListener(method, listeners[i]);
+    }
+  }
+};
+
+WebSocket.prototype.hasEventListener = function(method, listener) {
```
lpinca: This is used only for testing right? If so can we remove it?
In the test we can use something like this:

```js
assert.strictEqual(ws.listeners().find((listener) => listener._listener === openListener), undefined);
```

test/WebSocket.test.js
```diff
@@ -1616,6 +1616,24 @@ describe('WebSocket', function() {
       });
     });
 
+    it('should remove added event listener', function(done) {
+      this.timeout(4000);
```
lpinca: There is no need to increase the timeout please keep the default value by removing this line.

test/WebSocket.test.js
```diff
@@ -1616,6 +1616,24 @@ describe('WebSocket', function() {
...
+        function messageListener() {}
+        ws.addEventListener('open', openListener);
+        ws.addEventListener('message', messageListener);
+        assert.equal(ws.hasEventListener('open', openListener), true);
+        assert.equal(ws.hasEventListener('message', messageListener), true);
+        ws.removeEventListener('open', openListener);
+        ws.removeEventListener('message', messageListener);
+        assert.equal(ws.hasEventListener('open', openListener), false);
+        assert.equal(ws.hasEventListener('message', messageListener), false);
+        done();
```
lpinca: Close the server before ending the test.

lpinca:
Merged as 261a91075dca2bcb00cc8ce70ca2c7bb330bf3a9.
Thanks!

maxcnunes:
@lpinca any plan publishing it?

lpinca:
@maxcnunes we are getting closer. I don't have an ETA but version 2.0.0 is not too far.

PR #811, opened 2016-08-20
increased performance of frameAndSend

JoshuaWise:
The V8 engine optimizes code on a per-function basis. But it can't optimize functions that have a `try {} catch {}` block in them. Because of this, `frameAndSend()` was not getting optimized. And I don't know about you, but I'd say that's a pretty important function for optimization.

To fix this, I moved the `try {} catch {}` block into a separate function.

Also, when converting from ArrayBuffers/TypedArrays to Buffers, the data was being copied in memory, which has a speed of _O(n)_. With this pull request, converting between the two will no longer copy the memory, which is much faster; _O(1)_.

Also, I renamed the function `getArrayBuffer ` to `getBufferFromNative ` to better describe what it actually does.

### DIFF
--- a/bench/sender.benchmark.js
+++ b/bench/sender.benchmark.js
@@ -20,6 +20,7 @@ require('./util');
  * Setup sender.
  */
 
+var sender;
 suite.on('start', function () {
   sender = new Sender();
   sender._socket = { write: function() {} };
@@ -34,7 +35,7 @@ suite.on('cycle', function () {
  * Benchmarks
  */
 
-framePacket = new Buffer(200*1024);
+var framePacket = new Buffer(200*1024);
 framePacket.fill(99);
 suite.add('frameAndSend, unmasked (200 kB)', function () {
   sender.frameAndSend(0x2, framePacket, true, false);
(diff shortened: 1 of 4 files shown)


JoshuaWise:
Disclaimer:
If there's an actual reason for the memory to be copied, please inform me.

JoshuaWise:
Can't figure out why the test is failing. Anyone have a clue?

JoshuaWise:
I ran *sender.benchmark.js*...

Before:
```
  Sender frameAndSend, unmasked (200 kB)
  3348536.80 ops/sec, 261996 times executed, benchmark took 5.1 sec.

  Sender frameAndSend, masked (200 kB)
  7371.89 ops/sec, 976 times executed, benchmark took 5.085 sec.
```

After:
```
  Sender frameAndSend, unmasked (200 kB)
  4045681.98 ops/sec, 314701 times executed, benchmark took 5.031 sec.

  Sender frameAndSend, masked (200 kB)
  7438.43 ops/sec, 976 times executed, benchmark took 5.037 sec.
```

LordMajestros:
Nice

3rd-Eden:
+1 and added as contributor 💃

PR #841, opened 2016-09-25
Cleanups

Nibbler999:
2 minor improvements in lib/Receiver.js:

Remove third argument to unmask as it is always true.
Remove readUInt16BE and readUInt32BE and use the builtin functions.


JoshuaWise:
LGTM

PR #842, opened 2016-09-25
Replace bufferSize with highWatermark.

Nibbler999:
bufferSize was removed from node in 0.9.12 so this test was ineffective

### DIFF
--- a/test/WebSocket.test.js
+++ b/test/WebSocket.test.js
@@ -928,8 +928,7 @@ describe('WebSocket', function() {
         var ws = new WebSocket('ws://localhost:' + port);
         var callbackFired = false;
         ws.on('open', function() {
-          var fileStream = fs.createReadStream('test/fixtures/textfile');
-          fileStream.bufferSize = 100;
+          var fileStream = fs.createReadStream('test/fixtures/textfile', { highWaterMark: 100 });
           ws.send(fileStream, {binary: true}, function(error) {
             assert.equal(null, error);
             callbackFired = true;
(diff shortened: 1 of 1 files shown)


test/WebSocket.test.js
```diff
@@ -928,8 +928,7 @@ describe('WebSocket', function() {
         var ws = new WebSocket('ws://localhost:' + port);
         var callbackFired = false;
         ws.on('open', function() {
-          var fileStream = fs.createReadStream('test/fixtures/textfile');
-          fileStream.bufferSize = 100;
+          var fileStream = fs.createReadStream('test/fixtures/textfile', { highWaterMark: 100 });
```
lpinca: Should be `100 * 1024` for 100kb.

test/WebSocket.test.js
```diff
@@ -953,9 +952,8 @@ describe('WebSocket', function() {
         var ws = new WebSocket('ws://localhost:' + port);
         var callbackFired = false;
         ws.on('open', function() {
-          var fileStream = fs.createReadStream('test/fixtures/textfile');
+          var fileStream = fs.createReadStream('test/fixtures/textfile', { highWaterMark: 100 });
```
lpinca: You might also want to pass the `encoding` as an option?

Nibbler999: Added

test/WebSocket.test.js
Nibbler999: Why? I'm pretty sure 100 bytes is correct here.

lpinca: Yeah my bad. It's 100 bytes not 100kb, ignore my comment :)

JoshuaWise:
Thanks!

PR #843, opened 2016-09-25
Fix and clean up parser and sender benchmarks

lpinca:
- Avoid using global variables.
- Remove unused functions from `bench/util.js`.
- Fix issues with strict mode and global variables.


JacksonTian:
LGTM

PR #844, opened 2016-09-26
Remove unreachable code

lpinca:
The `length` argument is always > 0.

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -131,10 +131,6 @@ class Receiver {
    */
 
   expectHeader(length, handler) {
-    if (length == 0) {
-      handler(null);
-      return;
-    }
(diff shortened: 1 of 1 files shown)


lib/Receiver.js
```diff
@@ -131,10 +131,6 @@ class Receiver {
    */
 
   expectHeader(length, handler) {
-    if (length == 0) {
-      handler(null);
-      return;
-    }
     this.expectBuffer = this.headerBuffer.slice(this.expectOffset, this.expectOffset + length);
```
lpinca: Unrelated but I'm not sure why slice is called with `this.expectOffset` and `this.expectOffset + length`. I think it is safe to use `0` and `length` respectively. Also I _think_ the `headerBuffer` on line 56 could be of length 8 instead of 10.

lpinca:
Closing this in favor of #849.

PR #846, opened 2016-09-26
Avoid using `Function.prototype.call()`

lpinca:
- First commit removes all occurrences of `Function.prototype.call()` in favor of direct function invocation.
- Second commit applies some style changes.

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -80,20 +80,20 @@ class Receiver {
 
   add(data) {
     if (this.dead) return;
-    var dataLength = data.length;
+    const dataLength = data.length;
     if (dataLength == 0) return;
     if (this.expectBuffer == null) {
       this.overflow.push(data);
       return;
     }
-    var toRead = Math.min(dataLength, this.expectBuffer.length - this.expectOffset);
+    const toRead = Math.min(dataLength, this.expectBuffer.length - this.expectOffset);
     fastCopy(toRead, data, this.expectBuffer, this.expectOffset);
     this.expectOffset += toRead;
     if (toRead < dataLength) {
       this.overflow.push(data.slice(toRead));
     }
     while (this.expectBuffer && this.expectOffset == this.expectBuffer.length) {
-      var bufferForHandler = this.expectBuffer;
+      const bufferForHandler = this.expectBuffer;
       this.expectBuffer = null;
       this.expectOffset = 0;
       this.expectHandler(bufferForHandler);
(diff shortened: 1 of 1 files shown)


lpinca:
It's a bit slower but the gain with this patch is not that significant.
Also there are some deoptimizations that goes away with this, check with `--trace_opt --trace-deopt`.

JoshuaWise:
I would think it's faster since direct function calls implicitly pass a `this`variable which is wasted (i.e., the function is passed 3 pieces of data, `this` and 2 arguments, rather than just `this` and 1 argument). But I've never tested it, and surely the difference is negligible, so I'll take your word for it. LGTM

JoshuaWise:
Before I merge, I'm curious about the deoptimizations. Could you paste the deoptimizations that are removed in this patch?

lpinca:
"Insufficient type feedback for combined type of binary operation" on `expectHeader`.
I should recreate the test as I've deleted it. If you can wait a bit (half an hour or so) I'll do that.

JoshuaWise:
Huh, interesting. I don't think I've ever encountered that deoptimization with `Function.prototype.call ` before. Sure I can wait for the test

lpinca:
```
[deoptimizing (DEOPT soft): begin 0x1a6a0bb40839 <JS Function expectHeader (SharedFunctionInfo 0x3c5d84dfaf09)> (opt #44) @37, FP to SP delta: 168, caller sp: 0x7fff5fbff218]
            ;;; deoptimize at 15847: Insufficient type feedback for combined type of binary operation
  reading input frame expectHeader => node=245, args=3, height=7; inputs:
```

There was another one but I can't reproduce it anymore, the test is simply creating a `Receiver` instance and adding a binary packet of 200 kb.

JoshuaWise:
Thanks for do that. It sure does bewilder me though. Do you know why your changes fix that deoptimization?

lpinca:
No, I'm not sure as it seems that `expectHeader` has nothing to do with `function.call()` but the other one was on the [`start`](https://github.com/websockets/ws/blob/da3b47a52e342dfe380f8be4ecc59b470ea31d10/lib/Receiver.js#L512) function and if I remember correctly it was also complaining about types so I tried with the direct function call.

lpinca:
This code has been the same for years so I'm sure there is room for improvements.

JoshuaWise:
Okay, well my curiosity has been satiated. I'll go ahead and merge now.

lpinca:
For completeness here is a benchmark of `Function.prototype.call()` vs direct call to show that it shouldn't harm anyway:

```js
'use strict';

const Benchmark = require('benchmark');

const suite = new Benchmark.Suite();
const context = {};

function f(a, b) {}
function g(ctx, a, b) {}

suite.add('Function.prototype.call()', function () {
  f.call(context, 1, 2);
});
suite.add('direct call', function () {
  g(context, 1, 2);
});
suite.on('cycle', function (event) {
  console.log(event.target.toString());
});
suite.on('complete', function () {
  console.log('Fastest is' + this.filter('fastest').map('name'));
});

suite.run({ async: true });
```

```
$ node -v
v6.6.0
$ node index.js
Function.prototype.call() x 43,310,991 ops/sec ±0.98% (83 runs sampled)
direct call x 61,497,935 ops/sec ±0.67% (82 runs sampled)
Fastest is direct call
```

PR #847, opened 2016-09-27
Clean up speed benchmark

lpinca:
This cleans up the speed benchmark.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "ultron": "1.0.x"
   },
   "devDependencies": {
-    "ansi": "0.3.x",
     "benchmark": "2.1.x",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -25,15 +25,13 @@
     "ultron": "1.0.x"
   },
   "devDependencies": {
-    "ansi": "0.3.x",
```
lpinca: I've removed these 2 dependencies as I don't care about the fancy colors but I will add them back if this is not wanted.

JoshuaWise:
This looks pretty good at first glance. I'll review it in full tomorrow hopefully.

3rd-Eden:
TY

PR #848, opened 2016-09-29
Improve ping and pong performance

Nibbler999:
Skip messageHandler mechanism for ping/pong packets and use premade empty ping and pong frames.

Improves the included ping benchmark from

```
ping message x 432,273 ops/sec ±1.38% (86 runs sampled)
ping with no data x 706,276 ops/sec ±1.79% (85 runs sampled)
```

to

```
ping message x 662,729 ops/sec ±0.77% (87 runs sampled)
ping with no data x 18,062,074 ops/sec ±1.18% (85 runs sampled)
```

### DIFF
--- a/bench/sender.benchmark.js
+++ b/bench/sender.benchmark.js
@@ -16,6 +16,8 @@ const suite = new benchmark.Suite();
 var sender = new Sender();
 sender._socket = { write() {} };
 
+suite.add('ping message', () => sender.ping('Hello', { mask: true }));
+suite.add('ping with no data', () => sender.ping(null, { mask: true }));
 suite.add('frameAndSend, unmasked (200 kB)', () => sender.frameAndSend(0x2, framePacket, true, false));
 suite.add('frameAndSend, masked (200 kB)', () => sender.frameAndSend(0x2, framePacket, true, true));
 suite.on('cycle', (e) => {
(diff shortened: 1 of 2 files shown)


lpinca:
I think that skipping the messageHandler mechanism might change the order of messages.

lpinca:
Order is indeed changed. Here is a test case:

```js
'use strict';

const ws = require('.');

const server = new ws.Server({
  clientTracking: false,
  port: 8080
}, () => {
  const socket = new ws('ws://localhost:8080');

  socket.on('message', (message) => console.log(message));
  socket.on('ping', () => console.log('ping'));
  socket.on('close', () => server.close());
});

server.on('connection', (socket) => {
  socket.send('hello');
  socket.ping();
  socket.close();
});
```

```shell
$ node gh-848.js 
ping
hello
```

With this said, the overhead added by the messageHandler mechanism is significant and it would be nice to have it turned off when permessage-deflate is disabled.

Nibbler999:
Does that matter? I don't think the websockets spec says that order needs to be preserved between ping and data frames. I'm sure not all implementations do so.

lpinca:
Well one of the nice features of the WebSocket protocol is that the order of messages is preserved. I'm not sure if the order should be respected with ping/pong frames but I would expect that.

PR #849, opened 2016-09-29
Replace BufferPool with a bl BufferList

Nibbler999:
This is much simpler and benchmarks similarly. Anyone have a use case where the buffer pool performs better?


lpinca:
This is slightly slower on my machine but I like this a lot.

lib/Receiver.js
```diff
@@ -80,22 +60,14 @@ class Receiver {
...
-    fastCopy(toRead, data, this.expectBuffer, this.expectOffset);
-    this.expectOffset += toRead;
-    if (toRead < dataLength) {
-      this.overflow.push(data.slice(toRead));
-    }
-    while (this.expectBuffer && this.expectOffset == this.expectBuffer.length) {
-      const bufferForHandler = this.expectBuffer;
-      this.expectBuffer = null;
-      this.expectOffset = 0;
+    if (!this.expectBytes) return;
```
lpinca: Is there a case where data can come in and `expectBytes` is 0? If so wouldn't it better to emit an error?

Nibbler999: I think it happens if the receiver gets data after being reset(). The only time I saw this was in the 'will not crash when it receives an unhandled opcode' test.

Nibbler999:
I've submitted some PRs to bl to improve performance.

lpinca:
> I've submitted some PRs to bl to improve performance.

Nice, that was one of my concerns. If there is a bottleneck in bl we have little control over it. We can only submit patches upstream and hope that they will be accepted.

3rd-Eden:
I'm +1 on these changes. The more code we can re-use from external parties the easier it will be to maintain the module. Once conflicts are resolved i'd say there's a green light for merging this.

lib/Receiver.js
```diff
@@ -110,12 +80,10 @@ class Receiver {
 
   cleanup () {
     this.dead = true;
-    this.overflow = null;
     this.headerBuffer = null;
```
lpinca: If I'm not wrong this line can be removed.

lpinca:
The only thing I don't like about this is that `bl` is a duplex stream. It may be handy but we are not using it as a stream and it may be a bit too heavy.

Each `WebSocket` is an `EventEmitter` which has a `Receiver` instance which is a `stream.Duplex` stream.

Maintaining a fork kinda defeat the purpose of using `bl` in the first place.

Thoughts?

Nibbler999:
We could integrate the parts we need into the receiver. https://github.com/Nibbler999/ws/commit/6d99f9e

lpinca:
@Nibbler999 LGTM 👍

PR #853, opened 2016-10-04
Only use messageHandlers if permessage-deflate extension is in use.

Nibbler999:
Reduces overhead while also preserving order of both ping and data messages.


lpinca:
Rubber stamp LGTM. I think the messageHandlers mechanism was always run to allow more extensions in the future (not only permessage-deflate) but I prefer to refactor later if needed an favor performance now. 👍

lib/Sender.js
```diff
@@ -41,13 +41,19 @@ class Sender extends EventEmitter {
...
-    });
-    this.flush();
+    if (this.extensions[PerMessageDeflate.extensionName]) {
+      this.enqueue([this.doClose, [dataBuffer, mask, cb]]);
+    } else {
+      this.doClose(dataBuffer, mask, cb);
+    }
+  }
+
+  doClose (data, mask, cb) {
```
lpinca: Can you add JSDoc for these new methods?

lib/Sender.js
```diff
@@ -41,13 +41,19 @@ class Sender extends EventEmitter {
...
+    this.frameAndSend(0x8, data, true, mask);
+    if (this.extensions[PerMessageDeflate.extensionName]) {
+      this.messageHandlerCallback();
+    }
+    if (typeof cb === 'function') cb();
```
lpinca: I think we can avoid checking the type of the callback, here and throughout the file.

lib/Sender.js
```diff
@@ -41,13 +41,24 @@ class Sender extends EventEmitter {
...
+    }
+  }
+
+  /**
+   * Sends a close frame.
+   *
+   * @api private
+   */
+  doClose (data, mask, cb) {
+    this.frameAndSend(0x8, data, true, mask);
```
lpinca: Unrelated but I wonder why the callback is not passed to `this.frameAndSend()` and is ignored in `ping()` and `pong()`.
Anyway I think it's better to keep it as is now preserving the current behavior and maybe change/fix this later in another PR.

PR #854, opened 2016-10-05
Simplify updateHead handling by using socket.unshift()

Nibbler999:
Simplify updateHead handling by using socket.unshift()


lpinca:
This is very elegant but I think that it's better to ignore the data that is passed along with the upgrade request.
The RFC says that data transfer should start only when the handshake is complete.

lpinca:
Anyway if maintainers want to keep the current behavior this is a nice approach.

Nibbler999:
The same code is used to handle the upgrade response, so it's worth changing regardless.

lpinca:
> The same code is used to handle the upgrade response, so it's worth changing regardless.

Yup, without the `unshift` 😉.

3rd-Eden:
I'm neutral on this change.

lib/WebSocket.js
```diff
@@ -808,35 +807,16 @@ function establishConnection(ReceiverClass, SenderClass, socket, upgradeHead) {
   ultron.on('error', cleanupWebsocketResources.bind(this));
 
   // ensure that the upgradeHead is added to the receiver
-  function firstHandler(data) {
-    if (called || self.readyState === WebSocket.CLOSED) return;
```
Nibbler999: That situation cannot arise now as we do everything in the same tick.

lpinca: The `data` event is not emitted on the same tick but I guess the `data` listener is removed when `WebSocket` is closed so it should be fine.

PR #855, opened 2016-10-10
Remove redundant check

lpinca:
If `data` is falsy, the function exits [early](https://github.com/websockets/ws/blob/cad49db8a825d99c68a9330be86e0aa590ecf53e/lib/Sender.js#L129-L134).


3rd-Eden:
+1 ty

PR #856, opened 2016-10-10
Improve performance of fallback mask and unmask functions

Nibbler999:
This simple implementation seems to be significantly faster in all benchmarks. Am I missing something?


lib/BufferUtil.fallback.js
```diff
@@ -16,36 +16,14 @@ exports.BufferUtil = {
...
-    switch (length % 4) {
-    case 3: output[offset + i + 2] = source[i + 2] ^ mask[2];
-    case 2: output[offset + i + 1] = source[i + 1] ^ mask[1];
-    case 1: output[offset + i] = source[i] ^ mask[0];
-    }
-    /* eslint-enable no-fallthrough */
   },
   unmask: function(data, mask) {
-    var maskNum = mask.readUInt32LE(0, true);
     var length = data.length;
```
lpinca: @Nibbler999 can you use `data.length` directly in the for loop? This variable is not very useful.

Nibbler999: It's there because of a regression in node I encountered when working on this - nodejs/node#9006

lpinca: Oh, ok.

lpinca: Maybe add a "remove when fixed" comment?

lpinca:
Apart from the comment above (add a reference to the node issue) this LGTM, big +1

PR #860, opened 2016-10-12
Use semistandard

### LINKED ISSUE #851: Use semistandard
I wonder if it's better to use a popular set of rules like [semistandard](https://github.com/Flet/semistandard) instead of building a `ws` specific `eslint` config. It will be easier to maintain:

- Install `eslint-config-semistandard`.
- Update `.eslintrc` to be just `{"extends": "semistandard"}`.
- Profit.

lpinca:
- Make `eslint` use `eslint-config-semistandard` instead of a custom configuration.
- Update code style in accordance with the new rules.

Closes #851.


3rd-Eden:
Ty <3

PR #861, opened 2016-10-13
Remove Sender inheritance from EventEmitter

lpinca:
The only event emitted by `Sender` is `error` so I think it makes sense to not make it inherit from `EventEmitter` and only use an `onerror` function instead.

This is also consistent with the `Receiver` class.


JacksonTian:
LGTM

PR #862, opened 2016-10-14
Replace BufferPool with a buffer list

Nibbler999:
Replaces #849


lpinca:
I would say go ahead and merge it.

PR #867, opened 2016-10-19
Remove messageHandler mechanism in Receiver

Nibbler999:
We don't need this at all - we can pull from the receiver's buffer when we are ready for the next frame.

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -36,10 +36,8 @@ class Receiver {
     this.expectHandler = null;
     this.currentMessage = [];
     this.currentMessageLength = 0;
-    this.messageHandlers = [];
     this.expectData(2, this.processPacket);
     this.dead = false;
-    this.processing = false;
 
     this.onerror = function () {};
     this.ontext = function () {};
@@ -60,12 +58,21 @@ class Receiver {
 
   add (data) {
     if (this.dead) return;
-    if (!this.expectBytes) return;
 
     this.buffers.push(data);
     this.bufferedBytes += data.length;
 
-    while (this.expectBytes && this.expectBytes <= this.bufferedBytes) {
+    this.process();
+  }
+
+  /**
+   * Check buffer for data.
+   *
+   * @api private
+   */
+
+  process () {
+    if (this.expectBytes && this.expectBytes <= this.bufferedBytes) {
       var bufferForHandler = this.readBuffer(this.expectBytes);
       this.expectBytes = 0;
       this.expectHandler(bufferForHandler);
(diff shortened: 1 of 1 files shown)


Nibbler999:
Benchmarks results are better than I expected

```
ping message x 340,331 ops/sec ±1.52% (82 runs sampled)
ping with no data x 484,084 ops/sec ±1.75% (83 runs sampled)
close message x 420,477 ops/sec ±5.79% (74 runs sampled)
masked text message x 258,534 ops/sec ±1.24% (89 runs sampled)
binary data (125 bytes) x 292,561 ops/sec ±1.29% (90 runs sampled)
binary data (65535 bytes) x 135,557 ops/sec ±0.93% (87 runs sampled)
binary data (200 kB) x 62,863 ops/sec ±1.13% (84 runs sampled)
```
```
ping message x 910,546 ops/sec ±1.11% (83 runs sampled)
ping with no data x 4,514,044 ops/sec ±1.00% (89 runs sampled)
close message x 3,476,559 ops/sec ±1.70% (88 runs sampled)
masked text message x 644,002 ops/sec ±0.99% (88 runs sampled)
binary data (125 bytes) x 786,783 ops/sec ±1.66% (86 runs sampled)
binary data (65535 bytes) x 203,325 ops/sec ±1.41% (87 runs sampled)
binary data (200 kB) x 81,099 ops/sec ±0.91% (89 runs sampled)
```

3rd-Eden:
Nice <3

lib/Receiver.js
```diff
@@ -345,18 +329,70 @@ class Receiver {
...
+   * Handles compressed data.
+   *
+   * @api private
+   */
 
-/**
- * Buffer utilities
- */
+  handleDataCompressed (packet) {
+    this.applyExtensions(packet, this.state.lastFragment, (err, buffer) => {
```
lpinca: Can we use `extension.decompress()` directly instead of wrapping it in `applyExtensions()`?

lib/Receiver.js
```diff
@@ -36,10 +36,8 @@ class Receiver {
     this.expectHandler = null;
     this.currentMessage = [];
     this.currentMessageLength = 0;
-    this.messageHandlers = [];
     this.expectData(2, this.processPacket);
     this.dead = false;
-    this.processing = false;
 
     this.onerror = function () {};
```
lpinca: Unrelated to this change but I think it makes sense to initialize these handlers to `null` or make them point to the same `noop` function. It doesn't make sense to use 6 different `noop` functions.

I think it's fine to do this in this PR without opening a new one.

lpinca:
LGTM. We should just make sure that there isn't a performance regression when permessage-deflate is enabled.

Nibbler999:
Tested bench/speed.js with perMessageDeflate enabled - no significant difference with this change.

PR #870, opened 2016-10-21
Add typescript type definitions

RamIdeas:
This pull request adds typings for typescript, adapted from the docs.

This would allow not only typescript developers to have intellisense within their editors, but also javascript developers, within VS Code with no extra work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "author": "Einar Otto Stangvik <einaros@gmail.com> (http://2x.io)",
   "name": "ws",
   "description": "simple to use, blazing fast and thoroughly tested websocket client, server and console for node.js, up-to-date against RFC-6455",
+  "typings": "type-definitions/ws.d.ts",
   "version": "1.1.0",
   "license": "MIT",
   "main": "index.js",
--- a/type-definitions/ws.d.ts
+++ b/type-definitions/ws.d.ts
@@ -1,0 +1,106 @@
+
+import { EventEmitter } from 'events'
+import * as http from 'http'
+import * as https from 'https'
+import * as net from 'net'
+
+export class Server extends EventEmitter {
+
+    constructor( options: ServerOptions, cb?: () => any );
+
+    close();
+    handleUpgrade( request: http.ServerRequest, socket: net.Socket, upgradeHead, callback: (webSocket: WebSocket) => any );
+    on( event: 'error', handler: (error: Error) => any ): this;
+    on( event: 'headers', handler: (headers: any) => any ): this;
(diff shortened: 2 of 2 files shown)


lpinca:
@RamIdeas thanks for this patch.

I think it is better to keep this on hold until the next major version is released. There might be some API changes that should be included in the typings file.

What do you think?

Aleksey-Danchin:
Merge this one! Needed right now =)

lpinca:
It seems that typings are available at https://github.com/DefinitelyTyped/DefinitelyTyped.
Is there any advantage in adding TypeScript type definitions in this repo?

3rd-Eden:
Agreed with @lpinca - Typings should just stay in that repo where there actually authors who want to maintain it. Neither me, @lpinca or anyone of the admin is using it our selfs so there wouldn't be really a need from our side to maintain it.

elithrar:
Just to re-assess this a few months later: the type definitions are out of date, and having them in the library by default can make that easier to identify, fix and review by those most across the codebase.

I'm going to fix the ones in DefinitelyTyped—many of the events don't have the `http.IncomingRequest` object added—but I can also create PR here and add the `types` property to package.json.

@lpinca Of course, it's your repo, so it's your decision!

lpinca:
@elithrar the problem is the same. We don't use TypeScript so the definitions will not be updated by us.

We had a similar discussion in Primus https://github.com/primus/primus/pull/537 but definitions there are still broken. Now ws is way more popular so it's totally possible that there will be more involvement. What I don't want is:

1. maintain and update the definitions as I'm not qualified to do that.
2. ship something that is broken.
3. creating patch releases only to fix issues with the definitions (when community find them) as it is a bit tedious. 

This is basically why I prefer to have the definitions in a separate package.

lpinca:
Forgot to say thank you for updating the definitions on DefinitelyTyped.

elithrar:
@lpinca - understood. I've fixed them here - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/16775 - you are welcome to give them a quick review (if you have the time).

julien-c:
I would like to second this suggestion as well. To your point on shipping broken code @lpinca, in my opinion the likelihood would be really low considering you already maintain the document at https://github.com/websockets/ws/blob/master/doc/ws.md, and the typings are basically the same document (you can actually probably go from one to the other automatically). Your decision in the end though!

julien-c:
PS: you know this already, but upgrading to ws `v4.0` in a large-ish codebase makes me sweat. I can read the breaking changes and skim through my code base, but having synchronized typings and checking if the "compiler" reports anything broken would make my life so much more joyful 😇

PR #872, opened 2016-10-21
Add threshold option to compression.

Nibbler999:
I think we need an option like this to limit compression to larger payloads. The default value of 1024 bytes is taken from socket.io.

### DIFF
--- a/lib/Sender.js
+++ b/lib/Sender.js
@@ -78,7 +78,7 @@ class Sender {
    */
   doPing (data, options) {
     var mask = options && options.mask;
-    this.frameAndSend(0x9, data || '', true, mask);
+    this.frameAndSend(0x9, data ? Buffer.from(data.toString()) : null, true, mask);
     if (this.extensions[PerMessageDeflate.extensionName]) {
       this.messageHandlerCallback();
     }
@@ -104,7 +104,7 @@ class Sender {
    */
   doPong (data, options) {
     var mask = options && options.mask;
-    this.frameAndSend(0xa, data || '', true, mask);
+    this.frameAndSend(0xa, data ? Buffer.from(data.toString()) : null, true, mask);
     if (this.extensions[PerMessageDeflate.extensionName]) {
       this.messageHandlerCallback();
     }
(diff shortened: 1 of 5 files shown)


lib/Sender.js
```diff
@@ -142,6 +142,11 @@ class Sender {
    * @api private
    */
   sendCompressed (opcode, data, finalFragment, mask, compress, cb) {
+    if (data && data.length < this.extensions[PerMessageDeflate.extensionName].threshold) {
```
Nibbler999: We have to enqueue to preserve message ordering. I'll think about `length`.

lib/Sender.js
```diff
@@ -78,7 +78,7 @@ class Sender {
    */
   doPing (data, options) {
     var mask = options && options.mask;
-    this.frameAndSend(0x9, data || '', true, mask);
+    this.frameAndSend(0x9, data ? Buffer.from(data.toString()) : null, true, mask);
```
lpinca: Is `toString()` to make sure that whatever is passed to `ping` is converted to a string?

test/WebSocket.test.js
```diff
@@ -2226,7 +2226,7 @@ describe('WebSocket', function() {
     describe('#terminate', function() {
       it('will raise error callback, if any, if called during send data', function(done) {
         var wss = new WebSocketServer({port: ++port, perMessageDeflate: true}, function() {
-          var ws = new WebSocket('ws://localhost:' + port, {perMessageDeflate: true});
+          var ws = new WebSocket('ws://localhost:' + port, {perMessageDeflate: { threshold: 0 }});
```
lpinca: Why this test fails without this change?

lib/Sender.js
Nibbler999: Yes, we need to be able to accept a number here.

test/WebSocket.test.js
Nibbler999: The test relies on compression but the payload is below the default threshold size.

lib/Sender.js
lpinca: Ok, It might be better to use `String(data)` which won't fail with `null` or `undefined`.

PR #874, opened 2016-10-28
Remove Makefile in favor of npm scripts

lpinca:
The Makefile targets are very simple so I think it makes sense to favor npm scripts instead of working with Makefiles syntax.


JacksonTian:
LGTM

PR #875, opened 2016-10-28
Remove `stream` method and ability to send a stream

lpinca:
I would like to discuss a little about this change to see if it makes sense or it is better to keep things as they are.

In my opinion the functionality offered by `WebSocket.prototype.stream()` does not belong to `ws` as it is something that should be done in the application and not in the library. The same applies for the ability to use a stream in `WebSocket.prototype.send()`.

For example instead of

``` js
ws.send(stream, cb);
```

the developer could use something like this

``` js
(function read(readable) {
  const chunk = readable.read();

  if (chunk === null) {
    ws.send(chunk, { fin: true });
    return
  }

  ws.send(chunk, { fin: false }, (err) => {
    if (err) return handleError(err);

    read(readable);
  });
})(readable);
```

and queue/handle additional sends that can happen while reading from the stream.


JoshuaWise:
I completely agree.

How do you feel about implementing a proper DuplexStream interface to WebSocket? I know there's been a lot of talk about that.

lpinca:
Yeah making `WebSocket` a duplex stream would be the ideal solution.

WRT this particular change I wonder how many users depend on the removed features. I think not many but I may be wrong.

lpinca:
@websockets/admin any more opinions on this?

lpinca:
Another point in favor of this is the existence of modules like https://github.com/maxogden/websocket-stream which has 100k monthly downloads.

3rd-Eden:
+1

PR #885, opened 2016-11-06
Remove path registry

lpinca:
The `_webSocketPaths` object is added to the underlying HTTP sever of a `WebSocketServer` when the `path` option is specified. It is used to store the path of each `WebSocketServer` using the same underlying HTTP server in order to avoid path conflicts.

```js
const wss1 = new WebSocket.Server({ server, path: '/foo' });
const wss2 = new WebSocket.Server({ server, path: '/bar' });
```

This allows to use multiple WebSocket severs (one per path) and only one shared HTTP server.

The problem here is that each `WebSocketServer` adds a new listener for the `upgrade` event on the HTTP server and when that event is emitted, [`handleUpgrade`](https://github.com/websockets/ws/blob/ebf86b58f1fb1e96ab96ebf95782bad7bc5ab8ca/lib/WebSocketServer.js#L164-L180) is called on all servers.

Now, using the above example, a connection to `ws://example.com/foo` would first be established on `wss1` and then closed by `wss2` as they both write on the same socket.

Originally this worked because path validation did not close the connection on [mismatch](https://github.com/websockets/ws/blob/0cfa5cc94d50bebd6f85b78ec1094dd9a7a17d62/lib/WebSocketServer.js#L171).

Passing the same HTTP server in the constructor when instantiating multiple WebSocket servers is not a good idea imo and should be avoided. I can't think of a use case where it makes sense.

In order to have multiple WebSocket servers and only one shared HTTP server, a developer could use something like this:

```js
const wss1 = new WebSocket.Server({ noServer: true });
const wss2 = new WebSocket.Server({ noServer: true });

server.on('upgrade', (request, socket, head) => {
  const pathname = url.parse(request.url).pathname;

  if (pathname === '/foo') {
    wss1.handleUpgrade(request, socket, head, (ws) => {
      wss1.emit('connection', ws);
    });
    wss2.handleUpgrade(request, socket, head, (ws) => {
      wss2.emit('connection', ws);
    });
  } else {
    socket.destroy();
  }
});
```

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -67,15 +67,6 @@ function WebSocketServer (options, callback) {
     this._closeServer = () => this._server && this._server.close();
   } else if (options.server) {
     this._server = options.server;
-    if (options.path) {
-      // take note of the path, to avoid collisions when multiple websocket servers are
-      // listening on the same http server
-      if (this._server._webSocketPaths && options.server._webSocketPaths[options.path]) {
-        throw new Error('two instances of WebSocketServer cannot listen on the same http server path');
-      }
-      if (!this._server._webSocketPaths) this._server._webSocketPaths = {};
-      this._server._webSocketPaths[options.path] = 1;
-    }
   }
 
   if (this._server) {
@@ -124,14 +115,6 @@ WebSocketServer.prototype.close = function (callback) {
       } catch (e) {
         error = e;
       }
-    }
-  }
-
-  // remove path descriptor, if any
-  if (this.path && this._server._webSocketPaths) {
-    delete this._server._webSocketPaths[this.path];
-    if (Object.keys(this._server._webSocketPaths).length === 0) {
-      delete this._server._webSocketPaths;
     }
   }
 
--- a/test/WebSocketServer.test.js
+++ b/test/WebSocketServer.test.js
@@ -109,48 +109,6 @@ describe('WebSocketServer', function () {
       });
     });
 
-    it('can have two different instances listening on the same http server with two different paths', function (done) {
-      const server = http.createServer();
-
-      server.listen(++port, () => {
-        const wss1 = new WebSocketServer({ server, path: '/wss1' });
-        const wss2 = new WebSocketServer({ server, path: '/wss2' });
-        let doneCount = 0;
-
-        wss1.on('connection', (client) => {
-          wss1.close();
-          if (++doneCount === 2) {
-            server.close(done);
-          }
-        });
-
-        wss2.on('connection', (client) => {
-          wss2.close();
-          if (++doneCount === 2) {
-            server.close(done);
-          }
-        });
-
-        /* eslint-disable no-unused-vars */
-        const ws1 = new WebSocket(`ws://localhost:${port}/wss1`);
-        const ws2 = new WebSocket(`ws://localhost:${port}/wss2?foo=1`);
-        /* eslint-enable no-unused-vars */
-      });
-    });
-
-    it('cannot have two different instances listening on the same http server with the same path', function (done) {
-      const server = http.createServer();
-      const wss1 = new WebSocketServer({ server: server, path: '/wss1' });
-
-      try {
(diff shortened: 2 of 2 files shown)


dosaygo-coder-0:
Thanks for the great workaround write up!

antoniobusrod:
Hi @lpinca , is it possible to keep using different WS servers for different paths with `noServer` property? It seems it doesn't (I've started to do changes from 2.x in my code).

lpinca:
@antoniobusrod see the above example. Can't you do that?

antoniobusrod:
hi @lpinca , I'm preparing a PR, so let's discuss there

sugir93:
Thanks for the solution (Y)

mayeaux:
Can this be written somewhere in the docs? Lost a good chunk of time trying to figure out why one server worked but adding another blew up with an `Invalid frame header` error

yiakwy:
I am using the pattern. Can anybody explain to me what's meaning of websocket.emit("connection")? can it work together with websocket.on("connection", socket)?

yiakwy:
Here is my code:

```js
/*
 * @desc This techniques allow multiple websockets used, @see 
 */ 
function onUpgrade(req, HTTPSocket, head) {
 	const loc = url.parse(req.url).pathname

	switch(loc) {
		case "/sim_world":
			// WebSocketServer handle it 
			// console.log(`Receive websocket connection request ${stringify(req, null, 2)} from ${loc}`)
			console.log(`[server:onUpgrade] receive websocket connection request ${req} from ${loc}`)
			ws.handleUpgrade(req, HTTPSocket, head, (socket) => {
				// debug("ws.handleUpgrade %s, %s", socket, req) 
				console.log("[ws.handleUpgrade] emit connection event.")
				ws.emit('connection', socket)
			})

			ws.on("connection", connection)
			break
		default:
			console.log(`Not recognized websocket request ${loc}`)
			HTTPSocket.destroy()
	}

}

```

Here is my output
```bash
[server:onUpgrade] receive websocket connection request [object Object] from /sim_world
[ws.handleUpgrade] emit connection event.
events.js:167
      throw er; // Unhandled 'error' event
      ^

Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    at writeAfterEnd (_stream_writable.js:243:12)
    at Receiver.Writable.write (_stream_writable.js:292:5)
    at Socket.socketOnData (/Users/lei.wang1/Github/RandomMapGenerator/backend/node_modules/ws/lib/websocket.js:788:35)
    at Socket.emit (events.js:187:15)
    at Socket.EventEmitter.emit (domain.js:442:20)
    at addChunk (_stream_readable.js:280:12)
    at readableAddChunk (_stream_readable.js:265:11)
    at Socket.Readable.push (_stream_readable.js:220:10)
    at TCP.onread (net.js:638:20)
Emitted 'error' event at:
    at Receiver.receiverOnError (/Users/lei.wang1/Github/RandomMapGenerator/backend/node_modules/ws/lib/websocket.js:698:13)
    at Receiver.emit (events.js:182:13)
    at Receiver.EventEmitter.emit (domain.js:442:20)
    at writeAfterEnd (_stream_writable.js:245:10)
    at Receiver.Writable.write (_stream_writable.js:292:5)
    [... lines matching original stack trace ...]
    at TCP.onread (net.js:638:20)

```

sugir93:
Check my working code.

var server = http.createServer().listen(300);
var wss_event = new WebSocket.Server({
    noServer: true
});
    var pathname = url.parse(request.url).pathname;

    if (pathname === '/sim_world') {
        wss_event.handleUpgrade(request, socket, head, function (ws) {
            wss_event.emit('connection', ws);
        });
    } else {
        socket.destroy();
    }
});

ifree92:
Yeah, thanks! That's working for me likewise.
In addition, if someone uses websockets and socket.io servers in one instance (yes, I'm also surprised, but I haven't other way), **do not close the socket in upgrade function**:
```js
else {
    socket.destroy();
  }
```

PR #889, opened 2016-11-12
Use ultron to handle the events listeners on the server

lpinca:
We are already using `ultron` on the client so I think it makes sense to also use it for the server.

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -6,23 +6,35 @@
 
 'use strict';
 
-const util = require('util');
 const EventEmitter = require('events');
-const http = require('http');
 const crypto = require('crypto');
-const WebSocket = require('./WebSocket');
-const Extensions = require('./Extensions');
-const PerMessageDeflate = require('./PerMessageDeflate');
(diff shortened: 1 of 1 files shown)


lib/WebSocketServer.js
```diff
@@ -44,67 +56,52 @@ function WebSocketServer (options, callback) {
...
 
   if (this._server) {
-    this._onceServerListening = () => this.emit('listening');
-    this._server.once('listening', this._onceServerListening);
-    this._onServerError = (error) => this.emit('error', error);
-    this._server.on('error', this._onServerError);
-    this._onServerUpgrade = (req, socket, upgradeHead) => {
-      // copy upgradeHead to avoid retention of large slab buffers used in node core
-      var head = new Buffer(upgradeHead.length);
-      upgradeHead.copy(head);
```
lpinca: This is no longer needed. See https://github.com/jmatthewsr-ms/node-slab-memory-issues/issues/1.

3rd-Eden: <3

PR #890, opened 2016-11-13
Remove callback from `handleProtocols`

lpinca:
Right now `handleProtocols` takes a callback which must be called **synchronously** by the developer with two arguments: the outcome of the operation and the selected protocol.

I think it makes sense to remove the callback argument and use the returned value instead.


lpinca:
@websockets/admin WDYT?

3rd-Eden:
@lpinca Sorry, I thought I already pressed the merge button.

lpinca:
No problem.

PR #892, opened 2016-11-15
Invoke the close callback when the close event is emitted

lpinca:
Right now the close callback is invoked immediately but it should be called when the HTTP server emits the 'close' event.

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -103,27 +103,18 @@ util.inherits(WebSocketServer, EventEmitter);
  */
 WebSocketServer.prototype.close = function (cb) {
   // terminate all associated clients
-  var error = null;
-
   if (this.clients) {
-    for (const client of this.clients) {
-      try {
-        client.terminate();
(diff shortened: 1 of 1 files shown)


lib/WebSocketServer.js
```diff
@@ -103,27 +103,18 @@ util.inherits(WebSocketServer, EventEmitter);
...
-
   if (this.clients) {
-    for (const client of this.clients) {
-      try {
-        client.terminate();
-      } catch (e) {
-        error = e;
-      }
-    }
+    for (const client of this.clients) client.terminate();
```
lpinca: I've removed the `try...catch` statement because `WebSocket.prototype.terminate()` does not throw errors.

3rd-Eden: But doesn't the underlaying node code could be throwing errors.

3rd-Eden:
nvm, all those other functions are already wrapping all socket calls in `try { } catch () {}`

PR #908, opened 2016-11-21
Call `Sender#close()` callback when data is written out

lpinca:
Currently the [callback](https://github.com/websockets/ws/blob/199c248a4a65d175fdcb351ff1d29c87debaf855/lib/Sender.js#L72) of `Sender.prototype.close()` is immediately invoked regardless of data being sent or not.

This patch forwards the callback of `Sender.prototype.close()` to `Sender.prototype.frameAndSend()` to actually call the callback when data is written out.

### DIFF
--- a/lib/Sender.js
+++ b/lib/Sender.js
@@ -65,11 +65,10 @@ class Sender {
    * @private
    */
   doClose (data, mask, cb) {
-    this.frameAndSend(0x08, data, false, true, mask, false);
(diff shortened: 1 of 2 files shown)


3rd-Eden:
LGTM

lpinca:
My only concern here is that if there cases where the callback is not invoked. This could lead to a memory leak.

lpinca:
Anyway the callback should always be called, even when writing to the socket fails, merging.

PR #910, opened 2016-11-22
Add TLS option - servername

Anatoliy4041:
Since some certeficates have feature of CN(Common Name) or SNI(Server Name Indication) and they are had to match it, I propose to add that option to the existing code.


lpinca:
Why you closed it?

PR #914, opened 2016-11-26
Simplify closing handshake

lpinca:
Currently the close handshake works like this when initiated from the server:

```
        SERVER                                 CLIENT


+--------------------+
|ws.close()          |
|readyState = CLOSING|
+--------------------+
           |            +-----------+
           +----------->|close frame|-------------+
                        +-----------+             v
                                       +--------------------+
                                       |closeReceived = true|
                                       |ws.close()          |
                                       |readyState = CLOSING|
                                       +--------------------+
                        +-----------+             |
           +------------|close frame|<------------+
           v            +-----------+
+--------------------+
|closeReceived = true|
|ws.close()          |
|ws.terminate()      |
|socket.end()        |
+--------------------+
           |              +-------+
           +------------->|TCP FIN|---------------+
                          +-------+               v
                                        +-------------------+
                                        |readyState = CLOSED|
                                        |ws.emit('closed')  |
                                        |socket.end()       |
                                        +-------------------+
                          +-------+               |
           +--------------|TCP FIN|<--------------+
           |              +-------+
           v
 +-------------------+
 |readyState = CLOSED|
 |ws.emit('closed')  |
 +-------------------+
```

and like this when initiated from the client:

```
        CLIENT                                 SERVER


+--------------------+
|ws.close()          |
|readyState = CLOSING|
+--------------------+
           |            +-----------+
           +----------->|close frame|-------------+
                        +-----------+             v
                                       +--------------------+
                                       |closeReceived = true|
                                       |ws.close()          |
                                       |readyState = CLOSING|
                                       |ws.terminate()      |
                                       |socket.end()        |
                                       +--------------------+
                   +---------------------+        |
           +-------|close frame + TCP FIN|<-------+
           v       +---------------------+
+--------------------+
|closeReceived = true|
|ws.close()          |
|readyState = CLOSED |
|ws.emit('closed')   |
|socket.end()        |
+--------------------+
           |              +-------+
           +------------->|TCP FIN|---------------+
                          +-------+               v
                                        +-------------------+
                                        |readyState = CLOSED|
                                        |ws.emit('closed')  |
                                        +-------------------+
```

This patch uniforms the behavior making a close initiated by the server work in the same way of a close initiated by the client.

After reading https://tools.ietf.org/html/rfc6455#section-1.4 and https://tools.ietf.org/html/rfc6455#section-7, I think this is correct.

I have some doubts on this [section](https://tools.ietf.org/html/rfc6455#section-7.1.1) which is why the close handshake is currently implemented in this way for the server but I also checked other implementations for Node.js and they use the same procedure for both client and server as proposed in this patch.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -255,9 +255,7 @@ class WebSocket extends EventEmitter {
     }
 
     if (this.readyState === WebSocket.CLOSING) {
-      if (this._closeCode && this._isServer) {
-        this.terminate();
-      }
+      if (this._closeCode) this.terminate();
       return;
     }
 
@@ -265,7 +263,7 @@ class WebSocket extends EventEmitter {
     this._sender.close(code, data, !this._isServer, (err) => {
       if (err) this.emit('error', err);
 
-      if (this._closeCode && this._isServer) {
+      if (this._closeCode) {
         this.terminate();
       } else {
         //
--- a/test/WebSocket.test.js
+++ b/test/WebSocket.test.js
@@ -1010,6 +1010,39 @@ describe('WebSocket', function () {
 
       wss.on('connection', (ws) => ws.close(1013));
     });
+
+    it('closes the connection when an error occurs', function (done) {
+      const server = http.createServer();
+      const wss = new WebSocketServer({ server });
+      let closed = false;
+
+      wss.on('connection', (ws) => {
+        ws.on('error', (err) => {
+          assert.ok(err instanceof Error);
+          assert.strictEqual(err.message, 'RSV2 and RSV3 must be clear');
+
+          ws.on('close', (code, reason) => {
+            assert.strictEqual(code, 1006);
+            assert.strictEqual(reason, '');
+
+            closed = true;
+          });
+        });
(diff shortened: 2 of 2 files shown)


lpinca:
cc: @nkzawa

lpinca:
I guess the question to answer is:

Is there any advantage on calling `socket.end()` first on the client and then on the server when a close procedure is initiated by the server?

I don't know much about low level networking so if someone could shed some light on this, it would be great.

lpinca:
@websockets/admin any thoughts on this?

citrusjunoss:
When my client is Node and the server is Java(Netty), can it work?

lpinca:
@citrusjunoss yes.

citrusjunoss:
> @citrusjunoss yes.


But  the server does not receive any data when I call the close method

lpinca:
It should receive a close frame, if you don't specify a close code, it will be something like this:

```
88 80 45 68 03 77
```

The last 4 bytes are random (the mask). You can check this with wireshark if you have no access on the TCP socket.

PR #918, opened 2016-11-30
Remove `supports` attribute

lpinca:
Hixie-76 is no longer supported so this non-standard attribute can be removed as binary messages are always supported.


lpinca:
It doesn't need rebasing, merging.

PR #925, opened 2016-12-06
Clean up `WebSocket.createServer()` and `WebSocket.connect()`

lpinca:
This cleans up the factory functions `WebSocket.createServer()` and `WebSocket.connect()`.

Put this in a PR because to be honest I would like to remove them but I would like to hear what other collaborators think before doing that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,49 +1,60 @@
-'use strict';
-
 /*!
  * ws: a node.js websocket client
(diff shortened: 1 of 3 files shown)


3rd-Eden:
People should just use the default API's. Our current master is already breaking a shit ton of things so we might as well nuke this to pieces as well

PR #930, opened 2016-12-09
Refactor `WebSocket` to use class syntax

lpinca:
- Refactor `WebSocket` to use the ES6 class syntax.
- Move `cleanupWebsocketResources` and `establishConnection` to `WebSocket.prototype`.
- Rename `establishConnection` to `setSocket` and `cleanupWebsocketResources` to `finalize`.
- Keep `initAsServerClient` and `initAsClient` as completely private API.

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -11,7 +11,6 @@ const crypto = require('crypto');
 const Ultron = require('ultron');
 const https = require('https');
 const http = require('http');
-const util = require('util');
 const url = require('url');
 
 const PerMessageDeflate = require('./PerMessageDeflate');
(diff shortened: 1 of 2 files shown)


lpinca:
I think the new methods names (`setSocket` and `finalize`) are better than the original ones but I'm not very happy with them. If you have better ideas please share!

lib/WebSocket.js
```diff
@@ -25,277 +24,391 @@ const closeTimeout = 30 * 1000; // Allow 30 seconds to terminate the connection
...
- * Ready States
- */
-['CONNECTING', 'OPEN', 'CLOSING', 'CLOSED'].forEach(function each (state, index) {
-  WebSocket.prototype[state] = WebSocket[state] = index;
-});
+  /**
+   * @type {Number}
+   */
+  get bufferedAmount () {
+    var amount = 0;
```
ericmdantas: `let`?

lpinca: Not yet as it causes deoptimizations.

lpinca:
@ericmdantas thank you for reviewing, this is greatly appreciated! Let me know if you have more suggestions.

lib/WebSocket.js
```diff
@@ -25,277 +24,391 @@ const closeTimeout = 30 * 1000; // Allow 30 seconds to terminate the connection
...
-    // First cleanup any pre-existing timeout from an earlier "terminate" call,
-    // if one exists.  Otherwise terminate calls in quick succession will leak timeouts
-    // and hold the program open for `closeTimout` time.
-    if (this._closeTimer) { clearTimeout(this._closeTimer); }
-    this._closeTimer = setTimeout(cleanupWebsocketResources.bind(this, true), closeTimeout);
-  } else if (this.readyState === WebSocket.CONNECTING) {
-    cleanupWebsocketResources.call(this, true);
-  }
-};
+    if (!data) data = '';
```
3rd-Eden: @lpinca Now that node 6 is the new LTS, maybe we should use that our minimum which also allows us to use default params.

lpinca: @3rd-Eden I think that dropping Node 4 support is a bit too much. For example AWS Lambda does not support Node 6 yet.

Now using ws with Lambda does not make much sense 😄  but I think you get my point.

3rd-Eden: @lpinca ok, we'll wait another week. Or until node 4 LTS runs out completely.

ericmdantas:
@lpinca, no problem. I'm a bit busy at the moment, but I'll take a closer look when possible. But, from what I've seen, it's good to go.

lpinca:
I'll merge this later today if there are no objections.

lib/WebSocket.js
JoshuaWise: I thought only *compound* `let` statements caused deoptimization?
source: http://stackoverflow.com/questions/34595356/what-does-compound-let-const-assignment-mean

lib/WebSocket.js
JoshuaWise: Maybe use `=== undefined` to mirror the behavior of default parameters, so that making the eventual switch doesn't cause any unintended side effects?

lib/WebSocket.js
lpinca: I think it's still better to keep `var` as it is consistent with the rest of the code base and we are 100% safe.
Switching to `let` should be very easy with an eslint rule when we are sure it doesn't cause any harm.

lib/WebSocket.js
lpinca: @JoshuaWise no strong opinion on this but for example https://github.com/websockets/ws/blob/5a8ead573f1fb2b7c061431a85b3b79c91013f6c/lib/Sender.js#L154-L163 just check that `data` is truthy.

lpinca: We should probably also allow `0` which doesn't work atm, so this makes sense but maybe it's better to fix the issue in another PR/commit as we should fix the edge case also in `Sender`.

lib/WebSocket.js
JoshuaWise: Makes sense. I wish there was documentation somewhere for v8's optimizations. It would make things like this so much easier.

PR #933, opened 2016-12-12
Change `Sender.prototype.frameAndSend()` signature

lpinca:
This updates the signature of the `frameAndSend` method to make it accept an options object.
This should make it easier to add more options in future without having to add more arguments and remember their position.


lpinca:
@websockets/admin LGTY?

Nibbler999:
You should also update the sender benchmark with the new signature.

lpinca:
Oh right, thanks for pointing that out, on my way.

lpinca:
I'll merge this later today if there are no objections.

PR #937, opened 2016-12-16
Remove unnecessary `try...catch` statements

lpinca:
This patch removes some `try...catch` statements which seem to be redundant, see inline comments.

### DIFF
--- a/lib/Sender.js
+++ b/lib/Sender.js
@@ -376,15 +376,10 @@ function getRandomMask () {
  * @private
  */
 function sendFramedData (sender, outputBuffer, data, cb) {
-  try {
-    if (data) {
-      sender._socket.write(outputBuffer);
-      sender._socket.write(data, cb);
-    } else {
-      sender._socket.write(outputBuffer, cb);
-    }
-  } catch (e) {
-    if (cb) cb(e);
-    else sender.onerror(e);
+  if (data) {
+    sender._socket.write(outputBuffer);
+    sender._socket.write(data, cb);
+  } else {
+    sender._socket.write(outputBuffer, cb);
   }
 }
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -188,17 +188,13 @@ class WebSocket extends EventEmitter {
     this.emit('close', this._closeCode || 1000, this._closeMessage || '');
 
     if (this._socket) {
-      if (this._ultron) this._ultron.destroy();
+      this._ultron.destroy();
       this._socket.on('error', function onerror () {
-        try {
-          this.destroy();
-        } catch (e) {}
+        this.destroy();
       });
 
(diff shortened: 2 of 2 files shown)


lib/Sender.js
```diff
@@ -376,15 +376,10 @@ function getRandomMask () {
  * @private
  */
 function sendFramedData (sender, outputBuffer, data, cb) {
-  try {
```
lpinca: An error is not thrown on the same tick when writing to a closed socket, so I _think_ this only handle the case where `sender._socket` is not a `net.Socket` instance but that never happens as `sender._socket` is never cleared.

lib/WebSocket.js
```diff
@@ -188,17 +188,13 @@ class WebSocket extends EventEmitter {
     this.emit('close', this._closeCode || 1000, this._closeMessage || '');
 
     if (this._socket) {
-      if (this._ultron) this._ultron.destroy();
```
lpinca: If `this._socket` is truthy then `this._ultron` is also truthy as they are cleared together.

lib/WebSocket.js
```diff
@@ -188,17 +188,13 @@ class WebSocket extends EventEmitter {
...
+      this._ultron.destroy();
       this._socket.on('error', function onerror () {
-        try {
-          this.destroy();
-        } catch (e) {}
+        this.destroy();
```
lpinca: Calling `socket.destroy` does not seem to throw an error on the same tick in any circumstance.

lib/WebSocket.js
```diff
@@ -188,17 +188,13 @@ class WebSocket extends EventEmitter {
     this.emit('close', this._closeCode || 1000, this._closeMessage || '');
 
     if (this._socket) {
-      if (this._ultron) this._ultron.destroy();
+      this._ultron.destroy();
       this._socket.on('error', function onerror () {
-        try {
```
lpinca: Calling `socket.destroy` does not seem to throw an error on the same tick in any circumstance.

lib/WebSocket.js
```diff
@@ -267,28 +263,23 @@ class WebSocket extends EventEmitter {
       return;
     }
 
-    try {
-      this.readyState = WebSocket.CLOSING;
-      this._closeCode = code;
-      this._closeMessage = data;
-      var mask = !this._isServer;
-      this._sender.close(code, data, mask, (err) => {
```
lpinca: This can throw when `WebSocket#close()` is called with an invalid status code, but I think it's better to not wrap this in a `try...catch` as it is a programmer error.

`WebSocket#close()` is never called internally with an invalid status code.

lpinca:
@websockets/admin please take a look if you can. I'll merge this tomorrow if there are no objections.

PR #939, opened 2016-12-20
Use the same handler for binary and text messages

lpinca:
This merges `receiver.onbinary` and `receiver.ontext` into `receiver.onmessage`.
I think thank using 2 different handlers is pointless.

This is tagged as "major" because it changes the `Receiver` interface but I don't think many people are  using the `Receiver` and `Sender` classes directly. Their interface is not documented and imo they should be private and not used directly. Unfortunately they are exposed as static properties of the `WebSocket` class.

I was also thinking that maybe it makes sense to emit to the `WebSocket` instance directly instead of using these wrapper [functions](https://github.com/websockets/ws/blob/fc956f8071e3930de432676ea2ea9cc075ee9fde/lib/WebSocket.js#L139).

For example we could make the `Receiver` constructor accept a `WebSocket` instance, the one the receiver is bound to.

```js
class Receiver {
  constructor (websocket, extensions, maxPayload) {
    this.websocket = websocket;
    ...
  }
}
```

and then instead of calling the wrapper functions like [this](https://github.com/websockets/ws/blob/fc956f8071e3930de432676ea2ea9cc075ee9fde/lib/Receiver.js#L353), emit the event directly on the `websocket` object.

```js
this.websocket.emit('message', buf, { binary: true, masked: this.masked });
```

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -53,10 +53,9 @@ class Receiver {
 
     this.dead = false;
 
-    this.onbinary = noop;
+    this.onmessage = noop;
     this.onclose = noop;
     this.onerror = noop;
-    this.ontext = noop;
     this.onping = noop;
     this.onpong = noop;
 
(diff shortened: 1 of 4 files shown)


lpinca:
I hate that coveralls reports that coverage decreased only because there are fewer lines of codes  😑.

3rd-Eden:
tsk tsk, shame on you @lpinca for writing better code ;-)

Issue #923, opened 2016-12-06
Handle an error when message could not been sent


Josema:
This code is running in the same node.js script
```js
server.close(); // This is a socket created with new WebSocket('ws://localhost:1234/test');
clientX.send("something"); // This is a socket from wss.on('connection', function connection(clientX ...
```

Here I am trying to close the connection to the server from the client. The message is never caught by the client.

I tried `clientX.send("something", function(err){ ... });` but err is always undefined. And the readyState for clientX is OPEN.

Is this message being missed or is just because `server.close()` flag the client as close and do not emit more message events?

lpinca:
The message is probably sent but I'm not sure why it doesn't reach the client.
Can you post a reproducible test case?

Josema:
```js
var WebSocket = require('ws'),
    wss = new WebSocket.Server({ port:1234 });

//server
wss.on('connection', function connection(client) {
    client.send('something1');
    setTimeout(function(){
        serverConnection.close(); // Comenting this something2 will be handle
        client.send('something2');
    }, 2000)
});



//client
var serverConnection = new WebSocket('ws://localhost:1234');
serverConnection.on('message', function incoming(message) {
    console.log('received: %s', message);
});
```

lpinca:
From a quick look it seems a duplicate of #799.
Try to disable permessage-deflate and see if it works.

The problem is that the client receives a TCP FIN packet when data is being decompressed. It's a bug that should be fixed.

Josema:
Any help to disable `permessage-deflate`? Don't know how to do it.

lpinca:
Use `perMessageDeflate: false` in the server options:

```
const wss = new WebSocket.Server({ perMessageDeflate: false, port:1234 });
```

Josema:
It works! Thank you!

lpinca:
No problem.

Josema:
`perMessageDeflate` has a big side effect.  
If `serverConnection` close the connection just after the connection is established then the event `close` is not emitted either in client and server.

I wouldn't recommend use this.

PR #840, opened 2016-09-25
Wait for receiver to finish processing frames before closing. 

Nibbler999:
The socket close handler is now aware that frames can be processed asynchronously and will wait for the receiver to finish before emitting the close event.

Fixes #799.

PR #945, opened 2016-12-24
Consume all data before emitting the `close` event

### LINKED ISSUE #799: Last frame's payload not delivered and code 1006 returned
The title may not be 100% accurate, I apologize for that but it's not too clear to me what's happening...
### Versions

I can say for sure that this happens with version 1.1.1 but not with 0.4.32.

**Edit:** It doesn't work with version 0.8.1 either. The only difference with respect to 1.1.1 is that the close event code is still 1000.
### Client code

The client code is not really important, you can use the basic:

``` js
var ws = new WebSocket(process.argv[2]);

ws.on('message', function (data, flags) {
    console.log('>', data);
});

ws.on('close', function (code, message) {
    console.error('#', code, message);
});
```

or even [`wscat`](https://github.com/websockets/wscat), what really matter is the `ws` version.
### How to reproduce

This behavior has been observed with the Google Chrome's Remote Debugging Protocol WebSocket; (cyrus-and/chrome-remote-interface#39).

Here are the step to reproduce this issue:
1. start a new Chrome session with `--remote-debugging-port=9222`;
2. fetch a list of the currently open tabs:
   
   ```
   $ curl http://localhost:9222/json
   [ {
      "description": "",
      "devtoolsFrontendUrl": "/devtools/inspector.html?ws=localhost:9222/devtools/page/952808E7-26BC-470B-B1DF-B61519786481",
      "faviconUrl": "https://www.google.it/favicon.ico",
      "id": "952808E7-26BC-470B-B1DF-B61519786481",
      "title": "New Tab",
      "type": "page",
      "url": "chrome://newtab/",
      "webSocketDebuggerUrl": "ws://localhost:9222/devtools/page/952808E7-26BC-470B-B1DF-B61519786481"
   },
   [...]
   ```
   
   Locate the new empty tab (`chrome://newtab/`) and not the WebSocket URL (in this case `ws://localhost:9222/devtools/page/952808E7-26BC-470B-B1DF-B61519786481`).

3. connect to that URL:
   
   ```
   $ wscat -c ws://localhost:9222/devtools/page/952808E7-26BC-470B-B1DF-B61519786481
   connected (press CTRL+C to quit)
   ```
4. from Chrome open the DevTools for that tab (this will close the WebSocket).
#### What happens with 1.1.1

```
connected (press CTRL+C to quit)
disconnected
```

Using a custom client we obtain the close event code 1006.

Using Wireshark we notice:
- `Sec-WebSocket-Extensions` is set to `permessage-deflate; client_max_window_bits=15` in the handshake reply;
- `Fin: True` and `Reserved: 0x04` and the payload is not human-readable in the last (and only) WebSocket frame.
#### What happens with 0.4.32

```
connected (press CTRL+C to quit)
< { "method": "Inspector.detached", "params": { "reason": "replaced_with_devtools"} }
disconnected
```

Using a custom client we obtain the close event code 1000.

Using Wireshark we notice:
- `Sec-WebSocket-Extensions` is not set in the handshake reply;
- `Fin: True` and `Reserved: 0x00` and the payload is the above JSON text in the last (and only) WebSocket frame.

lpinca:
This makes `Receiver.prototype.cleanup()` accept an optional callback which is called when all buffered data is consumed or an error occurs.

Using this callback we can handle cases where the parser is decompressing data and the other peer closes the underlying TCP connection. Previously when this happened the buffered data was simply discarded. Now all data should be handled correctly.

Fixes #799 and #923.

This is based and supersedes #840.

@websockets/admin this really needs to be reviewed carefully so when you have time please take a look.

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -51,6 +51,8 @@ class Receiver {
     this.messageLength = 0;
     this.fragments = [];
 
+    this.cleanupCallback = null;
+    this.hadError = false;
     this.dead = false;
 
     this.onmessage = noop;
@@ -108,6 +110,21 @@ class Receiver {
   }
 
   /**
+   * Checks if the number of buffered bytes is bigger or equal than `n` and
+   * calls `cleanup` if necessary.
+   *
+   * @param {Number} n The number of bytes to check against
+   * @return {Boolean} `true` if `bufferedBytes >= n`, else `false`
+   * @private
+   */
+  hasBufferedBytes (n) {
+    if (this.bufferedBytes >= n) return true;
+
+    if (this.dead) this.cleanup(this.cleanupCallback);
+    return false;
+  }
+
+  /**
    * Adds new data to the parser.
    *
    * @public
@@ -142,7 +159,7 @@ class Receiver {
    * @private
    */
   start () {
-    if (this.bufferedBytes < 2) return;
+    if (!this.hasBufferedBytes(2)) return;
 
     const buf = this.readBuffer(2);
 
@@ -222,7 +239,7 @@ class Receiver {
    * @private
    */
   getPayloadLength16 () {
-    if (this.bufferedBytes < 2) return;
+    if (!this.hasBufferedBytes(2)) return;
 
     this.payloadLength = this.readBuffer(2).readUInt16BE(0, true);
     this.haveLength();
@@ -234,7 +251,7 @@ class Receiver {
    * @private
    */
   getPayloadLength64 () {
-    if (this.bufferedBytes < 8) return;
+    if (!this.hasBufferedBytes(8)) return;
 
     const buf = this.readBuffer(8);
     const num = buf.readUInt32BE(0, true);
@@ -277,7 +294,7 @@ class Receiver {
    * @private
    */
   getMask () {
-    if (this.bufferedBytes < 4) return;
+    if (!this.hasBufferedBytes(4)) return;
 
     this.mask = this.readBuffer(4);
     this.state = GET_DATA;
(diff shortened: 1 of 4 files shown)


lib/Receiver.js
```diff
@@ -466,21 +482,28 @@ class Receiver {
...
+      this.buffers = null;
+      this.mask = null;
+
+      this.onmessage = null;
+      this.onclose = null;
+      this.onerror = null;
+      this.onping = null;
+      this.onpong = null;
+
+      if (cb) cb();
```
3rd-Eden: Do we want to set the `cleanupCallback` to null here as well?

lib/WebSocket.js
```diff
@@ -214,7 +228,7 @@ class WebSocket extends EventEmitter {
     this.extensions = null;
 
     this.removeAllListeners();
-    this.on('error', function onerror () {}); // catch all errors after this
+    this.on('error', noop); // Catch all errors after this.
```
3rd-Eden: Not related to this PR, but we need start doing a cleaner way of handling this. Right now valid errors might get swallowed without us knowning. Maybe something like `diagnostics`/`debug` would be more useful here so we can at least still log when these errors get swallowed.

lib/Receiver.js
lpinca: 🔧 in 1c547cb.

lib/WebSocket.js
lpinca: I _think_ the only cases where an error event may be emitted here are when trying to write to a socket that has already been closed due to write operation being queued in the `Sender`. In particular.

1. https://github.com/websockets/ws/blob/549cbe11527320f123e423bb13a1b20bc1807874/lib/WebSocket.js#L266.

2. https://github.com/websockets/ws/blob/549cbe11527320f123e423bb13a1b20bc1807874/lib/WebSocket.js#L159.

Maybe we should remove all queued operations in `Sender` when `finalize` is called.

PR #950, opened 2017-01-02
Fix ember (browserify) usage correcting dependencies

Polve:
I'm trying to use this module for a library that uses websockets and that must work both from node (command line) and the browser (using ember).

While in node everything works fine, ember is unable to build the dependencies because it lacks those two packages:
* bufferutil
* utf-8-validate

I hence modified the package.json to define those modules as dependencies instead of dev-dependencies.

Since without this patch the package is unusable (at least with ember), I ask for this patch to be accepted upstream :-)

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,18 +25,18 @@
     "lint": "eslint ."
   },
   "dependencies": {
-    "ultron": "~1.0.2"
+    "bufferutil": "^1.3.0",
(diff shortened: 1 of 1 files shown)


3rd-Eden:
These are optional binary dependencies and should be installed by users them selfs if they want to them or not as they can cause the installation to fail if you do not have the correct tool chain setup.

PR #951, opened 2017-01-02
Use a boolean instead of an options object

lpinca:
With this change `Websocket.prototype.ping()` and `Websocket.prototype.pong()` take a boolean as second argument instead of an options object.

Ping/Pong messages can't be compressed or fragmented and are always binary, so I think it doesn't make sense to use an options object as the only thing which can be (arguably) configured is whether or not to mask data.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -320,25 +320,23 @@ receives an `OpenEvent` named "open".
 
 Pause the socket.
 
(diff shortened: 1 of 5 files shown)


3rd-Eden:
+1, 3 is about the max amount of arguments we should allow though.

PR #953, opened 2017-01-04
Update fileapi example to work with Express 4

hodannyy:
Express does not inherit from http in newer versions so express.createServer() is deprecated.


lpinca:
Thanks. How about tweaking the example to make it work with express 4 instead?
It shouldn't be much work.

hodannyy:
Sounds good. I've updated the fileapi example to work with Express 4.

examples/fileapi-express_4/server.js
```diff
@@ -101,6 +102,6 @@ fs.mkdir(path.join(__dirname, '/uploaded'), function () {
   // ignore errors, most likely means directory exists
   console.log('Uploaded files will be saved to %s/uploaded.', __dirname);
   console.log('Remember to wipe this directory if you upload lots and lots.');
-  app.listen(8080);
+  server.listen(8080);
```
lpinca: Can you add a callback for the `listening` event and move the next line (`console.log()`) there?

lpinca:
@hodannyy I think there is no need to rename the folder, please keep the original name.

hodannyy:
@lpinca I've renamed it back to the original name. Thought it was a naming convention since the Express 3 serverstats example has an express_3 prefix.

lpinca:
I see, I think it's better to merge those two and keep only one `serverstats` example which uses the latest express.

lpinca:
Thanks!

PR #954, opened 2017-01-05
Update serverstats to use Express 4.14 and remove older examples

hodannyy:
Addressing a suggestion from @lpinca in #953 to merge the two `serverstats` examples and update it to use latest Express.


lpinca:
👍 thanks!

PR #955, opened 2017-01-05
WebSocket object leakage in WebSocketServer.clients when using WebSocket.onclose = function(..){..}

jacobbogers:
Hi, 
I discovered a leakage in WebSocket.clients when using callbacks like so for WebSocket
```
 ws.onclose = function close(evt) {
           ..
}
```
In the code I discovered that on connection on the server side a callback is registered to remove the peer from the ```WebSocket.clients``` Set

 Unfortunately this hidden callback is removed when using  `ws.onclose = function `  style.

 Since this is a cleanup operation, (after closure of socket) I changed the 'hidden' emit event name to "cleanup".

this will prevent the hidden callback removal when using  `ws.onclose == function` style callback
(Look at how onclose property is defined, in removes all event handlers , so also any hidden "close" event handlers)

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -148,6 +148,7 @@ class WebSocket extends EventEmitter {
       this._closeMessage = reason;
       this._closeCode = code;
       this.close(code, reason);
+      this.emit('close');
     };
     this._receiver.onerror = (error, code) => {
       // close the connection when the receiver reports a HyBi error code
@@ -219,6 +220,9 @@ class WebSocket extends EventEmitter {
    */
   emitClose () {
     this.readyState = WebSocket.CLOSED;
+    // emit 'cleanup' before 'close' because close will change attributes 
+    // and closed ws-sockets will not be found 'WebSocketServer.clients'  js-Set collection
+    this.emit('cleanup');
     this.emit('close', this._closeCode || 1006, this._closeMessage || '');
 
     if (this.extensions[PerMessageDeflate.extensionName]) {
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -148,7 +148,6 @@ class WebSocket extends EventEmitter {
       this._closeMessage = reason;
       this._closeCode = code;
       this.close(code, reason);
-      this.emit('close');
     };
     this._receiver.onerror = (error, code) => {
       // close the connection when the receiver reports a HyBi error code
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -219,7 +219,7 @@ class WebSocket extends EventEmitter {
    */
   emitClose () {
     this.readyState = WebSocket.CLOSED;
-    // emit 'cleanup' before 'close' because close will change attributes 
+    // emit 'cleanup' before 'close' because close will change attributes
     // and closed ws-sockets will not be found 'WebSocketServer.clients'  js-Set collection
     this.emit('cleanup');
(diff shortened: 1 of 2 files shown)


jacobbogers:
excuse me it seems to fail checks on node 4 but node 6 and 7 are ok, please advise

lpinca:
@jacobbogers good catch, don't worry about the failure on node 4, it is not related to this change.
I wonder if we can fix the issue by changing how browser API emulation work. The problem is that `ws.onclose() = () => {}` calls `ws.removeAllListeners('close')` so maybe it's better to fix that.

jacobbogers:
yes I know the property removes all listeners on "onclose" set property, 
that is a good thing, and is compliant with web frontend behavior (like onload , onerror, etc)

but hidden "close" to cleanup is just piggyback

better to make separate event for cleanup also it intent is made clear

jacobbogers:
maybe call the event "remove-from-queue"?

lpinca:
@jacobbogers I was thinking about something like this:
```diff
diff --git a/lib/WebSocket.js b/lib/WebSocket.js
index 4baf5a8..f3f2217 100644
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -409,8 +409,7 @@ WebSocket.CLOSED = 3;
      * @public
      */
     get () {
-      const listener = this.listeners(method)[0];
-      return listener ? listener._listener ? listener._listener : listener : undefined;
+      return this.listeners(method).find((listener) => listener._listener);
     },
     /**
      * Add a listener for the event.
@@ -419,7 +418,9 @@ WebSocket.CLOSED = 3;
      * @public
      */
     set (listener) {
-      this.removeAllListeners(method);
+      this.listeners(method).forEach((listener) => {
+        if (listener._listener) this.removeListener(method, listener);
+      });
       this.addEventListener(method, listener);
     }
   });
```

jacobbogers:
ok, so you tag a listener with _listener attribute
I see it in your  ` return this.listeners(method).find((listener) => listener._listener);`

and when you "set" you loop over all existing listener and remove the one with _listener attribute

So the hidden 'close' would NOT have the _listener attribute tag

I understand the solution, you "tag" all "user defined" listener with _listener property.

I now see distinction between `addEventListener/addListener` and `removeEventListener/removeListener`

Ok, yes , I think yours is better solution but is good to document why you make exception in 
```if (listener._listener) ``` statement.

lpinca:
The only reason why I'm a bit hesitant about using a specific event is that this event should ideally be emitted only when `clientTracking` is enabled while `close` is already emitted in all cases.

jacobbogers:
Is good too  leverage what is already there,  I suggest add 2 lines of documentation one for the "close" and one for the if statement. Since we talk about "clients" collection, I see you change it too set, my code uses "Array.length" because it is array now, so will soon need to change it to "Set.size" )))  
Can you abstract what kind of collection you use ?

jacobbogers:
Shall i make your suggested changed? then i get some kuddos for finding the bug)))

lpinca:
> Since we talk about "clients" collection, I see you change it too set, my code uses "Array.length" because it is array now, so will soon need to change it to "Set.size" )))
Can you abstract what kind of collection you use ?

The next version will be a major (2.0.0) breaking changes will be documented.

lpinca:
@jacobbogers ping.

lpinca:
This needs a regression test but I'll add it in a bit.
Thanks!

PR #956, opened 2017-01-05
Abort the request if `close` is called while connecting

### LINKED ISSUE #388: WebSocket.close() does not close the connection
```
WebSocket.prototype.close = function(code, data) {
  if (this.readyState == WebSocket.CLOSING || this.readyState == WebSocket.CLOSED) return;
  if (this.readyState == WebSocket.CONNECTING) {
    this.readyState = WebSocket.CLOSED;
    return;
  }
```

So, if I create a client WebSocket and, while connecting, I call `close()` on it then it does nothing. Instead the handshake continues.

I see that `http.Agent` is in use and I do not know how it works. Anyhow I expect that a WebSocket client should be able to close/abort the TCP connection at any time once the GET has been sent and before the 101 arrives.

lpinca:
Currently `WebSocket.prototype.close()` only sets the `readyState` to [`CLOSED`](https://github.com/websockets/ws/blob/8aa65e2f4da33083d1c20e46240aacc4fe40822c/lib/WebSocket.js#L267) if called before receiving a handshake response. Then the connection is eventually closed when the response is [received](https://github.com/websockets/ws/blob/8aa65e2f4da33083d1c20e46240aacc4fe40822c/lib/WebSocket.js#L633-L639).
This is wrong as the `connection` event is still emitted on the server.

This patch aborts the client request if `WebSocket.prototype.close()` is called before receiving a handshake response.

Fixes #388

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -262,10 +262,10 @@ class WebSocket extends EventEmitter {
    */
   close (code, data) {
     if (this.readyState === WebSocket.CLOSED) return;
-
     if (this.readyState === WebSocket.CONNECTING) {
-      this.readyState = WebSocket.CLOSED;
-      return;
+      this._req.abort();
+      this.emit('error', new Error('closed before the connection is established'));
+      return this.finalize(true);
     }
 
     if (this.readyState === WebSocket.CLOSING) {
@@ -374,6 +374,11 @@ class WebSocket extends EventEmitter {
    */
   terminate () {
     if (this.readyState === WebSocket.CLOSED) return;
+    if (this.readyState === WebSocket.CONNECTING) {
+      this._req.abort();
+      this.emit('error', new Error('closed before the connection is established'));
+      return this.finalize(true);
+    }
 
     if (this._socket) {
       this.readyState = WebSocket.CLOSING;
@@ -385,8 +390,6 @@ class WebSocket extends EventEmitter {
       //
       clearTimeout(this._closeTimer);
       this._closeTimer = setTimeout(this._finalize, closeTimeout, true);
-    } else if (this.readyState === WebSocket.CONNECTING) {
-      this.finalize(true);
     }
   }
 }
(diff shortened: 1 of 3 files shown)


lib/WebSocket.js
```diff
@@ -262,10 +262,10 @@ class WebSocket extends EventEmitter {
    */
   close (code, data) {
     if (this.readyState === WebSocket.CLOSED) return;
-
     if (this.readyState === WebSocket.CONNECTING) {
-      this.readyState = WebSocket.CLOSED;
-      return;
+      this._req.abort();
+      this.emit('error', new Error('closed while connecting'));
```
lpinca: This is added to match the browser behavior. I'm not sure if we want this error.
I'm also not sure if the error message is ok, suggestions?

lib/WebSocket.js
```diff
@@ -610,43 +613,32 @@ function initAsClient (address, protocols, options) {
 
   if (agent) requestOptions.agent = agent;
 
-  const req = httpObj.get(requestOptions);
+  this._req = httpObj.get(requestOptions);
+
+  this._req.on('error', (error) => {
+    if (this._req.aborted) return;
```
lpinca: This is to avoid "socket hang up" errors which are emitted when a request is aborted and a response is not received.

We can remove the check as the error would be swallowed anyway but I think being explicit is better here.

lpinca:
@websockets/admin please review when/if you can.

3rd-Eden:
Looks sane to me.

lib/WebSocket.js
3rd-Eden: Well it makes sense, so just keep it.

lpinca: What about the error message? Can you think of something better?

lpinca: Some alternatives:

- "closed before the connection is established"
- "closed before receiving a handshake response"

I have no strong opinions.

3rd-Eden: `closed before the connection is established` is a good one as well.

lpinca: Ok, I'll change it to that, merge this and release `2.0.0-beta.0` if there are no objections.

Nibbler999:
@lpinca This breaks engine.io - if you try to connect to a server which is down you get an infinite loop because engine.io calls close when you emit the error in close.

lpinca:
Should this be fixed in Engine.IO?

lpinca:
I guess Engine.IO is not the only one who calls `close()` or `terminate()` inside an `error` listener and I think this should work without issues.
I'll try to fix this later.

julien-f:
@3rd-Eden Throwing while connecting deviates from the web API in which `close()` can be used to [*closes the  connection attempt*](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#close()).

What's the proper way to abort a connection with `ws`?

lpinca:
@julien-f add a listener for the `'error'` event. The browser invokes `onerror()` when `close()` is called while connecting.

julien-f:
My bad, I'm mistaken, `close()` does not throw.

@lpinca Thanks for your answer and sorry for bothering you :)

PR #962, opened 2017-01-11
Accept hostname lookup `family` option when creating WebSockets

hansonw:
`http.request` accepts a `family` option to specify the IP address family to use when resolving the server hostname. Seems pretty reasonable to accept this from the WebSocket side and forward it over to the `http` request call.

Specifically, this is useful for Facebook since we have IPv6 only development servers with IPv4 proxies (in which case we need websockets to connect to the IPv6 address)

### DIFF
--- a/test/WebSocket.test.js
+++ b/test/WebSocket.test.js
@@ -80,6 +80,17 @@ describe('WebSocket', function () {
         /must be a valid IP: 123.456.789.428/
       );
     });
+
+    it('should accept the family option', function (done) {
+      const wss = new WebSocketServer({ host: '::1', port: ++port }, () => {
+        const ws = new WebSocket(`ws://localhost:${port}`, { family: 6 });
+      });
+
+      wss.on('connection', (ws) => {
+        assert.strictEqual(ws.upgradeReq.connection.remoteAddress, '::1');
+        wss.close(done);
+      });
+    });
   });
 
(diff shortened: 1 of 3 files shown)


test/WebSocket.test.js
```diff
@@ -80,6 +80,16 @@ describe('WebSocket', function () {
         /must be a valid IP: 123.456.789.428/
       );
     });
+
+    it('should accept the family option', function (done) {
+      const wss = new WebSocketServer({ host: '::1', port: ++port }, () => {
+        const ws = new WebSocket(`ws://localhost:${port}`, {family: 6});
```
lpinca: Can you please add spaces inside the options object? `{ family: 6 }`.

test/WebSocket.test.js
```diff
@@ -80,6 +80,16 @@ describe('WebSocket', function () {
...
+      });
```
lpinca: Can you please add a blank line after this?

hansonw:
Done! I think there's an eslint rule for the spaces-in-objects, btw :)

lpinca:
@hansonw yeah :) we are using semistandard and it doesn't enforce that.
We can extend it wit our own rules but it kinda defeat the point of using a popular and premade config.

lpinca:
Thanks!

PR #966, opened 2017-01-12
Add WebSocket compression section

lpinca:
This adds a WebSocket compression section to specify that permessage-deflate should be used only when really necessary.


lpinca:
@websockets/admin do you agree?

3rd-Eden:
Agreed.

PR #968, opened 2017-01-14
Use Object.assign() for send() options

Nibbler999:
This avoids mutating the passed in arguments.

The current behaviour breaks sending if you use the same options object for multiple send() calls. I don't know if anyone relies on the arguments being mutated or in being able to pass in undefined explicitly.


lpinca:
LGTM, it should still work when using `undefined` and I'm +1 for not mutating the options object.
Is there any performance impact? I guess it's irrelevant.

PR #971, opened 2017-01-18
Take into account the data queued in the sender

### LINKED ISSUE #970: Extend bufferedAmount property
The bufferedAmount property is good for your application to control how much messages you can send (to avoid memory overhead), but it works only if you disable perMessageDeflate.

I propose to extend the property by calculating _sender.queue size as well.

Example:

    get bufferedAmount () {
        var amount = 0;
        var senderQueueSize = 0;
        
        if (this._socket) amount = this._socket.bufferSize || 0;
        
        if (this._sender && this._sender.queue) {
          for (i = 0; i < this._sender.queue.length; i += 1) {
            senderQueueSize += this._sender.queue[i].data.length;
          }
        }
        return amount + senderQueueSize;
    }

The best solution is to avoid cycle at all by adding queueSize variable to _sender and add/subtract length when adding a new element to queue.

    get bufferedAmount () {
        var amount = 0;
        var senderQueueSize = 0;
        
        if (this._socket) amount = this._socket.bufferSize || 0;
        return amount + this._sender.queueSize;
    }

lpinca:
This makes the `bufferedAmount` getter take into account the data queued in the sender.

Fixes #970.

cc: @alexions

### DIFF
--- a/lib/Sender.js
+++ b/lib/Sender.js
@@ -24,12 +24,16 @@ class Sender {
    */
   constructor (socket, extensions) {
     this.perMessageDeflate = (extensions || {})[PerMessageDeflate.extensionName];
-    this.firstFragment = true;
-    this.processing = false;
-    this.compress = false;
     this._socket = socket;
-    this.onerror = null;
+
+    this.firstFragment = true;
+    this.compress = false;
+
+    this.processing = false;
+    this.bufferedBytes = 0;
     this.queue = [];
+
+    this.onerror = null;
   }
 
   /**
@@ -86,10 +90,23 @@ class Sender {
    * @public
    */
   ping (data, mask) {
+    var readOnly = true;
+
+    if (data && !Buffer.isBuffer(data)) {
+      if (data instanceof ArrayBuffer) {
+        data = Buffer.from(data);
+      } else if (ArrayBuffer.isView(data)) {
+        data = viewToBuffer(data);
+      } else {
+        data = Buffer.from(data);
+        readOnly = false;
+      }
+    }
+
     if (this.perMessageDeflate) {
-      this.enqueue([this.doPing, data, mask]);
+      this.enqueue([this.doPing, data, mask, readOnly]);
     } else {
-      this.doPing(data, mask);
+      this.doPing(data, mask, readOnly);
     }
   }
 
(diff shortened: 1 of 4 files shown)


lib/Sender.js
```diff
@@ -86,10 +90,23 @@ class Sender {
...
+      if (data instanceof ArrayBuffer) {
+        data = Buffer.from(data);
+      } else if (ArrayBuffer.isView(data)) {
+        data = viewToBuffer(data);
+      } else {
+        data = Buffer.from(data);
+        readOnly = false;
+      }
+    }
+
```
alexions: I'd move this block of code to a separate function, like 

    makeBuffer (data) {
      var readOnly = true;

      if (data && !Buffer.isBuffer(data)) {
        if (data instanceof ArrayBuffer) {
          data = Buffer.from(data);
        } else if (ArrayBuffer.isView(data)) {
          data = viewToBuffer(data);
        } else {
          data = Buffer.from(data);
          readOnly = false;
        }
      }

      return [data, readOnly];
    }

    Usage:
    [data, readOnly] = makeBuffer(data);

lpinca: I also don't like having this code duplication. The reason for not using a separate function is to avoid creating an array.

alexions: Agree. Array can make additional overhead... ok :) Speed is better than duplicates :)

lpinca: It should be benchmarked to see if it actually makes any difference :)

lpinca: Another option would be to add the `readOnly` flag to the returned buffer. For example:

```js
const kReadOnly = Symbol('read-only');

function toBuffer(data) {
  var readOnly = true;

  if (!Buffer.isBuffer(data)) {
    if (data instanceof ArrayBuffer) {
      data = Buffer.from(data);
    } else if (ArrayBuffer.isView(data)) {
      data = viewToBuffer(data);
    } else {
      data = Buffer.from(data);
      readOnly = false;
    }
  }

  data[kReadOnly] = readOnly;
  return data;
}
```

but I'm still not happy with it as that changes the buffer hidden class.

lpinca:
I would like to have a LGTM from at least one of the other @websockets/admin members before merging.

lpinca:
@alexions thanks but I'm -1 on that approach as that proposes to check the things twice (2x `Buffer.isBuffer()`, 2x `ArrayBuffer.isView()`, etc.), first to see if it's safe to treat the data as `readOnly` then to convert it to a buffer. I would rather the duplicated code.

alexions:
ok. Agree that your changes will work faster.

lpinca:
Before this change the whole `if...else`chain was used in 2 functions, now in 3 so I think it's not so bad. I'm just waiting for the approval of another @websockets/admin member :)

lpinca:
I'll merge this tomorrow if there are no objections.

lpinca:
@alexions thanks!

alexions:
@lpinca you're welcome!

PR #991, opened 2017-02-06
Avoid stack overflow when processing payloads with a large number of fragments

Nibbler999:
Sending ~10000 empty fragments in one message can crash the server otherwise.

### DIFF
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -370,7 +370,7 @@ class Receiver {
     }
 
     this.state = START;
-    this.start();
+    process.nextTick(() => this.start());
   }
 
   /**


lpinca:
Yeah, I came up with a similar solution yesterday:

```js
diff --git a/lib/Receiver.js b/lib/Receiver.js
index 4a23311..d9b223e 100644
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -310,13 +310,14 @@ class Receiver {
       if (this.masked) bufferUtil.unmask(data, this.mask);
     }
 
+    this.state = HANDLE_DATA;
+
     if (this.opcode > 0x07) {
-      this.controlMessage(data);
+      process.nextTick(() => this.controlMessage(data));
     } else if (this.compressed) {
-      this.state = HANDLE_DATA;
       this.decompress(data);
     } else if (this.pushFragment(data)) {
-      this.dataMessage();
+      process.nextTick(() => this.dataMessage());
     }
   }
```

it works but I wonder if we can remove recursion altogether.

PR #994, opened 2017-02-08
Security: Use crypto.randomBytes, not Math.random

sgress454:
This is a backport of https://github.com/websockets/ws/pull/832/files to v1.x, so that projects that still need to support Node < 4 can get the security patch.

### DIFF
--- a/lib/Sender.js
+++ b/lib/Sender.js
@@ -6,6 +6,7 @@
 
 var events = require('events')
   , util = require('util')
+  , crypto = require('crypto')
   , EventEmitter = events.EventEmitter
   , ErrorCodes = require('./ErrorCodes')
   , bufferUtil = require('./BufferUtil').BufferUtil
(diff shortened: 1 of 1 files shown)


lib/Sender.js
```diff
@@ -315,10 +315,5 @@ function getArrayBuffer(data) {
 }
 
 function getRandomMask() {
-  return new Buffer([
-    ~~(Math.random() * 255),
-    ~~(Math.random() * 255),
-    ~~(Math.random() * 255),
-    ~~(Math.random() * 255)
-  ]);
+  return require('crypto').randomBytes(4);
```
lpinca: Only a tiny nit: would you mind moving the require at the top of the file along with the others?

sgress454: Done!

lpinca:
Thanks!

PR #1016, opened 2017-02-23
Make `Sender#frameAndSend()` a static method

lpinca:
This transforms `Sender.prototype.frameAndSend()` into a static method (`Sender.frame()`) which can be used to frame data.

Having a function which returns a frame gives developers more fine grained control and allows them to optimize certain operations like a server broadcast where the same frame should be sent to all clients.

### DIFF
--- a/lib/Sender.js
+++ b/lib/Sender.js
@@ -37,6 +37,71 @@ class Sender {
   }
 
   /**
+   * Frames a piece of data according to the HyBi WebSocket protocol.
+   *
+   * @param {Buffer} data The data to frame
+   * @param {Object} options Options object
+   * @param {Number} options.opcode The opcode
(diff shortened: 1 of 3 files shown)


lib/Sender.js
```diff
@@ -37,6 +37,71 @@ class Sender {
...
+    } else if (payloadLength === 127) {
+      target.writeUInt32BE(0, 2, true);
+      target.writeUInt32BE(data.length, 6, true);
+    }
+
+    if (!options.mask) {
+      target[1] = payloadLength;
+      if (merge) {
+        data.copy(target, offset);
+        return [target];
```
lpinca: I'm not too happy with this. I chose to return an array instead of the buffer mostly for consistency with the other case so the return value is always an array but I'm happy to revisit this. Thoughts?

guymguym:
LGTM

PR #1018, opened 2017-02-26
Receiver support binaryType ‘fragments’ …

guymguym:
- emit message fragments array to avoid copyfull concat

Note:
binaryType used to affect before only in WHATWG API emulation
and the optional conversion to ArrayBuffer was done inside MessageEvent.
However in this change the receiver emits messages according to binaryType
so now also EventEmitter listeners added with on(‘message’, data => {})
will get the data according to binaryType.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -3,3 +3,5 @@
 exports.GUID = '258EAFA5-E914-47DA-95CA-C5AB0DC85B11';
 exports.EMPTY_BUFFER = Buffer.alloc(0);
 exports.NOOP = () => {};
+
+exports.BINARY_TYPES = ['nodebuffer', 'arraybuffer', 'fragments'];
--- a/lib/Receiver.js
+++ b/lib/Receiver.js
@@ -28,10 +28,12 @@ class Receiver {
    *
    * @param {Object} extensions An object containing the negotiated extensions
    * @param {Number} maxPayload The maximum allowed message length
+   * @param {String} binaryType The type for binary data, see constants.BINARY_TYPES
    */
-  constructor (extensions, maxPayload) {
+  constructor (extensions, maxPayload, binaryType) {
     this.extensions = extensions || {};
     this.maxPayload = maxPayload | 0;
+    this.binaryType = binaryType || constants.BINARY_TYPES[0];
 
     this.bufferedBytes = 0;
     this.buffers = [];
@@ -352,20 +354,28 @@ class Receiver {
    */
   dataMessage () {
     if (this.fin) {
-      const buf = this.fragments.length > 1
-        ? Buffer.concat(this.fragments, this.messageLength)
-        : this.fragments.length === 1
-          ? this.fragments[0]
-          : constants.EMPTY_BUFFER;
-
+      const fragments = this.fragments;
+      const messageLength = this.messageLength;
       this.totalPayloadLength = 0;
-      this.fragments.length = 0;
+      this.fragments = [];
       this.messageLength = 0;
       this.fragmented = 0;
 
       if (this.opcode === 2) {
-        this.onmessage(buf, { masked: this.masked, binary: true });
+        var data;
+
+        if (this.binaryType === 'nodebuffer') {
+          data = bufferFromFragments(fragments, messageLength);
+        } else if (this.binaryType === 'arraybuffer') {
+          data = new Uint8Array(bufferFromFragments(fragments, messageLength)).buffer;
+        } else {
+          data = fragments;
+        }
+
+        this.onmessage(data, { masked: this.masked, binary: true });
       } else {
+        const buf = bufferFromFragments(fragments, messageLength);
+
         if (!isValidUTF8(buf)) {
           this.error(new Error('invalid utf8 sequence'), 1007);
           return;
@@ -507,6 +517,22 @@ class Receiver {
       if (cb) cb();
     }
   }
+}
+
+/**
+ * Make a buffer from a list of fragments.
+ * Optimized for the common case of a single fragment in order to
+ * avoid copyfull concat and simply return the fragment buffer.
+ *
+ * @param {Buffer[]} fragments
+ * @param {Number} messageLength
+ * @return {Buffer}
+ * @private
+ */
+function bufferFromFragments (fragments, messageLength) {
+  if (fragments.length === 1) return fragments[0];
+  if (fragments.length > 1) return Buffer.concat(fragments, messageLength);
+  return constants.EMPTY_BUFFER;
 }
 
 module.exports = Receiver;
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -53,7 +53,7 @@ class WebSocket extends EventEmitter {
     this.protocol = '';
 
     this._finalize = this.finalize.bind(this);
-    this._binaryType = 'nodebuffer';
+    this._binaryType = constants.BINARY_TYPES[0];
     this._finalizeCalled = false;
     this._closeMessage = null;
     this._closeTimer = null;
@@ -98,10 +98,17 @@ class WebSocket extends EventEmitter {
   }
 
   set binaryType (type) {
-    if (type === 'arraybuffer' || type === 'nodebuffer') {
-      this._binaryType = type;
-    } else {
-      throw new SyntaxError('unsupported binaryType: must be either "nodebuffer" or "arraybuffer"');
+    // silently ignore unsupported types
+    if (constants.BINARY_TYPES.indexOf(type) < 0) {
+      return;
+    }
+
+    this._binaryType = type;
+
+    // update the receiver if already created,
+    // if not then it will take the value once created
+    if (this._receiver) {
+      this._receiver.binaryType = type;
     }
   }
 
@@ -116,7 +123,7 @@ class WebSocket extends EventEmitter {
     socket.setTimeout(0);
     socket.setNoDelay();
 
-    this._receiver = new Receiver(this.extensions, this.maxPayload);
+    this._receiver = new Receiver(this.extensions, this.maxPayload, this.binaryType);
     this._sender = new Sender(socket, this.extensions);
     this._ultron = new Ultron(socket);
     this._socket = socket;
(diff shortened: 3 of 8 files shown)


guymguym:
@lpinca Hi, would love to get your review.
See the Note I wrote above - let me know if you think this is a breaking change or not, and if you think it's a good direction.
Thanks

lib/Constants.js
```diff
@@ -3,3 +3,6 @@
 exports.GUID = '258EAFA5-E914-47DA-95CA-C5AB0DC85B11';
 exports.EMPTY_BUFFER = Buffer.alloc(0);
 exports.NOOP = () => {};
+
+exports.DEFAULT_BINARY_TYPE = 'nodebuffer';
```
lpinca: I would get rid of this and use `BINARY_TYPES[0]` instead where needed.

lib/Receiver.js
```diff
@@ -352,20 +354,28 @@ class Receiver {
...
+      const messageLength = this.messageLength;
       this.totalPayloadLength = 0;
-      this.fragments.length = 0;
+      this.fragments = [];
       this.messageLength = 0;
       this.fragmented = 0;
 
       if (this.opcode === 2) {
-        this.onmessage(buf, { masked: this.masked, binary: true });
+        let data;
```
lpinca: Please use `var` here due to possible deoptimizations.

lib/Receiver.js
```diff
@@ -352,20 +354,28 @@ class Receiver {
...
+
+        if (this.binaryType === 'nodebuffer') {
+          data = bufferFromFragments(fragments, messageLength);
+        } else if (this.binaryType === 'arraybuffer') {
+          data = new Uint8Array(bufferFromFragments(fragments, messageLength)).buffer;
```
lpinca: I'm not sure about the "semverness" of this change as previously conversion to `ArrayBuffer` was only done when using the `EventTarget` interface.

I like this as it makes sense and we also had a couple of issues opened because people were surprised to get a `Buffer` when they specified `binaryType` to be `arraybuffer` so this change can also be seen as a bug fix.

lib/Receiver.js
```diff
@@ -352,20 +354,28 @@ class Receiver {
...
+        } else { // if (this.binaryType === 'fragments') {
```
lpinca: I don't think this comment is necessary. `data = fragments` is clear enough imo.

lib/WebSocket.js
```diff
@@ -98,10 +98,15 @@ class WebSocket extends EventEmitter {
   }
 
   set binaryType (type) {
-    if (type === 'arraybuffer' || type === 'nodebuffer') {
-      this._binaryType = type;
-    } else {
-      throw new SyntaxError('unsupported binaryType: must be either "nodebuffer" or "arraybuffer"');
+    if (constants.BINARY_TYPES.indexOf(type) < 0) {
+      throw new SyntaxError('unsupported binaryType: must be either "' +
```
lpinca: This is a breaking change as the error message is changed but I **think** there is not much code that relies on this. We can try to search for this error message on Github.

I think it's better to remove the error completely. The browser no longer throw this error if an invalid `binaryType` is specified. We should do the same and fail silently if an invalid `binaryType` is supplied.

test/Receiver.test.js
```diff
@@ -827,4 +827,84 @@ describe('Receiver', function () {
...
+
+/**
+ * Adds a list of binary fragments to the receiver prefixing each one
+ * with info byte and length.
+ *
+ * @param {Receiver} receiver
+ * @param {Buffer[]} frags
+ * @private
+ */
+function addBinaryFragments (receiver, frags) {
```
lpinca: Maybe we can get rid of this and use `Sender.frame()` if #1016 is merged. It will be easier to grok.

lpinca:
@guymguym I've requested some changes but LGTM.

test/Receiver.test.js
lpinca: Merged #1016 so you can rebase and use that.

lib/WebSocket.js
guymguym: In chrome it ignores unsupported values but with a console.warn (see below)
Do you think we should do console.warn too?

image

lib/Constants.js
guymguym: np

lib/Receiver.js
guymguym: np. 
For my education - would you mind explaining why this is a case for deoptimization?

lib/Receiver.js
guymguym: Yeah, it makes sense to me too, this is why I tried to suggest it this way.
If you want to play extra safe for the other folks then best release as a new major version.

lib/Receiver.js
guymguym: np

lib/WebSocket.js
lpinca: Firefox doesn't warn, it just fails silently. What does WHATWG spec say?

lib/Receiver.js
lpinca: I know for sure that [compund let assignment](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax) are not optimizable on older V8 versions.

I also saw deoptimizations for non compound assignment when rewriting the parser so for now I prefer to stick with `var` when in doubt. This same is also done in Node.js core.

lib/WebSocket.js
lpinca: https://html.spec.whatwg.org/multipage/comms.html#the-websocket-interface

> When a WebSocket object is created, its binaryType IDL attribute must be set to the string "blob". On getting, it must return the last value it was set to. On setting, the user agent must set the IDL attribute to the new value.

> Note: This attribute allows authors to control how binary data is exposed to scripts. By setting the attribute to "blob", binary data is returned in Blob form; by setting it to "arraybuffer", it is returned in ArrayBuffer form. User agents can use this as a hint for how to handle incoming binary data: if the attribute is set to "blob", it is safe to spool it to disk, and if it is set to "arraybuffer", it is likely more efficient to keep the data in memory. Naturally, user agents are encouraged to use more subtle heuristics to decide whether to keep incoming data in memory or not, e.g. based on how big the data is or how common it is for a script to change the attribute at the last minute. This latter aspect is important in particular because it is quite possible for the attribute to be changed after the user agent has received the data but before the user agent has fired the event for it.

This is all so it should be ok to just remove the error.

test/Receiver.test.js
guymguym: done

lib/Receiver.js
guymguym: Thanks. That's good to know.

lib/WebSocket.js
guymguym: done

guymguym:
@lpinca Fixed all comments. Thanks for the review!

lpinca:
Can you please run the parser benchmark and post the results before and after this change? Thanks.

guymguym:
Before:
```
$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.

$ node bench/parser.benchmark.js 
ping message x 1,516,912 ops/sec ±0.66% (90 runs sampled)
ping with no data x 14,433,226 ops/sec ±1.00% (88 runs sampled)
close message x 12,816,502 ops/sec ±1.34% (87 runs sampled)
masked text message (20 bytes) x 1,010,852 ops/sec ±1.13% (88 runs sampled)
binary data (125 bytes) x 1,284,809 ops/sec ±1.29% (90 runs sampled)
binary data (65535 bytes) x 186,970 ops/sec ±1.17% (87 runs sampled)
binary data (200 KiB) x 69,283 ops/sec ±1.10% (86 runs sampled)
binary data (1 MiB) x 12,406 ops/sec ±0.90% (89 runs sampled)

$ node bench/parser.benchmark.js 
ping message x 1,500,888 ops/sec ±0.68% (89 runs sampled)
ping with no data x 13,664,804 ops/sec ±1.78% (83 runs sampled)
close message x 12,325,815 ops/sec ±1.19% (85 runs sampled)
masked text message (20 bytes) x 945,751 ops/sec ±0.98% (91 runs sampled)
binary data (125 bytes) x 1,219,381 ops/sec ±1.06% (87 runs sampled)
binary data (65535 bytes) x 184,835 ops/sec ±1.39% (87 runs sampled)
binary data (200 KiB) x 69,053 ops/sec ±1.00% (86 runs sampled)
binary data (1 MiB) x 12,375 ops/sec ±0.96% (87 runs sampled)
```

After:
```
$ git checkout guymguym-receiver-fragments 
Switched to branch 'guymguym-receiver-fragments'
Your branch is up-to-date with 'origin/guymguym-receiver-fragments'.

$ node bench/parser.benchmark.js 
ping message x 1,537,659 ops/sec ±0.98% (89 runs sampled)
ping with no data x 14,070,824 ops/sec ±0.85% (87 runs sampled)
close message x 12,873,083 ops/sec ±1.19% (85 runs sampled)
masked text message (20 bytes) x 944,098 ops/sec ±0.89% (87 runs sampled)
binary data (125 bytes) x 1,404,478 ops/sec ±1.30% (85 runs sampled)
binary data (65535 bytes) x 188,281 ops/sec ±1.31% (84 runs sampled)
binary data (200 KiB) x 70,403 ops/sec ±1.05% (86 runs sampled)
binary data (1 MiB) x 12,387 ops/sec ±0.91% (87 runs sampled)

$ node bench/parser.benchmark.js 
ping message x 1,528,134 ops/sec ±0.96% (90 runs sampled)
ping with no data x 14,189,599 ops/sec ±0.87% (89 runs sampled)
close message x 12,886,198 ops/sec ±1.64% (88 runs sampled)
masked text message (20 bytes) x 1,048,379 ops/sec ±1.38% (89 runs sampled)
binary data (125 bytes) x 1,412,051 ops/sec ±0.91% (89 runs sampled)
binary data (65535 bytes) x 189,763 ops/sec ±1.09% (89 runs sampled)
binary data (200 KiB) x 70,405 ops/sec ±1.15% (86 runs sampled)
binary data (1 MiB) x 12,353 ops/sec ±0.86% (86 runs sampled)
```

guymguym:
Travis failed for node v4 but I think it's transient.
I don't have the option to re-trigger without commit, do you?

lpinca:
Yes, test seems to be flaky. The timeout is probably too short. I'll restart.

I'll keep this open for a while so others can weigh in. If there are no objections I'll merge this tomorrow.

lpinca:
Thanks!

PR #1026, opened 2017-03-01
Avoid using `Buffer.concat()`

lpinca:
This patch adds a new function to merge an array of buffers and replaces all occurrences of `Buffer.concat()` with it.

The reason for this change is that `Buffer.concat()` [checks](https://github.com/nodejs/node/blob/87a039d721b7e9fd73365774ac3e27166e4e1f6e/lib/buffer.js#L333) that every element in the array is an instance of `Buffer` and this makes it perform worse.

### DIFF
--- a/lib/BufferUtil.js
+++ b/lib/BufferUtil.js
@@ -6,27 +6,32 @@
  * MIT Licensed
  */
 
+/**
+ * Merges an array of buffers into a new buffer.
+ *
+ * @param {Buffer[]} list The array of buffers to concat
+ * @param {Number} totalLength The total length of buffers in the list
+ * @return {Buffer} The resulting buffer
+ * @public
+ */
+const concat = (list, totalLength) => {
+  const target = Buffer.allocUnsafe(totalLength);
+  var offset = 0;
+
+  for (var i = 0; i < list.length; i++) {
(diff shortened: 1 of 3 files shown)


lpinca:
We should find a way to remove all `instanceof` checks in the code base. It's unbelievably slow.

3rd-Eden:
Is it slower than Object.prototype.toString.call(item)?

3rd-Eden:
And as far as I could see there's only 1, which is in sender.

lpinca:
`Buffer.isBuffer()` also uses `instanceof` internally but yes I think we only have those checks in the sender.

lpinca:
The fastest way is probably to use `thing.constructor === Buffer` but it will not work if `thing` is a subclass.

lpinca:
```
Object.prototype.toString() x 1,927,356 ops/sec ±0.99% (82 runs sampled)
Buffer.isBuffer() x 3,755,305 ops/sec ±0.88% (85 runs sampled)
process.binding('util').isUint8Array x 23,694,865 ops/sec ±1.97% (82 runs sampled)
.constructor x 52,849,107 ops/sec ±2.57% (79 runs sampled)
```

PR #1036, opened 2017-03-08
Error message for unsupported protcol version

Dorian:
I was playing with making a websocket server and connecting to it using `wscat`, so I did: `wscat -c ws://0.0.0.0 -p 3000` and I got:

> Error: unsupported protocol version

That made me think my webserver was wrong when in fact `-p` is the protocol version and not the port, e.g. `wscat -c ws://0.0.0.0:3000` works.

So I improved the error message:

> Error: unsupported protocol version 3000 (supported versions: 8, 13)

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -21,7 +21,8 @@ const Receiver = require('./Receiver');
 const Sender = require('./Sender');
 
 const closeTimeout = 30 * 1000; // Allow 30 seconds to terminate the connection cleanly.
-const protocolVersion = 13;
+const protocolVersions = [8, 13];
+const protocolVersion = protocolVersions[1];
 
 /**
  * Class representing a WebSocket.
@@ -527,8 +528,11 @@ function initAsClient (address, protocols, options) {
     ca: null
   }, options);
 
-  if (options.protocolVersion !== 8 && options.protocolVersion !== 13) {
-    throw new Error('unsupported protocol version');
+  if (protocolVersions.indexOf(options.protocolVersion) === -1) {
+    throw new Error(
+      `unsupported protocol version: ${options.protocolVersion} ` +
+      `(supported versions: ${protocolVersions.join(', ')})`
+    );
   }
 
   this.protocolVersion = options.protocolVersion;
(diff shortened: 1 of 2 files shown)


Dorian:
Not sure about that coveralls report, seems completely unrelated

lpinca:
Technically this is a breaking change as it changes an error message.

lib/WebSocket.js
```diff
@@ -22,6 +22,7 @@ const Sender = require('./Sender');
 
 const closeTimeout = 30 * 1000; // Allow 30 seconds to terminate the connection cleanly.
 const protocolVersion = 13;
+const supportedProtocolVersions = [8, 13];
```
lpinca: Would you mind merging this line with the previous one?

```js
const protocolVersions = [8, 13];
```

and use 

```js
protocolVersion: protocolVersions[1]
```

on line 510?

lib/WebSocket.js
```diff
@@ -527,8 +528,10 @@ function initAsClient (address, protocols, options) {
     ca: null
   }, options);
 
-  if (options.protocolVersion !== 8 && options.protocolVersion !== 13) {
-    throw new Error('unsupported protocol version');
+  if (supportedProtocolVersions.indexOf(options.protocolVersion) === -1) {
+    let versionError = `unsupported protocol version: ${options.protocolVersion}`;
+    versionError += ` (supported versions: ${supportedProtocolVersions.join(', ')})`;
+    throw new Error(versionError);
```
lpinca: Can you please rewrite this without using the `versionError` variable?

```js
throw new Error(
  `unsupported protocol version: ${options.protocolVersion} ` +
  `(supported versions: ${protocolVersion.join(', ')})`
);
```

lpinca:
I don't want to keep this open indefinitely and version 3 is not on sight. I think that very few people (if any) rely on this error message so I'll merge this once comments are addressed and release with the next version.

Dorian:
@lpinca Thanks for the review, I updated the code and I think this could help a lot of people setting up / trying websockets and trying to make it work

lpinca:
Thanks!

PR #1056, opened 2017-03-29
Change "const client of" to "let client of"

colinligertwood:
All but the most recent versions of firefox explode on the
use of for (const variable of ...). Switching declaration to
let solves it.

My team discovered this after updating npm modules of a
project which makes use of Autobahn.js.

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -104,7 +104,7 @@ class WebSocketServer extends EventEmitter {
     // Terminate all associated clients.
     //
     if (this.clients) {
-      for (const client of this.clients) client.terminate();
+      for (let client of this.clients) client.terminate();
     }
 
     const server = this._server;


lpinca:
I don't understand. This is made to run on Node.js so there is no need for this change. You can't run the WebSocket server in the browser anyway.

colinligertwood:
Understood. The issue I'm having is WebsocketServer is being bundled in my project with Autobahn.js for the browser, and then fails in Firefox, which until recently considered the for const of invalid.. If you'd rather keep the const, which I also agree is valid, I'll try and find another solution.

lpinca:
I find it hard to believe that Autobahn.js is bundling this lib for the browser. Can you show me where?
It seems that they are correctly [excluding](https://github.com/crossbario/autobahn-js/blob/5309a51f63d964f89599aeb799110ccd094895d4/package.json#L32) ws from the bundle.

colinligertwood:
Indeed. It's being improperly included in the build by Meteor which ignores 'browser' in package.json if it's an object, as is the case with Autobahn. It's a known Meteor issue. https://github.com/meteor/meteor/issues/6890

As I said before, If you'd rather keep the const, I'll find another solution.

lpinca:
@colinligertwood I'm only trying to understand why this issue happens.
If this is a very edge case and there is a not too complex workaround I would prefer to keep `const`. If not feel free to reopen. The change is trivial and shouldn't cause any harm.

colinligertwood:
Thanks, I've managed to get around it for the time being by simply removing the ws library from the project. It results in a compile time warning, but it still builds and works fine. Since, as you pointed out, it shouldn't be loaded into the browser anyway, and the firefox issue is itself a bug that's been since fixed, I'm happy call it and edge case and work around it for a while.

PR #1059, opened 2017-04-02
Add support for Node.js 4.1.0 - 4.4.7

lpinca:
It seems that in some cases upgrading to v2 is not possible due to the minimum required Node.js version (v4.5.0). See for example the discussions in https://github.com/maxogden/websocket-stream/pull/106 and https://github.com/maxogden/websocket-stream/issues/108.

This patch makes `ws@2` works with all v4 versions.

It is tested in v4.1.0 because the previous release v4.0.0 has known bugs and tests fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,6 @@ node_js:
   - "7"
   - "6"
   - "4"
+  - "4.1.0"
 after_success:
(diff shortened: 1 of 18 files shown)


3rd-Eden:
LGTM.

PR #1062, opened 2017-04-04
WebSocketServer path as callback function

stefanpoeter:
Hi everybody,

this pull request changes the behaviour for the path attribute on the WebSocketServer. The path can also be a callback function to check the request. In my case for example a path like `/some/:key/path` can be used to create sockets for variable resources.

For example:

```` 
  ...
    this._pattern = new require('url-patter')('/some/:key/path)
  ...
    let customShouldHandle (req) {
      const url = require('url').parse(req.url).pathname
      const match = this._pattern.match(url)

      if (!match) {
        return false
      }

      if (!req.params) {
        req.params = {}
      }
      req.params.id = match.id
      return true
  }

  ...
  let server = new WebSocket.Server({ 
    server: httpServer, 
    path: this.customShouldHandle.bind(this) 
  })
  ...
````

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -133,7 +133,15 @@ class WebSocketServer extends EventEmitter {
    * @public
    */
   shouldHandle (req) {
-    if (this.options.path && url.parse(req.url).pathname !== this.options.path) {
+    if (!this.options.path) {
+      return true;
+    }
+
+    if (typeof this.options.path === 'function') {
+      return this.path(req);
+    }
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by lpinca** [comment]
Why don't you override `WebSockeServer.prototype.shouldHandle()`? That method has been explicitly added for cases where a custom logic is required.

```js
WebSocketServer.prototype.shouldHandle =  function (req) {
  // Add your own logic and return `true` or `false` accordingly.
};
```

stefanpoeter:
Well that is way smarter than my approach :+1:  Thanks @lpinca

Issue #985, opened 2017-02-02
browserify --node fails because peer modules are required (bufferutil, utf-8-validate)


gabssnake:
Hello, thank you for your work with this project!

While not using the compiled optional modules and using `browserify --node` to concatenate files, I get the following message :

```
> browserify --node --entry index.js

Error: Cannot find module 'bufferutil' from '/project/node_modules/ws/lib'
    at /project/node_modules/resolve/lib/async.js:47:17
    at process (/project/node_modules/resolve/lib/async.js:174:43)
    at ondir (/project/node_modules/resolve/lib/async.js:189:17)
    at load (/project/node_modules/resolve/lib/async.js:70:43)
    at onex (/project/node_modules/resolve/lib/async.js:93:31)
    at /project/node_modules/resolve/lib/async.js:23:47
    at FSReqWrap.oncomplete (fs.js:123:15)
```

`lib/BufferUtil.jsL9` tries to load the module like this : 
```
try {
  module.exports = require('bufferutil');
} catch (e) {
  module.exports = require('./BufferUtil.fallback');
}
```

The following workaround worked in my case (tell browserify to ignore the modules):
```
> browserify --node --external bufferutil --external utf-8-validate --entry index.js
```

lpinca:
You can probably tell browserify to exclude those module from the bundle.

lpinca:
Closing, please comment or reopen if needed.

Issue #719, opened 2016-05-01
Why are bufferutils and utf-8-validate dev-dependencies?


CxRes:
I am sorry if this sounds like a support request but it causes warnings when I try to bundle my app with ws as a dependency. Webpack sees the require statements and is then upset that it cannot find the two libs because obviously they are not installed. 

I know there are fallbacks for the two concerned libs but wouldn't it be better to define them as dependencies and let npm install them and something like webpack the being able to find and bundle them?

3rd-Eden:
They are devDependencies because they are optional binary dependencies. Adding them as optionalDependencies didn't work because npm still cuased the installation to fail when it failed to compile. 

You should completely ignore this library in your webpack build in order to make it work. There's no client side logic here that can be bundled anyways.

Issue #659, opened 2016-01-08
Dependencies bufferutil and utf-8-validate are listed as devDependencies


TheLudd:
My app has this library as a child dependency. I can currently not start it since I get errors saying that module `bufferutil` and `utf-8-validate` cannot be found. The errors originate from this module I notice that the two dependencies are listed as devDependencies. Could they be included as regular dependencies?

3rd-Eden:
No, they are dev dependencies, nothing in this library depends directly on it. The error is somewhere in your own code or the library that consumes this library.

Arnout Kazemier

janhoogeveen:
For what it's worth, I had the same trouble in my custom version of https://github.com/erikras/react-redux-universal-hot-example and removing everything related to Socket.IO got rid of all the errors. Maybe worth investigating if they have new minor/patch versions released that got automatically updated.

TheLudd:
Yes I now understand that the error occurs when browserify is used with this project. It does not handle the try-catch around the non existing dependencies. Is there a good solution for handling this?

janhoogeveen:
Check https://github.com/socketio/socket.io-client/issues/933

3rd-Eden:
@TheLudd It's really a bug in browserify for not understanding conditional imports. 

Anyways, i'm going to close this bug for now. If there are concrete fixes that I can do in this module to make bundling easier, feel free to give me a poke.

wayofthefuture:
Just installed 3.0.0, can't even get it to build with webpack. Same error.

import WebSocket from 'ws';

This is client-side browser code. Am I not suppose to use the 'ws' lib client-side? Should I use the native browser implementation?

Is this incorrect?
Thanks for your help.

3rd-Eden:
This is a node client, not a front-end library.

wayofthefuture:
@3rd-Eden So I should just use native websocket?

3rd-Eden:
Yes

wayofthefuture:
Thank you for your time.

fabcotech:
I just found a solution.

The problem has been described above : those two libraries are not at all required by `ws` module. So the problem is that some bundlers (mine is rollup) still try to resolve it, and if they do not resolve it, you have a `Cannot find module 'bufferutil'` or similar error at runtime.

My simple solution was simply to create fake modules in my `node_modules` folder.

> mkdir ./node_modules/bufferutil/ && touch ./node_modules/bufferutil/index.js
[...]

PR #1066, opened 2017-04-05
Added a 'browser' field in package.json to ignore unused modules

dgrcode:
I was having warnings when using webpack to bundle some code that used `ws` in the client. I checked that `ws` was causing the warnings and I also found the issues #659, #719 and #985 closed in the repo. At least one of them was using webpack and other browserify.

Searching about how to ignore a specific module when bundling, I found [this post][1], and it just solved the issues. There are no warnings after including those lines in the package.json

[1]: https://github.com/defunctzombie/package-browser-field-spec

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,5 +38,9 @@
     "mocha": "~3.2.0",
     "nyc": "~10.2.0",
     "utf-8-validate": "~3.0.0"
+  },
+  "browser": {
+    "utf-8-validate": false,
+    "bufferutil": false
   }
 }


3rd-Eden:
Can't we just do `browser: false`instead? Because we don't have browser files.

lpinca:
I agree with @3rd-Eden.

When bundling for the browser `ws` should be replaced with a shim. `websocket-stream` for example uses this [shim](https://github.com/maxogden/websocket-stream/blob/92871ebf2b0a3407376c5faec1ffcfd38acc14ab/package.json#L42).

When bundling for node (which only makes sense in few cases e.g. AWS lambda) only `bufferutil` and `utf-8-validate` should be excluded and ideally they should be still used if available.

I think it would be better to leave this to the user as it's not clear how `ws` should be bundled.

dgrcode:
@3rd-Eden I've just tried with `browser: false` and it throws the warnings. I don't actually understand the logic behind the browser field, but I think its value can only be a string or an object.

If the object is a string, this is the path of the file that will replace the module (in this case `ws`) when required for the browser.

If the value is an object, each `key` sub-module will be replaced with the file at its `value`, or ignored if it is `false`.

@lpinca I'm using it with node, but the warning only shows when using it for the client. I only know the warning that I get, surely you have a bigger picture perspective.

Thanks both of you for checking it that fast!

lpinca:
@dgrcode If I understand correctly you have a project that should work in both Node.js and the browser. To make it work in the browser you are using a module bundler. Is this correct?

dgrcode:
@lpinca Correct.

I'm using `new WebSocket.Server()` for node, and `new WebSocket()` for the client. I am using webpack. I can share the config if you want.

In my project, I want to use WebSockets instead of AJAX calls, so from the server, I plan to send data to the client and vice-versa. So far I am facing a CORS issue, but that's another story.

lpinca:
Ok then I think what you need is to shim `ws` for the browser like `websocket-stream` and other projects do. In your `package.json` add this

```json
{
  "browser": {
    "ws": "./ws-browser.js"
  }
}
```

where `ws-browser.js` is

```js
'use strict';

var ws = null;

if (typeof WebSocket !== 'undefined') {
  ws = WebSocket;
} else if (typeof MozWebSocket !== 'undefined') {
  ws = MozWebSocket;
} else {
  ws = window.WebSocket || window.MozWebSocket;
}

module.exports = ws;
```

Browserify and webpack should work out of the box with this config.

dgrcode:
This actually fixes everything when bundling.

However, the following code throws an error.

```js

const ws = new WebSocket('ws://www.host.com/path', {
  perMessageDeflate: false
});
```

> Failed to construct 'WebSocket': The subprotocol '[object Object]' is invalid

I see that the snippet in ws-browser.js is telling the browser to use the native WebSocket if available. After I change my client code everything works perfectly. Even the CORS issue that I was facing is solved now! :smiley:

Is this a good approach? I mean using native WebSockets instead of `ws`.

Thanks for the solution! I'll close the PR then.

lpinca:
> Is this a good approach? I mean using native WebSockets instead of ws.

It is the only approach. There is no way to get access to a raw TCP socket socket in the browser using JavaScript, so there is not way to "browserify" `ws` or any other WebSocket implementation.

WRT the error you can work around that. You should only pass the options object if you are using `ws`. One way to do this is to check the `WebSocket` constructor `length`. It will be `3` for `ws` and `1` for the native `WebSocket` object.

```js
if (WebSocket.length === 3) {
  // It's `ws`, pass the options object
} else {
  // It's the native WebSocket contructor. Don't pass the options object.
}
```

dgrcode:
Great! I have it running already and no warnings when bunlding with webpack.

Thanks!

lpinca:
Another possible way is:

```js
var isBrowser = global.WebSocket && process.title === 'browser';

if (isBrowser) {
  // You can't use the options object.
} else {
  // You can use the options object.
}
```

dgrcode:
After a few days working with WebSockets, I came back to the docs and I realised why I tried to use this library in the browser, instead of the native WebSockets. When I read "client" I understood "browser", but now I think that "client" means the back end with the role of the client in the WebSocket communication.

I know that this was caused only by my lack of knowledge of WebSockets, but do you think we could add a line at the docs stating something like:

> **Note for use in the browser**: you should use the native WebSocket. The _client_ in the docs are a reference to a back end with the role of a client in the WebSocket communication.

Maybe I'm still wrong and this also not the meaning of _client_ in the docs.

What do you think?

lpinca:
Happy to improve the documentation. If it is confusing then it is not good enough.
The difference is in server clients and browser clients. In the former case you should use `ws` or any other WebSocket client (it doesn't matter what language). In the browser instead you can only use the native WebSocket object.

Feel free to send a PR to clarify this.

PR #1070, opened 2017-04-06
Make all hooks have access to the upgrade request

### LINKED ISSUE #525: WS server: Append `req` to argumentlist when emitting 'headers' or save response headers
When emitting the `headers` event in [WebSocketServer.js](https://github.com/websockets/ws/blob/2d0e338af86e1a04ceb63a0a77a4432c3b72532f/lib/WebSocketServer.js#L231) no context is passed and since response headers are not saved anywhere subsequent code cannot act on the response headers.

Because of this, there is no way to either dynamically create response headers depending on the connection info (since we do not have access to that in the `headers` listener) or later on act on whichever headers were sent.

Here is one use case:

> We have a client which communicates with a server through both WS and REST. Initial authentication happens on the WS server, which is expensive cpu-cycle wise (many hashing rounds etc.). Because of that we create a session ID which can be used for requests against the REST server, where authentication is a matter of doing a simple lookup.
> To set that session ID we would like to send a cookie with the upgrade response headers.
> 
> After that `verifyClient()` should remember that the ID belongs to an authenticated user id (if that's the case).
> 
> None of this is of course possible, because async operations are not possible in the `headers` listener and `verifyClient()` has no clue which session ID was sent.

I propose changing `self.emit('headers', headers);` to `self.emit('headers', headers, req);`. This change would be minimal, backwards compatible, and allows a connection to be identifiable throughout the upgrade process.

EDIT: When writing the use case, I misunderstood the order of the `headers` event and `verifyClient()`. We of course run `verifyClient()` first and _then_ emit the `headers` event.

### LINKED ISSUE #683: request on('headers') get cookie in WS fix bug lib ws
```
./node_modules/ws/lib/WebSocketServer.js  
string 243

`self.emit('headers', headers);`
change to 
`self.emit('headers', headers, req.headers);`

req header  for request cookie 
```

WebSocketServer.on("headers", function (headers, req) {
 console.log( req.cookie)
})

### LINKED ISSUE #783: Different protocols depending on endpoint
My websocket server needs to support two different protocols depending on which HTTP endpoint was used to create the websocket. Unfortunately I don't seem to have access to the upgrade request in the `handleProtocols()` function so all I can do is accept both protocols or parse the protocols header myself in `verifyClient()` and reject them there already.

Is there a better solution for this? Should `handleProtocols()` have access to the upgrade request?

lpinca:
This adds the upgrade request to the argument list when emitting the `headers` event and also pass it as second argument to the `handleProtocols()` hook.

Closes #787
Fixes #783
Fixes #683
Fixes #525

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -51,10 +51,11 @@ if `verifyClient` is provided with two arguments then those are:
 
 
 If `handleProtocols` is not set then the handshake is automatically accepted,
-otherwise the function takes a single argument:
+otherwise the function takes two arguments:
 
 - `protocols` {Array} The list of WebSocket subprotocols indicated by the
   client in the `Sec-WebSocket-Protocol` header.
+- `request` {http.IncomingMessage} The client HTTP GET request.
 
 If returned value is `false` then the handshake is rejected with the HTTP 401
 status code, otherwise the returned value sets the value of the
@@ -100,6 +101,7 @@ Emitted when an error occurs on the underlying server.
 ### Event: 'headers'
 
 - `headers` {Array}
+- `request` {http.IncomingMessage}
 
 Emitted before the response headers are written to the socket as part of the
 handshake. This allows you to inspect/modify the headers before they are sent.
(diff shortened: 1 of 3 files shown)


3rd-Eden:
LGTM'd

PR #1078, opened 2017-04-11
Added a note about browser usage in the Readme

dgrcode:
This addition is to make clearer to confused users that they should be using the native WebSocket on the browser.


3rd-Eden:
Thanks for your contribution.

Issue #1081, opened 2017-04-19
Get response headers from client?


jwalton:
Would be nice if there was a 'headers' event on the client when we get the [upgrade event here](https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L655).

lpinca:
How would you use it? I'm curious to know why you need it.

jwalton:
I am essentially building a transparent websocket proxy.  So, an upgrade
request comes in, I make a new WebSocket connection to my target host.
When the upgrade reply comes back, though, if there are any extra header
fields in the reply I should copy them into the upgrade reply to the
client.  If there was a set-cookie header in the upgrade reply, for
example, it should get passed on.  (I mean, it would be a bit weird for
there to be a set-cookie headed, but if it's there, I should forward it.)

jwalton:
Actually, RFC 6455 s2 explicitly states "Additional header fields may also
be present, such as cookies," so I guess it isn't that weird.  The other
obvious use case here is parsing cookies from a web site, then.

lpinca:
If I understand correctly you are using `ws` to do the actual proxying. You use `ws` to send the upgrade request to the upstream server and want to send back the response to the client that initiated the request. Is this correct? If so I think using `ws` is not a good idea.

jwalton:
Yeah, `ws` is definitely not the perfect solution here.  I *could* just proxy the raw bytes from the request straight through to the server (possibly I'd need to rewrite some headers, but even that is pretty easy - I could just handle the upgrade event straight from node).  But, for my use case I also want to be able to inspect each websocket message as it goes through, or be able to filter out specific messages.  In a perfect world I'd pass the raw bytes through and I'd have some kind of "ws parser" library that I could for the data to.  `ws` seemed like an easy "this will get me started" solution, though, and it's working pretty well aside from the fact that I need to do some serious hackery to get the response headers.  :)

But, I mean, you provide an API for inspecting and modifying the response headers on the server side, only makes sense you'd have one on the client side too.

lpinca:
Why not using a lib that is specifically designed to do this? Like https://github.com/nodejitsu/node-http-proxy or similar?

jwalton:
node-http-proxy is awesome, and I've used it before, but node-http-proxy doesn't meet my needs for a few reasons:

* AFAIK node-http-proxy knows nothing about the contents of the websocket frames being passed back and forth - I can't inspect the contents of the websocket messages, or alter them.
[...]

PR #1082, opened 2017-04-20
#1081: Add headers event to WebSocket.

jwalton:
Adds headers event.

The only thing that's slightly weird about this is that WebSocketServer emits an array of headers, and this emits an object of headers.  We could emit the [rawHeaders](https://nodejs.org/api/http.html#http_message_rawheaders) here, and then they'd match... But most people probably want the headers object anyways.

### DIFF
--- a/test/WebSocket.test.js
+++ b/test/WebSocket.test.js
@@ -333,9 +333,8 @@ describe('WebSocket', function () {
       const wss = new WebSocketServer({ port: ++port }, () => {
         const ws = new WebSocket(`ws://localhost:${port}`);
         ws.on('headers', (headers, res) => {
-          assert.equal(headers, res.headers);
-          wss.close();
-          done();
+          assert.strictEqual(headers, res.headers);
+          wss.close(done);
         });
       });
     });
--- a/test/WebSocket.test.js
+++ b/test/WebSocket.test.js
@@ -328,6 +328,17 @@ describe('WebSocket', function () {
 
       wss.on('connection', (client) => client.pong());
     });
+
+    it('emits a headers event', function (done) {
+      const wss = new WebSocketServer({ port: ++port }, () => {
+        const ws = new WebSocket(`ws://localhost:${port}`);
+        ws.on('headers', (headers, res) => {
+          assert.equal(headers, res.headers);
+          wss.close();
+          done();
+        });
+      });
+    });
   });
 
   describe('connection establishing', function () {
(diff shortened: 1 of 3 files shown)


test/WebSocket.test.js
```diff
@@ -328,6 +328,17 @@ describe('WebSocket', function () {
 
       wss.on('connection', (client) => client.pong());
     });
+
+    it('emits a headers event', function (done) {
+      const wss = new WebSocketServer({ port: ++port }, () => {
+        const ws = new WebSocket(`ws://localhost:${port}`);
+        ws.on('headers', (headers, res) => {
+          assert.equal(headers, res.headers);
```
lpinca: Please use `assert.strictEqual()`.

test/WebSocket.test.js
```diff
@@ -328,6 +328,17 @@ describe('WebSocket', function () {
...
+          wss.close();
```
lpinca: Can you please change this to `wss.close(done);`.

lpinca:
Using `rawHeaders` would be still inconsistent as header field name and header field value are two different elements. In the server sent `headers` event they are a single element.

I think I'm fine with this as is.

test/WebSocket.test.js
jwalton: Should this be true of the 'emits a ping event' and 'emits a pong event' tests above?  That's where I copied it from.  :P

lpinca: Yes, there are some tests that need to be updated. Not a biggie though.

jwalton: Alright.  I updated this one.  Should be good to go.  Thanks!

lpinca:
Thank you.

lpinca:
On second thought it was probably better to call this event `response` and only use the `http.IncomingMessage` as argument but it's too late now, the ship has sailed.

jwalton:
Yes, I like that better too, actually.  :P  Or, actually, I wonder if it would make sense to just add the `response` to the "open" event?  That's kind of a natural place for it.

lpinca:
That's goes against the spec. I should have waited a little more before cutting a new release.

jwalton:
Well, it's never too late to turn 'headers' into an undocumented event
until removing it in 3.x.

lpinca:
Yes, anyway I think 'headers' is also correct as the event is emitted when the upgrade headers are received. It gives you also access to whole `IncomingMessage` object but it's up to the user to not mess with it. So overall I'm still fine with it.

PR #1099, opened 2017-05-09
Remove the `upgradeReq` property

lpinca:
This removes the `upgradeReq` property from the `WebSocket` object. The `http.IncomingMessage` object is instead passed as the second argument to the `connection` event.

In this way most of the memory retained by the `http.IncomingMessage` object can be GC'ed reducing memory usage by ~20%.

Developers can save only the required info or the whole `http.IncomingMessage` object, if they have to, when the event is emitted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,10 +197,10 @@ app.use(function (req, res) {
 const server = http.createServer(app);
 const wss = new WebSocket.Server({ server });
 
-wss.on('connection', function connection(ws) {
-  const location = url.parse(ws.upgradeReq.url, true);
+wss.on('connection', function connection(ws, req) {
+  const location = url.parse(req.url, true);
   // You might use location.query.access_token to authenticate or share sessions
-  // or ws.upgradeReq.headers.cookie (see http://stackoverflow.com/a/16395220/151312)
+  // or req.headers.cookie (see http://stackoverflow.com/a/16395220/151312)
 
   ws.on('message', function incoming(message) {
     console.log('received: %s', message);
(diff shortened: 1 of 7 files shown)


3rd-Eden:
This is a major breaking change tho.

lpinca:
This is indeed a big change. It will break a lot of existing code, but along with https://github.com/nodejs/node/pull/11926 will greatly reduce memory usage.

In the worst case scenario all is needed to bring back the old behavior is this:

```js
  ws.upgradeReq = req;
});
```

Currently the opposite is needed to free the memory.

```js
  ws.upgradeReq = null;
});
```

Basically it will make the feature opt-in instead of opt-out.

lpinca:
Ops, I deleted the branch instead of force pushing 😄.

projetoarduino:
Hi

Where is sec-websocket-key

lpinca:
@projetoarduino in the `req` object.

```js
  const key = req.headers['sec-websocket-key'];
});
```

projetoarduino:
Thanks my friend.

projetoarduino:
Hi, this method work, but i need one unique id per client.

I need send  private messages.

Josephine7:
Hey @projetoarduino 
So I am following your train of thought. I also needed unique identifiers for our customers. How did you then send that message to a particular user ?

projetoarduino:
`  function senMess(get_dest, message){
      console.log(client);
      if (client !== ws && client.readyState === WebSocket.OPEN && client.ws._ultron.id == CLIENTS[get_dest]) {
        client.send(message);
      }
    });`

lpinca:
Don't use private properties as they may disappear at any time. If you need a unique id per connection generate it.

```js
  ws.id = uuid();
});
```

projetoarduino:
@lpinca Thanks, i tested this and work fine.

PR #1100, opened 2017-05-10
Remove unnecessary events

lpinca:
The canonical `connection` event is sufficient as the URL is exposed via `ws.upgradeReq.url`.


3rd-Eden:
+1

PR #1101, opened 2017-05-10
Remove the `flags` argument

lpinca:
The `flags` object specifies whether the received data is binary or not and whether it was masked or not. This object is actually redundant and therefore removed in this commit as the spec requires client-sent frames to be masked and server-sent frame to be unmasked.

The `typeof` operator can be used to check whether the received data is binary or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,9 +112,8 @@ ws.on('open', function open() {
   ws.send('something');
 });
 
-ws.on('message', function incoming(data, flags) {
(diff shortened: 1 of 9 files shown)


3rd-Eden:
+1 for the sake of API consistency.

rimmartin:
``
typeof message 
``
produces 'object'; can't tell if it holds binary or not

rimmartin:
@lpinca ah thanks

PR #1106, opened 2017-05-14
[test] Skip test if 127.0.0.2 unavailable

gibfahn:
This is the case for other platforms too.

I think this is the right way to do this, let me know if I got anything wrong!

cc/ @lpinca


lpinca:
Thanks!

PR #1112, opened 2017-05-16
Enhancements

### LINKED ISSUE #1097: Enhancements
Following are some possible enhacements that require a major version bump.

- [x] Remove the `upgradeReq` property from the `WebSocket` object and instead pass the `http.IncomingMessage` as the second argument to the `'connection'` event. This reduces memory usage by ~20%.
- [x] Remove the `flags` argument from the `'message'`, `'ping'`, and `'pong'` events. This object is not needed.
- [x] The permessage-deflate extension should be disabled by default due to its overhead on performance and memory comsumption.
- [x] ~~Use symbols for private properties?~~ Prefix all private properties.

lpinca:
This PR contains the following changes:

- Removes the `upgradeReq` property from the `WebSocket` object and instead pass the `http.IncomingMessage` as the second argument to the `'connection'` event. This reduces memory usage by ~20%.
- Removes the `flags` argument from the `'message'`, `'ping'`, and `'pong'` events. This object is not needed.
- Disables permessage-deflate by default on the server.
- Prefixes all private properties.

Closes #1097.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,9 +112,8 @@ ws.on('open', function open() {
   ws.send('something');
 });
 
-ws.on('message', function incoming(data, flags) {
-  // flags.binary will be set if a binary data is received.
-  // flags.masked will be set if the data was masked.
+ws.on('message', function incoming(data) {
+  console.log(data);
 });
 ```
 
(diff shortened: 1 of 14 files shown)


3rd-Eden:
Didn't I already +1 and merge these ._.?

3rd-Eden:
We might want to check that we're not breaking any "major" framework that integrates with us. (Engine.io and Primus)

lpinca:
> Didn't I already +1 and merge these ._.?

Yes but those PRs targeted this branch, not master :)

> We might want to check that we're not breaking any "major" framework that integrates with us. (Engine.io and Primus)

👍 

- In Primus the `upgradeReq` breaking change is handled here https://github.com/primus/primus/commit/fb5cf83ecdab3d8565d13522e14e0c9cd9386819.
- Engine.IO does not rely on `upgradeReq` property.

PR #1120, opened 2017-05-23
Add `req` argument to `completeUpgrade` callback

antoniobusrod:
When working with `noServer` and `handleUpgrade` from an extenal server, we must receive `req` parameter on `handleUpgrade` callback.

```
const http = require('http')
const WebSocket = require('ws')
const url = require('url')
const wss1 = new WebSocket.Server({ noServer: true })
const wss2 = new WebSocket.Server({ noServer: true })
const server = http.createServer()

server.on('upgrade', (request, socket, head) => {
  const pathname = url.parse(request.url).pathname

  if (pathname === '/foo') {
    wss1.handleUpgrade(request, socket, head, (ws, req) => {
      // req must exist
      wss1.emit('connection', ws, req)
    })
    wss2.handleUpgrade(request, socket, head, (ws, req) => {
      // req must exist
      wss2.emit('connection', ws, req)
    })
  } else {
    socket.destroy()
  }
})
const port = 3055
server.listen(port, () => {
  console.log(`Server listening on port ${port}`)
})
```

On last 3.x version (from https://github.com/websockets/ws/pull/885 discussion).

### DIFF
--- a/lib/WebSocketServer.js
+++ b/lib/WebSocketServer.js
@@ -267,7 +267,7 @@ class WebSocketServer extends EventEmitter {
     }
 
     socket.removeListener('error', socketError);
-    cb(client);
+    cb(client, req);
   }
 }
 
--- a/test/WebSocketServer.test.js
+++ b/test/WebSocketServer.test.js
@@ -321,7 +321,11 @@ describe('WebSocketServer', function () {
         const wss = new WebSocketServer({ noServer: true });
 
         server.on('upgrade', (req, socket, head) => {
-          wss.handleUpgrade(req, socket, head, (client) => client.send('hello'));
+          wss.handleUpgrade(req, socket, head, (client, req) => {
+            assert.strictEqual(req.method, 'GET');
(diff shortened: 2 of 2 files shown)


lpinca:
You can reuse the `request` argument of the `'upgrade'` listener.

```js
server.on('upgrade', (request, socket, head) => {
  wss.handleUpgrade(request, socket, head, (ws) => {
    wss.emit('connection', ws, request);
  });
});
```

antoniobusrod:
I thought that API should return same callback properties for `connection` event and `handleUpgrade` method callback.

I don't really care, but if you want to accept this feature, I'll should update https://github.com/websockets/ws/blob/master/doc/ws.md#serverhandleupgraderequest-socket-head-callback documentation too.

lpinca:
I honestly find this redundant. When using the `'connection'` event you don't have access to the upgrade request:

```js

  // There is no other way to inspect the request object.
});
```

When using `wss.handleUpgrade()` instead you already have the `request` object as it is the first argument to pass.

Issue #1144, opened 2017-06-18
wasClean property might be wrong


roymiloh:
When I try to use the Chrome's (same for Safari / Firefox) `WebSocket` instance directly and call `close` with any valid code (1000 or 3000 - 4999), I always get `wasClean === true`.
On the other hand, [we have this](https://github.com/websockets/ws/blob/fe44ff81cc414495f19416847ce658cbea079d06/lib/EventTarget.js#L58).
Yes, we are talking about browsers' implementation, but still, I believe it should be consistent (especially in my case where I use `ws` in my tests so I can simulate WebSocket client in nodejs).

[By W3C spec](https://www.w3.org/TR/2011/WD-websockets-20110419/#dfnReturnLink-1):

> The wasClean attribute represents whether the connection closed cleanly or not.

I'd suggest to add check for 3000 - 4999 range as well in order to determine `wasClean`'s value.

lpinca:
It makes sense. Please open a pull request.

lpinca:
Fixed by #1146.
Thanks.

PR #1146, opened 2017-06-19
Set wasClean to true when close code is between 3000 and 4999

roymiloh:
Due to #1144.


3rd-Eden:
<3! Thanks for taking the time to contribute <3

PR #1177, opened 2017-07-25
allow client to specify handshake request timeout

id0Sch:
Hi,
I have the need to control the connection timeout when initialising the connection,
I did my best to fall into your coding style and added the option in the documentation and added a unit test.
Hoping you will find this to your liking,

Let me know if there's anything I need to change, thanks!

### DIFF
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -471,6 +471,7 @@ function initAsServerClient (socket, head, options) {
  * @param {Object} options Connection options
  * @param {String} options.protocol Value of the `Sec-WebSocket-Protocol` header
  * @param {(Boolean|Object)} options.perMessageDeflate Enable/disable permessage-deflate
+ * @param {Number} options.handshakeTimeout Timeout in milliseconds for the handshake request
  * @param {String} options.localAddress Local interface to bind for network connections
  * @param {Number} options.protocolVersion Value of the `Sec-WebSocket-Version` header
  * @param {Object} options.headers An object containing request headers
@@ -493,6 +494,7 @@ function initAsClient (address, protocols, options) {
     protocolVersion: protocolVersions[1],
     protocol: protocols.join(','),
     perMessageDeflate: true,
+    handshakeTimeout: null,
     localAddress: null,
     headers: null,
     family: null,
@@ -631,6 +633,14 @@ function initAsClient (address, protocols, options) {
   if (agent) requestOptions.agent = agent;
 
   this._req = httpObj.get(requestOptions);
+
+  if (options.handshakeTimeout) {
+    this._req.setTimeout(options.handshakeTimeout, () => {
+      this._req.abort();
+      this.emit('error', new Error('timeout'));
+      this.finalize(true);
+    });
+  }
 
   this._req.on('error', (error) => {
     if (this._req.aborted) return;
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -636,10 +636,8 @@ function initAsClient (address, protocols, options) {
 
   if (options.handshakeTimeout) {
     this._req.setTimeout(options.handshakeTimeout, () => {
-      const err = new Error('opening handshake has timed out');
-      err.code = 'TIMEOUT';
       this._req.abort();
-      this.emit('error', err);
+      this.emit('error', new Error('opening handshake has timed out'));
       this.finalize(true);
     });
   }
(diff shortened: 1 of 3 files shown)


test/WebSocket.test.js
```diff
@@ -461,6 +461,30 @@ describe('WebSocket', function () {
         req.abort();
       });
     });
+
+    it('error is emitted if server response exceeds timeout', function (done) {
```
lpinca: I'd use "emits an error if the opening handshake timeout expires".

test/WebSocket.test.js
```diff
@@ -461,6 +461,30 @@ describe('WebSocket', function () {
...
+      const timeout = 100;
+      server.once('upgrade', (req, socket) => {
```
lpinca: I think it's fine to just make the listener a noop.

test/WebSocket.test.js
```diff
@@ -461,6 +461,30 @@ describe('WebSocket', function () {
...
+            `HTTP/1.1 401 ${http.STATUS_CODES[401]}\r\n` +
+            'Connection: close\r\n' +
+            'Content-type: text/html\r\n' +
+            `Content-Length: ${http.STATUS_CODES[401].length}\r\n` +
+            '\r\n'
+          );
+        }, timeout * 1.5);
+      });
+
+      const ws = new WebSocket(`ws://localhost:${port}`, null, { handshakeTimeout: timeout });
```
lpinca: Can you please reduce the length of this line?

```js
const ws = new WebSocket(`ws://localhost:${port}`, null, {
  handshakeTimeout: timeout
});
```

lib/WebSocket.js
```diff
@@ -632,6 +634,14 @@ function initAsClient (address, protocols, options) {
 
   this._req = httpObj.get(requestOptions);
 
+  if (options.handshakeTimeout) {
+    this._req.setTimeout(options.handshakeTimeout, () => {
```
lpinca: This may not work exactly as you expect but it's fine in most cases. All `request.setTimeout()` does is calling `socket.setTimeout()` on the request socket.

This means that the timeout expires only when there are `x` milliseconds of inactivity. It's possible that the response arrives after the specified timeout with no 'timeout' event emitted.

test/WebSocket.test.js
id0Sch: I tried that, but for some reason it doesn't work.

id0Sch:
@lpinca thanks for the quick reply , I indeed suck at picking out message strings and such.

lib/WebSocket.js
```diff
@@ -632,6 +634,16 @@ function initAsClient (address, protocols, options) {
...
+      const err = new Error('opening handshake has timed out');
+      err.code = 'TIMEOUT';
```
lpinca: Do you need `code`? It's useful and many Node.js errors have this property but currently we don't have this property on our errors. I prefer to not add it for consistency and maybe revisit later.

test/WebSocket.test.js
lpinca: Does it fail on the `afterEach()` hook?

lib/WebSocket.js
id0Sch: No problem, thought since the message is a sentence adding code could be a better way to "locate" the error, but all good.

test/WebSocket.test.js
id0Sch: the test case's timeout exceeds.
the "error" event is never called.

lpinca: Ok, this is becuase `allowHalfOpen` is set to true and the server socket is not automatically closed.

This should fix the issue.

```js
server.once('upgrade', (req, socket) => socket.on('end', socket.end));
```

test/WebSocket.test.js
```diff
@@ -461,6 +461,22 @@ describe('WebSocket', function () {
         req.abort();
       });
     });
+
+    it('emits an error if the opening handshake timeout expires', function (done) {
+      const timeout = 100;
+      server.once('upgrade', (req, socket) => socket.on('end', socket.end));
+
+      const ws = new WebSocket(`ws://localhost:${port}`, null, {
+        handshakeTimeout: timeout
```
lpinca: One last nit: would you mind to inline the value and remove the variable?

id0Sch: now that it's not used, no problem.

lpinca:
@websockets/admin LGTY?

lpinca:
Thank you.

id0Sch:
@lpinca thanks for being so helpful and fast moving! 
i'ts been a pleasure contributing to this awesome project! 
keep up the good work!

Issue #1227, opened 2017-11-01
Can't connect from ws client to ws server: TLS handshake failure


Hativ3:
- [X] I've searched for any related issues and avoided creating a duplicate issue.

#### Description

If I try to connect with the attached ws client to the attached ws server I get the following error:

```
events.js:193
      throw er; // Unhandled 'error' event
      ^

Error: write EPROTO 140695292278592:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:../deps/openssl/openssl/ssl/s23_clnt.c:772:

    at _errnoException (util.js:1031:13)
    at WriteWrap.afterWrite [as oncomplete] (net.js:873:14)
```

But Chrome and Firefox can connect to the ws server, so the problem must be in the ws client.

#### Reproducible in:

version: 3.2.0
Node.js version(s): 9.0.0
OS version(s): elementary OS 0.4.1 Loki, based on Ubuntu 16.04.3 LTS

#### Steps to reproduce:

1. Run attached websocket server
2. Run attached websocket client

### Expected result:

Connecting to websocket without error.

### Actual result:

Not connecting to websocket, producing an error.

### Attachments:
Debug output:
```
hativ@pc:~# NODE_DEBUG=net,tls,crypto node index.js 
NET 1149: pipe false null
NET 1149: connect: find host websocket.example.com
NET 1149: connect: dns options { family: undefined, hints: 32 }
NET 1149: _read
NET 1149: _read wait for connection
NET 1149: afterConnect
TLS 1149: start
NET 1149: _read
NET 1149: Socket._read readStart
NET 1149: afterWrite -71
NET 1149: write failure { Error: write EPROTO 140067809924928:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:../deps/openssl/openssl/ssl/s23_clnt.c:772:

    at _errnoException (util.js:1031:13)
    at WriteWrap.afterWrite [as oncomplete] (net.js:873:14) errno: 'EPROTO', code: 'EPROTO', syscall: 'write' }
NET 1149: destroy
NET 1149: close
NET 1149: close handle
events.js:193
      throw er; // Unhandled 'error' event
      ^

Error: write EPROTO 140067809924928:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:../deps/openssl/openssl/ssl/s23_clnt.c:772:

[...]

PR #1228, opened 2017-11-01
Add ecdhCurve option

Hativ3:
Support ecdhCurve option as discussed in https://github.com/websockets/ws/issues/1227


lpinca:
Thanks!

PR #1233, opened 2017-11-10
Use the `broadcast` function in the example.

jayantbh:
It was not being used earlier, this change shows the broadcast function being actually used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,11 +161,7 @@ wss.broadcast = function broadcast(data) {
 wss.on('connection', function connection(ws) {
   ws.on('message', function incoming(data) {
     // Broadcast to everyone else.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,11 +161,7 @@ wss.broadcast = function broadcast(data) {
 wss.on('connection', function connection(ws) {
   ws.on('message', function incoming(data) {
     // Broadcast to everyone else.
-    wss.clients.forEach(function each(client) {
-      if (client !== ws && client.readyState === WebSocket.OPEN) {
-        client.send(data);
-      }
-    });
+    wss.broadcast(data);
```
lpinca: This is not the same as it broadcasts to **all** clients without excluding the one that received the message.

PR #1245, opened 2017-11-30
mocha 4 require explicit exit

dcharbonnier:
https://mochajs.org/#--exit----no-exit
> Prior to version v4.0.0, by default, Mocha would force its own process to exit once it was finished executing all tests. This behavior enables a set of potential problems; it’s indicative of tests (or fixtures, harnesses, code under test, etc.) which don’t clean up after themselves properly. Ultimately, “dirty” tests can (but not always) lead to false positive or false negative results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "lib"
   ],
   "scripts": {
-    "test": "eslint . && nyc --reporter=html --reporter=text mocha test/*.test.js",
+    "test": "eslint . && nyc --reporter=html --reporter=text mocha --exit test/*.test.js",
     "integration": "eslint . && mocha test/*.integration.js",
     "lint": "eslint ."
   },


lpinca:
Thanks but I prefer to not force exit. If the process doesn't exit there is something wrong.

lpinca:
I didn't merge https://github.com/primus/primus/pull/611 on Primus for the same reason. There is actually not much we can do there due to how polling works.

3rd-Eden:
I agree, found issues in some of my other projects where I upgraded to latest mocha were timers were never cleaned up correctly. If there is nothing more alive in the process, it should just exit automatically. If it doesn't do that we're leaking somewhere.

dcharbonnier:
right there is a bug related to the dns that create this problem

dcharbonnier:
the problem was https://github.com/websockets/ws/pull/1246

PR #1250, opened 2017-12-09
Fix rendering of history in md file

GeorgeTaveras1231:
*Feel free to reject if these weren't your intentions* 

- Uses an unordered list for the vulnerability list to render each item on a new line.
- Fixes typo: vulnerablity -> vulnerability


lpinca:
Thank you.

PR #1252, opened 2017-12-15
[doc] Remove duplicate 'is' in doc.

H1Gdev:
Remove duplicate 'is' in doc.

before:
If it **is is** provided with a single argument then that is:

after:
If it **is** provided with a single argument then that is:


lpinca:
Thank you.

PR #1261, opened 2017-12-23
Breaking changes

lpinca:
- The close status code is now set to 1005 if the received close frame contains
  no status code.
- Error messages and types have been updated.
- The ErrorCodes module has been removed.
- Modules files have been renamed.
- The `onerror` event handler now receives an `ErrorEvent` instead of JavaScript
  error.
- The third argument of `WebSocket.prototype.ping()` and
  `WebSocket.prototype.pong()` is no longer a boolean but an optional callback.
- The non-standard `protocolVersion` and `bytesReceived` attributes have been
  removed.
- The `extensions` attribute is no longer an object but a string representing
  the extensions selected by the server.
- The `'headers'` event on the client has been renamed to `'upgrade'`. Listeners
  of this event now receive only the `response` argument.
- The `WebSocket.prototype.pause()` and `WebSocket.prototype.resume()` methods
  have been removed to prevent the user from interfering with the state of the
  underlying `net.Socket` stream.


lpinca:
@shellscape this is a good time to improve error messages/types if you have suggestions.

shellscape:
@lpinca thanks for the ping. I'm out of town until Jan 3rd. Not sure what your schedule is for merging and publishing but it'll be a bit before I can sit down to take a look at this.

lpinca:
I can definitely wait until new year before publishing version 4.0

shellscape:
> The 'headers' event on the client has been renamed to 'upgrade'. Listeners
of this event now receive only the response argument.

It'd be helpful to list the justification (aka reasons) for some of the changes. On the surface it's difficult to tell why that particular change was made.

lpinca:
@shellscape yes, there is a reference to the PR (#1082) where the `'headers'` event was introduced in the commit message. The reason for the change is in PR comments.

lpinca:
@websockets/admin should we remove `WebSocket.prototype.pause()` and `WebSocket.prototype.resume()`?

Read backpressure is currently not handled so if the receiver peer is slower than the sender and permessage-deflate is enabled, memory usage may grow indefinitely on the receiver.

We could handle read backpressure by pausing the `net.Socket` when we start inflating data and resuming it when all buffered data has been consumed, but for this to work, the internal socket must not be paused/resumed by third parties.

lpinca:
Anyone, please take a look at https://github.com/websockets/ws/pull/1261/commits/4e8bd12c1232af3aecf68e4df38c55b51508ce3d and see if error messages/types are ok. Suggestions are welcome.

shellscape:
@lpinca sorry it took me a few more days than anticipated, but I left my comments on there. Huge improvement and I'm sure it'll be much appreciated. I dropped some comments that I think would further add value as well.

lpinca:
@TimothyGu, @shellscape PTAL 087d0e8.

lpinca:
@shellscape @TimothyGu thank you for reviewing!

shellscape:
@lpinca my pleasure. thank you for giving me the opportunity!

PR #1278, opened 2018-01-15
[minor] Move 'use strict' declarations on top of the file.

forty:
Some JS parsers seem to ignore it otherwise.

### DIFF
--- a/bench/parser.benchmark.js
+++ b/bench/parser.benchmark.js
@@ -1,9 +1,3 @@
-/*!
- * ws: a node.js websocket client
- * Copyright(c) 2011 Einar Otto Stangvik <einaros@gmail.com>
- * MIT Licensed
- */
-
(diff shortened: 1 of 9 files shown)


lpinca:
Why is this needed? I mean, is it make it work when transpiling? Otherwise as far as I know it works with V8 parser.

forty:
Yes, it seems to work properly with node, but it does not work at least with jsdoc[1] parser it seems. I tried to find out in the spec who was right, but parsing the ECMA spec is not easy...

[1] http://usejsdoc.org/

lpinca:
Got it, how about removing the license comments completely? The LICENSE file should be sufficient.

forty:
Sure! I'll update the PR

lpinca:
Thank you.

PR #1290, opened 2018-01-27
Supports dynamic  pattern in "path" option for WebSocket.Server

Ericbla:
Hello, 

Here is a minor Pull-Request to add support for "dynamic" pattern in the `path` option of the *WebSocket.Server* constructor.

### Motivation:
When using this websocket server with a user provided HTTP server, it is often useful to control where the websocket server is hooked up in the URL path tree. This is possible with the `path` option, but unfortunately only a fixed path prefix is allowed.

However it is sometime convenient to allow this path prefix to be variable, for example to let an upstream reverse proxy to use some rules based on this variable part.

### Proposal:
Simply allow to support some "REST" like URL patterns in the `path` option and use a pattern matching rule instead of a strict comparison when comparing `path`and incoming URL pathname.

The proposed code supports the following dynamic patterns: `{keyword}`, `:keyword` and `*`. All these
patterns will match any uri character up to the next `/` path delimiter or the end of the uri pathname.

E.g.: Things like:
  - `/foo/{foo-id}/bar/{bar-id}`   # Like OPENAPI of Swagger notation
  - `/foo/:foo-id/bar/:bar-id`      # Like Expressjs, ... notation
  - `/foo/*/bar/*`
are all equivalent and will match any url path like: `/foo/123/bar/456`

I have intentionaly not supported a full RegExp notation at it will be more complicated to secure it.

### Pull-Request:
- `lib/websocket-server.js` : Implementation
- `test/websocker-server.test.js` : Unit tests
- `doc/ws.md` :  Some explanations for the `path`option new format

--
Eric

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -16,6 +16,11 @@ This class represents a WebSocket server. It extends the `EventEmitter`.
   - `handleProtocols` {Function} A function which can be used to handle the
     WebSocket subprotocols. See description below.
   - `path` {String} Accept only connections matching this path.
+    Note: `path` supports dynamic path elements like:
+    `{keyword}`, `:keyword` or `*` that matches any uri characters up to next
+    '/' path delimiter.
+    E.g.: a path like: `/foo/{foo-id}/bar/:bar-id/any/*` would match url pathes
+    like: `/foo/123/bar/456/any/789`.
   - `noServer` {Boolean} Enable no server mode.
   - `clientTracking` {Boolean} Specifies whether or not to track clients.
   - `perMessageDeflate` {Boolean|Object} Enable/disable permessage-deflate.
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -59,6 +59,16 @@ class WebSocketServer extends EventEmitter {
       );
     }
 
+    if (options.path) {
+      // transform the path to a regexp that can match the represented pattern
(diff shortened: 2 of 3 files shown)


lpinca:
Can't you simply override [`WebSocketServer.prototype.shouldHandle()`](https://github.com/websockets/ws/blob/master/doc/ws.md#servershouldhandlerequest)?

Ericbla:
Do you mean don't use the path option at all ?
Maybe yes, but in this case I would accept an Upgrade from any URL. In my case it is not a real big issue, but  I will have to check the url then after accepting the websocket.

lpinca:
No I mean something like this:

```js

WebSocket.Server.prototype.shouldHandle = function (req) {
  // Add your custom path handling here, and return `true` or `false` accordingly.
};
```

This is why `shouldHandle` exists.

Ericbla:
Ah! Sorry I didn't noticed that this method was in the API doc.

So yes you are right it is a better approach.

Forget my PR.

Tanks.

lpinca:
Thanks anyway.

lpinca:
The above example actually overwrites `shouldHandle`. To override it add the custom version to the server instance.

```js
const wss = new WebSocket.Server(options);

wss.shouldHandle = function (request) {
  // ...
};
```

PR #1294, opened 2018-02-04
Provide http server port directly from _server

tspn:
Provide http server port directly from _server

Do I need to write test or anything else to complete this PR ?
Or better way to get port if I bind with 0

Thanks

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -91,6 +91,24 @@ class WebSocketServer extends EventEmitter {
   }
 
   /**
+   * Returns the bound address, the address family name, and port of the server
+   * as reported by the operating system if listening on an IP socket.
+   * If the server is listening on a pipe or UNIX domain socket, the name is
+   * returned as a string.
+   *
+   * @return {(Object|String|null)} The address of the server
+   * @public
+   */
+  address () {
+    if (this.options.noServer) {
+      throw new Error('The server is operating in "noServer" mode');
+    }
+
+    if (!this._server) return null;
+    return this._server.address();
+  }
+
+  /**
    * Close the server.
    *
    * @param {Function} cb Callback
@@ -274,23 +292,6 @@ class WebSocketServer extends EventEmitter {
     }
 
     cb(ws);
-  }
-
-  /**
-   * Returns the bound address, the address family name, and port of the server
-   * as reported by the operating system if listening on an IP socket.
-   * If the server is listening on a pipe or UNIX domain socket, the name is
-   * returned as a string.
-   *
-   * @return {(Object|String|null)} The address of the server
-   * @public
-   */
-  address () {
-    if (this.options.noServer) {
-      throw new Error('This server is operating in "noServer" mode');
-    }
-    if (!this._server) return null;
-    return this._server.address();
   }
 }
 
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -278,11 +278,14 @@ class WebSocketServer extends EventEmitter {
 
   /**
    * Provide address of server
+   * @return {Object} Returns the bound address, the address family name, and port of the server as reported by the operating system if listening on an IP socket
    * @public
    */
   address () {
-    if (!this._server) throw new Error(`Server was not initialized`);
-    console.log(this._server.address());
+    if (this.options.noServer) {
+      throw new Error('This server is operating in "noServer" mode');
+    }
+    if (!this._server) return null;
     return this._server.address();
   }
 }
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -281,10 +281,8 @@ class WebSocketServer extends EventEmitter {
    * @public
    */
   address () {
-    if (!this._server) {
-      throw new Error(`Server was not initialized`);
-    }
-
+    if (!this._server) throw new Error(`Server was not initialized`);
+    console.log(this._server.address());
     return this._server.address();
   }
 }
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -277,11 +277,15 @@ class WebSocketServer extends EventEmitter {
   }
 
   /**
-   * Upgrade the connection to WebSocket.
+   * Provide address of server
    * @public
    */
-  port () {
-    return this._server.address().port;
+  address () {
+    if (!this._server) {
+      throw new Error(`Server was not initialized`);
+    }
+
+    return this._server.address();
   }
 }
 
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -277,8 +277,12 @@ class WebSocketServer extends EventEmitter {
   }
 
   /**
-   * Provide address of server
-   * @return {Object} Returns the bound address, the address family name, and port of the server as reported by the operating system if listening on an IP socket
+   * Returns the bound address, the address family name, and port of the server
+   * as reported by the operating system if listening on an IP socket.
+   * If the server is listening on a pipe or UNIX domain socket, the name is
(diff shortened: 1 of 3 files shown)


lib/websocket-server.js
```diff
@@ -275,6 +275,14 @@ class WebSocketServer extends EventEmitter {
 
     cb(ws);
   }
+
+  /**
+   * Upgrade the connection to WebSocket.
+   * @public
+   */
+  port () {
```
lpinca: A couple of thoughts:

1. It's probably better to call this `address` and return `this._server.address()`. This will handle the case where the HTTP server is closed.
2. `WebSocketServer` could be initialized in `noServer` mode. In this case `this._server` is `null`. What should we return? Returning `null` doesn't make sense as it's possible that the external server is actually bound and listening for connections. Throwing an error is probably the best option.

tspn: understood. Thank for the guide

lib/websocket-server.js
```diff
@@ -275,6 +275,16 @@ class WebSocketServer extends EventEmitter {
 
     cb(ws);
   }
+
+  /**
+   * Provide address of server
```
lpinca: We need a better description and specify what is returned via `@return ...`. Copying https://nodejs.org/dist/latest/docs/api/net.html#net_server_address is ok.

tspn:
@lpinca Can we Merge ? :)

lib/websocket-server.js
```diff
@@ -275,6 +275,16 @@ class WebSocketServer extends EventEmitter {
...
+   * @public
+   */
+  address () {
+    if (!this._server) throw new Error(`Server was not initialized`);
```
lpinca: Use `this.options.noServer` as we don't wan't to throw the error if `wss.address()` is called after `wss.close()`

```js
if (this.options.noServer) {
  throw new Error('This server is operating in "noServer" mode');
}
```

lib/websocket-server.js
```diff
@@ -275,6 +275,16 @@ class WebSocketServer extends EventEmitter {
...
+    console.log(this._server.address());
```
lpinca: Replace this line with

```js
if (!this._server) return null;
```

This will handle the case where `address()` is called after `close()`.

test/websocket-server.test.js
```diff
@@ -724,4 +724,24 @@ describe('WebSocketServer', function () {
       });
     });
   });
+
+  describe('Server address', function () {
```
lpinca: Please move this block after `describe('#ctor', ...);` and use `'#address'` as title

lib/websocket-server.js
tspn: done sir :)

tspn: Noooo CI fail

lpinca: Please put the description on this line and wrap lines at 80 chars.

```
/**
 * Returns the bound address, the address family name, and port of the server
 * as reported by the operating system if listening on an IP socket.
 * If the server is listening on a pipe or UNIX domain socket, the name is
 * returned as a string.
 *
 * @return {(Object|String|null)} The address of the server
 * @public
 */
```

test/websocket-server.test.js
```diff
@@ -137,6 +137,25 @@ describe('WebSocketServer', function () {
     });
   });
 
+  describe('#address', function () {
+    it('can provide server address', function (done) {
```
lpinca: `'returns the address of the server'`

test/websocket-server.test.js
```diff
@@ -137,6 +137,25 @@ describe('WebSocketServer', function () {
...
+      const wss = new WebSocket.Server({ port: 0 }, () => {
+        const addr = wss.address();
+        assert.deepEqual(wss._server.address(), addr);
```
lpinca: `deepStrictEqual`

test/websocket-server.test.js
```diff
@@ -137,6 +137,25 @@ describe('WebSocketServer', function () {
...
+        wss.close(done);
+      });
+    });
+
+    it(`will thrown exception if server was closed`, function (done) {
```
lpinca: `'throws an error when operating in "noServer" mode'`

test/websocket-server.test.js
```diff
@@ -137,6 +137,25 @@ describe('WebSocketServer', function () {
...
+      const wss = new WebSocket.Server({ port: 0, noServer: true }, () => {
```
lpinca: If you add the `port` option it will create a server. We don't want that.

```js
it('throws an error when operating in "noServer" mode', function () {

  assert.throws(
    () => wss.address(),
    /^Error: This server is operating in "noServer" mode$/
  );
});
```

test/websocket-server.test.js
```diff
@@ -137,6 +137,25 @@ describe('WebSocketServer', function () {
...
+        assert.throws(() => {
+          wss.address();
+        }, /This server is operating in "noServer" mode/);
+        wss.close(done);
+      });
+    });
```
lpinca: Add a new test for the `address()` after `close()` case where `null` is returned.

lpinca:
We also need to add documentation for this method in https://github.com/websockets/ws/blob/master/doc/ws.md

test/websocket-server.test.js
tspn: okidokie

test/websocket-server.test.js
```diff
@@ -137,6 +137,32 @@ describe('WebSocketServer', function () {
     });
   });
 
+  describe('#address', function () {
+    it('returns the address of the server', function (done) {
+      const wss = new WebSocket.Server({ port: 0 }, () => {
+        const addr = wss.address();
+        assert.deepStrictEqual(wss._server.address(), addr);
```
lpinca: Sorry for being picky, we are getting closer. Can you please swap the arguments? `actual` first `expected` second.

test/websocket-server.test.js
```diff
@@ -137,6 +137,32 @@ describe('WebSocketServer', function () {
...
+        wss.close(done);
+      });
+    });
+
+    it('throws an error when operating in "noServer" mode', function () {
+      const wss = new WebSocket.Server({ noServer: true }, () => {
+        assert.throws(() => {
+          wss.address();
+        }, /This server is operating in "noServer" mode/);
```
lpinca: Please use `/^Error: This server is operating in "noServer" mode$/` so we also validate the type of the error.

test/websocket-server.test.js
```diff
@@ -137,6 +137,32 @@ describe('WebSocketServer', function () {
...
+      });
+    });
+
+    it('returns null when called after server closed', function (done) {
+      const wss = new WebSocket.Server({ port: 0 }, () => {
+        wss.close();
+        assert.deepStrictEqual(null, wss.address());
```
lpinca: Please swap the arguments and use `strictEqual` not deep.

test/websocket-server.test.js
lpinca: We can use `strictEqual` here as well.

tspn:
It seems like code coverage doesn't see the cover of some test
```js
    it('throws an error when operating in "noServer" mode', function () {
      const wss = new WebSocket.Server({ noServer: true }, () => {
        assert.throws(() => {
          wss.address();
        }, /^Error: This server is operating in "noServer" mode$/);
      });
    });
```

lpinca:
Thank you!

tspn:
Hey, Thank you so much to complete this PR (you rewrite it all haha) 
You’re awesome !!

PR #1298, opened 2018-02-07
feat(README): add a link to isomorphic-ws

julien-f:
`isomorphic-ws` is a tiny wrapper around `ws` (peer dependency) which returns `global.WebSocket` on the browser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Passes the quite extensive Autobahn test suite: [server][server-report],
 **Note**: This module does not work in the browser. The client in the docs is a
 reference to a back end with the role of a client in the WebSocket
 communication. Browser clients must use the native
-[`WebSocket`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) object.
+[`WebSocket`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) object. If you're looking for an isomorphic wrapper, take a look at [isomorphic-ws](https://github.com/heineiuo/isomorphic-ws).
 
 ## Table of Contents
 
--- a/README.md
+++ b/README.md
@@ -14,8 +14,9 @@ Passes the quite extensive Autobahn test suite: [server][server-report],
 **Note**: This module does not work in the browser. The client in the docs is a
 reference to a back end with the role of a client in the WebSocket
 communication. Browser clients must use the native
-[`WebSocket`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket)
-object. If you're looking for an isomorphic wrapper, take a look at
+[`WebSocket`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) object.
+To make the same code work seamlessy on Node.js and the browser, you can use
+one of the many wrappers available on npm, like
 [isomorphic-ws](https://github.com/heineiuo/isomorphic-ws).
 
 ## Table of Contents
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ Passes the quite extensive Autobahn test suite: [server][server-report],
 **Note**: This module does not work in the browser. The client in the docs is a
 reference to a back end with the role of a client in the WebSocket
 communication. Browser clients must use the native
-[`WebSocket`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) object.
+[`WebSocket`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) object. If you're looking for an isomorphic wrapper, take a look at [isomorphic-ws](https://github.com/heineiuo/isomorphic-ws).
```
lpinca: Can you please keep the limit of 80 chars per line?

julien-f: Sure thing.

lpinca:
There are a lot module like `isomorphic-ws`, what makes it special? I think that https://github.com/heineiuo/isomorphic-ws/blob/605be49fea91a3785d9bae8d976f7f20e684b0e2/browser.js#L1 is a bit limiting. Why is it better than say, https://github.com/maxogden/websocket-stream/blob/48dc3ddf943e5ada668c31ccd94e9186f02fafbd/ws-fallback.js which also supports old browsers?

julien-f:
What's better with `isomorphic-ws` is:
- delimited perimeter: only a small wrapper around `ws`
- `ws` is a peer dependency: the user can choose which version to use

Though I agree the `websocket-stream` fallback is better, I will create a PR to improve `isomorphic-ws`.

At the end, I don't care which module to use, but I think there should be an obvious one that as many people as possible gather around 😃

3rd-Eden:
I really don't want to start endorse a single "blessed" module here. There are so many edge cases that you need to consider in the browser file nearly all that are available right now have their own flaws. 

E.g. assuming that `global` exists or that `this` is the absolute global. The fact difference in constructor arguments, we require 3 while WebSocket supports 2. So you end up writing your own Constructor function. Which then needs to inherit from WebSocket. And then you release that in addition to inheriting the prototype you also need to inherit the static properties of WebSocket. The list of potential incompatibilities here is endless.

julien-f:
The module should probably warn that there are limitations but IMHO it's still better than forcing everyone to come up with their own solution.

lpinca:
> IMHO it's still better than forcing everyone to come up with their own solution.

I agree.

lpinca:
@3rd-Eden so are you 👎 on this?

3rd-Eden:
We should probably advertise multiple options, instead of single one.

3rd-Eden:
But I'm indifferent about merging this, so not a +1 but also no -1.

lpinca:
I also have no strong opinions, I only find a bit weird to say "use this module" when there isn't a particular reason to actually use _that_ module.

Anyway from what I've gathered the problem here is not suggesting a particular module. I think @julien-f wants to redirect users that need to shim ws towards a common shared solution.

The shim was removed from this repo for the reasons @3rd-Eden mentioned above so re-adding it here doesn't seem a good idea.

julien-f:
Yep, I understand you do not want the burden of maintaining this, but I think it would makes sense to help the users who look for something like this.

lpinca:
How about using a different wording, like

"To make the same code work seamlessly on Node.js and the Browser, you can use one of the many wrappers available on npm, like isomorphic-ws".

This should make everyone happy. It still endorses isomorphic-ws but to a lesser extent.

lpinca:
@julien-f would you mind updating?

julien-f:
Done 😃

lpinca:
Thank you.

PR #1300, opened 2018-02-10
feat(websocket): 'emit error' if send fail and error listener is registered

JacopoDaeli:
Currently if the `send` method fails, and no `callback` is passed as argument, it throws.

However, it would be nice that if an `error` listener is registered, the error is "emitted" instead of thrown.

This PR is to implement this feature.

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -371,6 +371,7 @@ class WebSocket extends EventEmitter {
       );
 
       if (cb) return cb(err);
+      if (this.listenerCount('error') > 0) return this.emit('error', err);
       throw err;
     }
 


lpinca:
You have two options:

1. Check the `readyState` before calling `WebSocket#{send,ping,pong}()`.
```js
if (ws.readyState === WebSocket.OPEN) {
  // ...
}
```
2. Always add a callback.

Emitting an error doesn't make much sense in this case imho.

JacopoDaeli:
@lpinca why doesn't make sense? If we register an `error` listener, don't we want the `error` being emitted?

lpinca:
Unlike other errors, this is predictable and is a programmer error, so I prefer to throw it.

JacopoDaeli:
Personally this looks to me more like an application error. It always happen to me after the connection is established. For some reasons, the socket become "closed", but the emitter doesn't fired "close" event, and once I try to "write" something on the write, it throws. What do you think about it?

lpinca:
Every `WebSocket` goes in `CLOSING` state before switching to `CLOSED` and emitting the `'close'` event and in `CLOSING` state the `WebSocket` is no longer writable.

The rule of thumb is to always check the `readyState` before writing.

JacopoDaeli:
👍

Issue #1256, opened 2017-12-17
read ECONNRESET error on disconnect (ws 3.3.3)


G-Rath:
- [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description

Upon updating to ws 3.3.3, when a client disconnects the following exception is thrown:

```
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at _errnoException (util.js:1024:11)
    at TCP.onread (net.js:615:25)
```

Code works fine in previous version.

#### Reproducible in:

version: 3.3.3
Node.js version(s): v8.9.1
OS version(s): Windows 10 Pro, Fall update

#### Steps to reproduce:

1. Create a websocket server
2. Connect to the websocket server via Chrome (`new WebSocket('ws://localhost:3000'))
3. Disconnect from the websocket server (via refreshing the page in Chrome)

### Expected result:

The disconnect is handled gracefully, with the 'close' event being called.

### Actual result:

An exception is thrown, which crashes the app:

```
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at _errnoException (util.js:1024:11)
    at TCP.onread (net.js:615:25)
```

G-Rath:
Issue is caused by lack of error listener on the socket, can be fixed like so:

```
ws.on('error', () => console.log('errored'));
```

This isn't documented in the README.md, and seems to be a pretty big breaking change.

lpinca:
You need to always a listener for the `'error'` event as there are other errors that can be emitted.
`net.Socket` errors are no longer swallowed.

wangjian1119:
I heard it elsewhere in the latest Google browsers

wangjian1119:
his is the case for the latest Google browsers

polkovnikov-ph:
@lpinca So can we add it to README? If it's a required thing to do, why is it absent there?

lpinca:
https://github.com/websockets/ws/blob/180b388a2ed83fddd667d88970a943d0f47d32ec/doc/ws.md#event-error-1

G-Rath:
@lpinca My point is more about the fact that this is happening on client disconnect, which I expect to be a perfectly normally event in a socket servers lifecycle. 

To me, because you have a 'close' event, I'd assume that exceptions caused as a result of a client disconnecting would be handled by this library, since it's something that the library seems to be able to detect and act on.

[...]

PR #1302, opened 2018-02-15
Make `Receiver` inherit from `stream.Writable`

lpinca:
This makes `Receiver` inherit from `stream.Writable`.

The main reason for this change is to have back-pressure properly handled when permessage-deflate is enabled.

There are no significant performance differences but memory usage is a lot higher. 100k connections use ~400 MB on master and ~600 MB with this patch.

I'm opening this PR to discuss whether or not this makes sense.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,6 @@ node_js:
   - "8"
   - "6"
   - "4"
-  - "4.2.0"
+  - "4.2.2"
 after_success:
   - "npm install coveralls@3 && nyc report --reporter=text-lcov | coveralls"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ environment:
     - nodejs_version: "8"
     - nodejs_version: "6"
     - nodejs_version: "4"
-    - nodejs_version: "4.2.0"
+    - nodejs_version: "4.2.2"
 platform:
   - x86
   - x64
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -244,8 +244,7 @@ human-readable string explaining why the connection has been closed.
 
 - `error` {Error}
 
-Emitted when an error occurs. Errors from the underlying `net.Socket` are
-forwarded here.
+Emitted when an error occurs.
 
 ### Event: 'message'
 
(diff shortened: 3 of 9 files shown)


lpinca:
Managed to reduce memory consumption by not using `Readable.prototype.pipe()`. Now 100k connections use ~450 MB which is ~10% more memory than master.

I think this is a good trade off.

lpinca:
If there is anyone watching this, can you please share your opinion about `net.Socket` errors? Should we re-emit them or not?

- Node.js core is not consistent about these errors on the HTTP stack. On the server they are not re-emitted, or better they are re-emitted using the [`'clientError'`](https://nodejs.org/dist/latest/docs/api/http.html#http_event_clienterror) event on the server but you can omit the `'error'` listener. On the client they are re-emitted on the [`ClientRequest`](https://nodejs.org/dist/latest/docs/api/http.html#http_class_http_clientrequest) instance and you can't omit the `'error'` listener.
- From https://github.com/websockets/ws/issues/1256 it seems that these errors end up being ignored as there is not much that can be done to prevent them. If the common action is to filter them out, I think there is no point in re-emitting them as the WebSocket is closed anyway with an abnormal closure status code.
- The WebSocket browser implementation does not call the `onerror` listener if a network error occurs after the handshake completes.

cc: @websockets/admin

lpinca:
This library has ~15M monthly downloads, it's a dependency of many popular project, it's used by some big corporations but almost no one actively participate in its development. This is weird.

@3rd-Eden can you please weigh in when you have time?

3rd-Eden:
I simply do not know enough of Node.js net code internal to make a comment that will be remotely useful here. 

The performance loss is acceptable as we finally handle back pressure correctly. We can see if we can performance back in different ways.

lpinca:
Performance is the same, it's memory usage that's slightly higher but that's acceptable.

My doubt is whether or not we should re-emit network errors like `ECONNRESET`, `EPIPE`, `ECONNABORTED`. All of these were swallowed before version `3.3.3` and from #1256 it seems that the common action is to ignore them.

3rd-Eden:
@lpinca I agree, I don't think those errors are usefull because it just means that the connection got closed in some way.

lpinca:
Ok, thank you.

PR #1306, opened 2018-02-18
[minor] add zlibOptions to perMessageDeflate init

STRML:
It can be useful to adjust other options, and rather than support and document every single one, it is simpler to allow passing on an object wholesale.

`level` and `memLevel` are still supported but should be removed in 4.x.

I'm still working on solutions to https://github.com/nodejs/node/issues/8871 (tracked in https://github.com/nodejs/node/issues/15301). In the meantime, depending on workload, memory fragmentation can be greatly reduced by adjusting `chunkSize`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,15 +81,16 @@ The extension is disabled by default on the server and enabled by default on
 the client. It adds a significant overhead in terms of performance and memory
 consumption so we suggest to enable it only if it is really needed.
 
-Note that NodeJS has a variety of issues with high-performance compression,
+Note that Node.js has a variety of issues with high-performance compression,
 where increased concurrency, especially on Linux, can lead to
 [catastrophic memory fragmentation][node-zlib-bug] and slow performance.
 If you intend to use `permessage-deflate` in production, it is worthwhile
-to set up a test representative of your workload and ensure Node/zlib will
+to set up a test representative of your workload and ensure Node.js/zlib will
 handle it with acceptable performance and memory usage.
 
-Tuning of `permessage-deflate` can be done via the `zlibOptions` option,
-which is passed directly into [zlib.createDeflateRaw()][node-zlib-deflaterawdocs].
+Tuning of `permessage-deflate` can be done via the options defined below.
+You can also use `zlibOptions`, which is passed directly into
+[zlib.createDeflateRaw()][node-zlib-deflaterawdocs].
 Do not set `windowBits`, which is set dynamically by the client.
 
 See [the docs][ws-server-options] for more options.
@@ -108,7 +109,7 @@ const wss = new WebSocket.Server({
     // Other options settable:
     clientNoContextTakeover: true, // defaults to negotiated value
     serverNoContextTakeover: true, // defaults to negotiated value
-    clientMaxWindowBits: true,     // defaults to negotiated value
+    clientMaxWindowBits: 10,       // defaults to negotiated value
     serverMaxWindowBits: 10,       // defaults to negotiated value
     // Below options specified as default values
     concurrencyLimit: 10,          // limits zlib concurrency for perf
--- a/README.md
+++ b/README.md
@@ -89,9 +89,9 @@ to set up a test representative of your workload and ensure Node.js/zlib will
 handle it with acceptable performance and memory usage.
 
 Tuning of `permessage-deflate` can be done via the options defined below.
-You can also use `zlibOptions`, which is passed directly into
-[zlib.createDeflateRaw()][node-zlib-deflaterawdocs].
-Do not set `windowBits`, which is set dynamically by the client.
+You can also use `zlibDeflateOptions` and `zlibInflateOptions`, which is passed
+directly into the creation of
+[raw deflate/inflate streams][node-zlib-deflaterawdocs].
 
 See [the docs][ws-server-options] for more options.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -81,6 +81,43 @@ The extension is disabled by default on the server and enabled by default on
...
+  perMessageDeflate: {
+    zlibOptions: { // See zlib defaults
+      chunkSize: 128,
+      memLevel: 7,
+      level: 3,
+    },
+    // Other options settable:
+    clientNoContextTakeover: true, // defaults to negotiated value
+    serverNoContextTakeover: true, // defaults to negotiated value
+    clientMaxWindowBits: true,     // defaults to negotiated value
```
lpinca: This should be a number, or `false` because the value can only be a number in an extension negotiation response. Using `true` is the same of leaving it `undefined`.

lib/permessage-deflate.js
```diff
@@ -44,8 +44,7 @@ class PerMessageDeflate {
    *     use of a custom server window size
    * @param {(Boolean|Number)} options.clientMaxWindowBits Advertise support
    *     for, or request, a custom client window size
-   * @param {Number} options.level The value of zlib's `level` param
-   * @param {Number} options.memLevel The value of zlib's `memLevel` param
+   * @param {Number} options.zlibOptions Options to pass to zlib on deflate
```
lpinca: Type is `Object`.

lib/permessage-deflate.js
```diff
@@ -412,12 +411,20 @@ class PerMessageDeflate {
...
-      });
+      };
+      if (this._options.zlibOptions != null) {
+        if (typeof this._options.zlibOptions === 'number') {
+          throw new Error('Do not set `windowBits` on `zlibOptions`!');
+        }
+        Object.assign(zlibOptions, this._options.zlibOptions);
+      }
+
+      this._deflate = zlib.createDeflateRaw(zlibOptions);
```
lpinca: I would just pass the options object here, overriding the value of the `windowBits`.

```js
this._deflate = zlib.createDeflateRaw(
  Object.assign(
    {
      memLevel: this._options.memLevel,
      level: this._options.level,
      flush: zlib.Z_SYNC_FLUSH
    },
    this._options.zlibOption,
    { windowBits }
  )
);
```

lib/permessage-deflate.js
```diff
@@ -434,7 +441,7 @@ class PerMessageDeflate {
 
     this._deflate.write(data);
     this._deflate.flush(zlib.Z_SYNC_FLUSH, () => {
-      var data = bufferUtil.concat(
+      let data = bufferUtil.concat(
```
lpinca: Please keep `var` until we switch to `let` everywhere.

lib/permessage-deflate.js
STRML: The idea was to avoid the Object.assign() if nothing is passed (the far more common case).

STRML:
New commit has fixes for most comments, see note on Object.assign().

lib/permessage-deflate.js
lpinca: Is the perf impact big? The next version will probably be a major due to https://github.com/websockets/ws/pull/1302 so backward compatibility for `options.memLevel` and `options.level` can be removed.

STRML: Actually no, it won't be, because it's only per-connection, I had erroneously assumed it was per message. Will update.

STRML: Updated; one thing to note, setting `flush` should not be necessary because we pass `zlib.Z_SYNC_FLUSH` when flushing manually; please correct me if wrong.

lpinca: I was about to ask exactly this :smile: . I think you are correct bit I'm not 100% sure.

lpinca: You are also correct wrt to the stream instance. It is instantiated per message if context takeover is disabled.

lpinca:
Should we also do the same for the inflate stream?

STRML:
Added to the inflate & added test to verify options were passed onto the zlib streams.

lib/permessage-deflate.js
```diff
@@ -345,7 +344,13 @@ class PerMessageDeflate {
         ? zlib.Z_DEFAULT_WINDOWBITS
         : this.params[key];
 
-      this._inflate = zlib.createInflateRaw({ windowBits });
+      this._inflate = zlib.createInflateRaw(
+        Object.assign(
+          {},
+          this._options.zlibOptions,
```
lpinca: I think we should use a different option object. For the `chunkSize ` could be different from the one used for the deflate stream.

test/permessage-deflate.test.js
```diff
@@ -390,6 +390,55 @@ describe('PerMessageDeflate', function () {
       });
     });
 
+    it('honors the `zlibOptions` option', function (done) {
+      const lev9 = new PerMessageDeflate({ threshold: 0, zlibOptions: {level: 9, chunkSize: 128} });
```
lpinca: Can you please keep the limit of 80 chars, if possible?

STRML: There's tons of this file that goes out to 100 characters, does it really matter?

lpinca: I know, my idea is to add prettier sooner or later so feel free to ignore this comment.

STRML:
Split into two options.

doc/ws.md
```diff
@@ -71,8 +71,8 @@ provided then that is extension parameters:
   takeover.
 - `serverMaxWindowBits` {Number} The value of `windowBits`.
 - `clientMaxWindowBits` {Number} Request a custom client window size.
-- `level` {Number} The value of zlib's `level` param (0-9, default 8).
-- `memLevel` {Number} The value of zlib's `memLevel` param (1-9, default 8).
+- `zlibOptions` {Object} [Additional options][zlib-options] to pass to zlib
```
lpinca: `zlibDeflateOptions`, we also need to add `zlibInflateOptions`, I can do after merge if it's too boring.

lpinca:
Thank you.

STRML:
Thanks for the merge.

PR #1311, opened 2018-03-01
move dep to dev dep

xwa130:
When using `webpack` to compile, there are two warnings saying that 
```
Module not found: Error: Can't resolve 'utf-8-validate'
Module not found: Error: Can't resolve 'bufferutil'
```
This is because these two lib are in `devDependency` not in `dependency`, so when user run `npm i`, they got ignored. I suggest to move them back to `dependency`.

---
reference to another [issue](https://github.com/coinbase/gdax-node/issues/162):

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,6 @@
     "eslint-plugin-promise": "~3.6.0",
     "eslint-plugin-standard": "~3.0.0",
     "mocha": "~5.0.0",
-    "nyc": "~11.4.1",
+    "nyc": "~11.4.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -27,11 +27,12 @@
   },
   "dependencies": {
     "async-limiter": "~1.0.0",
-    "safe-buffer": "~5.1.0"
+    "bufferutil": "~3.0.0",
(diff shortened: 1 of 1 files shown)


lpinca:
From https://github.com/websockets/ws#ws-a-nodejs-websocket-library

> Note: This module does not work in the browser. The client in the docs is a reference to a back end with the role of a client in the WebSocket communication. Browser clients must use the native WebSocket object. To make the same code work seamlessly on Node.js and the browser, you can use one of the many wrappers available on npm, like isomorphic-ws.

If you don't want to use a wrapper like isomorphic-ws you need a shim like this https://github.com/heineiuo/isomorphic-ws/blob/d5ec6506fe8530bb331f07a99d4f2c4eb6a26273/browser.js and add a `browser` key in `package.json` which points to it: https://github.com/heineiuo/isomorphic-ws/blob/d5ec6506fe8530bb331f07a99d4f2c4eb6a26273/package.json#L6

lpinca:
Thanks for the PR but we can't merge this.

xwa130:
I'm sorry for the delayed reply. I appreciate your help.

PR #1313, opened 2018-03-02
Drop support for Node.js < 4.5.0

lpinca:
Node.js 4.x will go EOL at the end of April and all versions below 4.8.7 have at least one security vulnerability. I see no reason to keep supporting versions in the range 4.0.0 - 4.4.7.

I actually want to drop support for all 4.x versions but maybe it's better to wait a little more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,5 @@ node_js:
   - "8"
   - "6"
   - "4"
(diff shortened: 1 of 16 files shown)


3rd-Eden:
+1 kill it with fire.

PR #1324, opened 2018-03-10
[fix] remove buffer `noAssert` argument

BridgeAR:
Support for the `noAssert` argument dropped in the upcoming Node.js
10.x release. This removes it to make sure everything still works as
it should by already validating the input right away.

Refs: https://github.com/nodejs/node/pull/18395


lpinca:
Thank you.

Issue #1143, opened 2017-06-15
Allow to use url.URL as WebSocket constructor input


timanovsky:
- [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description

Currently `new Websocket(url)` expects string argument and fails with `TypeError: Parameter "url" must be a string, not object` if passed nodejs `URL` object. Would be convenient to pass URL objects too.

#### Reproducible in:

version: 3.0.0
Node.js version(s): 8
OS version(s):

#### Steps to reproduce:

1. 
```
const WebSocket = require('ws');
const { URL } = require('url');
new WebSocket(new URL("ws://localhost"))
```

### Expected result:

Properly initialized websocket

### Actual result:

Exception

lpinca:
This is consistent with the browser behavior where an error is thrown if the URL is not a string.
I have no strong opinions though.

timanovsky:
@lpinca true, but comparing to HTTP client libraries, which do accept `URL`, is also relevant

lpinca:
Yes it may be a nice to have feature. Anyway it should accept the object returned by both parsers: legacy and WHATWG.

An inefficient workaround is something like this.

```js
const url = require('url');
const WebSocket = require('ws');

const parsed = url.parse('ws://localhost:8080');

const ws = new WebSocket(url.format(parsed));
```

3rd-Eden:
> This is consistent with the browser behavior where an error is thrown if the URL is not a string.
I have no strong opinions though.

This is no longer the case. Browsers, or at least chrome, supports this.

lpinca:
Just tested and it also work on Firefox, but I think the spec has not been updated: https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface

PR #1329, opened 2018-03-15
[fix] Allow URL instances as URL for WebSocket constructors

3rd-Eden:
Addresses https://github.com/websockets/ws/issues/1143

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -27,7 +27,7 @@ class WebSocket extends EventEmitter {
   /**
    * Create a new `WebSocket`.
    *
-   * @param {String} address The URL to which to connect
+   * @param {(String|url.Url|url.URL)} address The URL to which to connect
    * @param {(String|String[])} protocols The subprotocols
    * @param {Object} options Connection options
    */
@@ -404,7 +404,7 @@ module.exports = WebSocket;
 /**
  * Initialize a WebSocket client.
  *
- * @param {String} address The URL to which to connect
+ * @param {(String|url.Url|url.URL)} address The URL to which to connect
  * @param {String[]} protocols The list of subprotocols
  * @param {Object} options Connection options
  * @param {String} options.protocol Value of the `Sec-WebSocket-Protocol` header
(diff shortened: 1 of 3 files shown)


test/websocket.test.js
```diff
@@ -5,6 +5,7 @@
 const assert = require('assert');
 const crypto = require('crypto');
 const https = require('https');
+const { URL } = require('url');
```
lpinca: Doesn't work on Node.js 4.

lib/websocket.js
```diff
@@ -463,9 +463,17 @@ function initAsClient (address, protocols, options) {
   }
 
   this._isServer = false;
-  this.url = address;
 
-  const serverUrl = url.parse(address);
+  let serverUrl;
```
lpinca: We are still using `var` when `const` can not be used.

3rd-Eden: FML, I've been writing to much ES6.

lpinca:
https://github.com/websockets/ws/blob/57d3dbb9e1b6d81132bbf2c18a35f09eab783a87/lib/websocket.js#L513

No longer works when using a WHATWG URL.

test/websocket.test.js
3rd-Eden: Node 4 doesn't have the URL instance either, so I've opted for adding in some horrible if statements in the test suite :/

lib/websocket.js
```diff
@@ -511,6 +519,9 @@ function initAsClient (address, protocols, options) {
   }
   if (options.host) requestOptions.headers.Host = options.host;
   if (serverUrl.auth) requestOptions.auth = serverUrl.auth;
+  else if (serverUrl.username && serverUrl.password) {
```
lpinca: Still not enough I think:

```
var u = new url.URL('ws://foo@example.org')
undefined
> u
URL {
  href: 'ws://foo@example.org/',
  origin: 'ws://example.org',
  protocol: 'ws:',
  username: 'foo',
  password: '',
  host: 'example.org',
  hostname: 'example.org',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: '' }
```

`auth` should be `'foo'` but it doesn't get set as `serverUrl.password` is falsy.

test/websocket.test.js
```diff
@@ -361,6 +362,15 @@ describe('WebSocket', function () {
 
         assert.strictEqual(ws.url, url);
       });
+
+      if (URL) {
+        it('accepts URL instances as url', function () {
```
lpinca: I would use `this.skip()` for consistency if  WHATWG is not supported.

lpinca:
I think this can be merged.

3rd-Eden:
yeah, sorry, my internet decided to shit it self today :/

Issue #1341, opened 2018-03-30
Explain the Behavior of server.close() More Precisely


anliting:
- [x] I've searched for any related issues and avoided creating a duplicate issue.

I think it would be good to explain the behavior of server.close() more precisely in the API docs:

- The pre-created HTTP server is not closed when 'server' is in the constructing options.
- No server is closed when 'noServer' is in the constructing options.

lpinca:
Please open a PR.

lpinca:
Fixed by https://github.com/websockets/ws/pull/1342.

PR #1342, opened 2018-03-30
[minor] Explain the behavior of server.close() more precisely (#1341)

anliting:


doc/ws.md
```diff
@@ -139,7 +139,11 @@ returned as a string.
 
 ### server.close([callback])
 
-Close the server and terminate all clients, calls callback when done.
+Close the server and terminate all clients, calls callback when done. The
```
lpinca: Can you please reword like this?

```
Close the HTTP server if created internally, terminate all clients and call
callback when done. If an external HTTP server is used via the `server` or
`noServer` constructor options, it must be closed manually.
```

I think it's more accurate.

Thank you.

anliting: Okay, I have updated it.

lpinca:
Thank you.

anliting:
Thank you too. `ws` is a great work.

Issue #1344, opened 2018-03-31
We should provide a valid WebSocket for Browser, or throw a Error explicitly.


zixia:
- [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description

When a module that is using `import * as WebSocket from 'ws'`, and it's user use it in the Browser, we will get a `WebSocket` provided by `ws` which does not work.

#### Reproducible in:

version: 5.1.0
Node.js version(s): 9.8.0
OS version(s): Darwin zixia-pro.local 16.7.0 Darwin Kernel Version 16.7.0: Mon Nov 13 21:56:25 PST 2017; root:xnu-3789.72.11~1/RELEASE_X86_64 x86_64

#### Steps to reproduce:

1. `import * as WebSocket from 'ws'` in a javascript file
2. use WebPack to pack it into a bundle
3. run it in Browser

### Expected result:

I understood that the `ws` module is designed for the Node.js environment, we should not expect it work in the Browser. 

However, when users used `ws` in their module code, sometimes the module user might use the module in the Browser, and without the knowledge of the limitation of the `ws` module.

#### Suggested Behavior

It should be either throw an error explicitly or provide a valid `WebSocket` implementation for the Browser.

### Actual result:

It provides a not worked WebSocket for the Browser silently.

### Attachments:

This issue cost me 2 days to find out the problem, even I know the `ws` not work in the Browser.

I had wroten a NPM module named `@chatie/graphql` which use `ws` to connect with the GraphQL server, it work without any problem in the Node.js enviornment.

After some time, I want to use this module in my Ionic/Angular project, then I got a very strange error. After digging it deep, I found out that it's because I'm using `ws` in my `@chatie/graphql`. 

See more at: 
1. https://github.com/Chatie/graphql/issues/12
1. https://www.graph.cool/forum/t/cors-problem-no-access-control-allow-origin-header-is-present-in-the-subscriptions-endpoint/2988/8?u=huan

lpinca:
This is documented, it is the first thing written after the project description: https://github.com/websockets/ws#ws-a-nodejs-websocket-library.

That said, I'm happy to revisit https://github.com/websockets/ws/pull/661 if @3rd-Eden agrees.

zixia:
@lpinca Yes, I understand that it is documented very clearly.

This issue comes out because there might be a module using `ws`, but others module require this module without any knowledge of the limitation of `ws`.

It seems I just sent a duplicated PR very similar to #661. 

[...]

PR #1345, opened 2018-03-31
throw a exception when be imported into Browser (#1344)

zixia:
#1344

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,3 +1,3 @@
-module.exports = function() {
-  throw new Error('ws only supports Node.js! Please consider to use `isomorphic-ws` for both Node.js & Browser support! (See issue #1344 at https://github.com/websockets/ws/issues/1344 for more information.')
-}
+module.exports = function () {
+  throw new Error('ws only supports Node.js! Please consider to use `isomorphic-ws` for both Node.js & Browser support! (See issue #1344 at https://github.com/websockets/ws/issues/1344 for more information.');
+};
--- a/browser.js
+++ b/browser.js
@@ -1,3 +1,4 @@
 module.exports = function () {
-  throw new Error('ws only supports Node.js! Please consider to use `isomorphic-ws` for both Node.js & Browser support! (See issue #1344 at https://github.com/websockets/ws/issues/1344 for more information.');
+  throw new Error('ws does not work in the browser. ' +
+                  'Browser clients must use the native WebSocket object');
 };
--- a/browser.js
+++ b/browser.js
@@ -1,4 +1,8 @@
+'use strict';
+
 module.exports = function () {
-  throw new Error('ws does not work in the browser. ' +
-                  'Browser clients must use the native WebSocket object');
+  throw new Error(
+    'ws does not work in the browser. Browser clients must use the native ' +
+      'WebSocket object'
+  );
 };
--- a/browser.js
+++ b/browser.js
@@ -1,0 +1,3 @@
+module.exports = function() {
+  throw new Error('ws only supports Node.js! Please consider to use `isomorphic-ws` for both Node.js & Browser support! (See issue #1344 at https://github.com/websockets/ws/issues/1344 for more information.')
+}
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "main": "index.js",
   "browser": "browser.js",
   "files": [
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
   "author": "Einar Otto Stangvik <einaros@gmail.com> (http://2x.io)",
   "license": "MIT",
   "main": "index.js",
+  "browser": "browser.js",
   "files": [
     "index.js",
```
lpinca: Add `browser.js`

browser.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function () {
+  throw new Error('ws only supports Node.js! Please consider to use `isomorphic-ws` for both Node.js & Browser support! (See issue #1344 at https://github.com/websockets/ws/issues/1344 for more information.');
```
lpinca: Please keep the limit of 80 chars per line. Also, I don't want to endorse any particular wrapper, we are already doing  it in the README and there is no need to point to that particular issue, there a few already which are all duplicates.

I'd just use a very simple message like "ws does not work in the browser. Browser clients must use the native `WebSocket` object". I think this is sufficient to address the user into the right direction.

zixia:
@lpinca Changed as you request.

lpinca:
cc: @3rd-Eden

3rd-Eden:
I'm fine with this but it might be a major release as there is a chance of breaking people's code if they do a weird fallback like:

```js
var WS = require('ws') || WebSocket;
```

zixia:
@3rd-Eden Thanks for accepting this.

I feel this will not breaking anything because according to the current code base, the return value of `require('ws')` will neither be `undefined` nor `null`, so the `|| WebSocket` part in the fallback you mentioned will never get chance to be executed.

lpinca:
@zixia I think @3rd-Eden wanted to write:

```js
var WS = WebSocket || require('ws');
```

I think this will throw now if the code is bundled for Node.js?

zixia:
@lpinca I believe it will not throw in the Node.js environment.

Because in Node.js `global.WebSocket` will be undefined, then `WS` will be set as `require('ws')`, which should be right.

lpinca:
Yes but isn't the browser shim used in that case?

lpinca:
The `--no-browser-field` option can be used to make it work again but I think it's still a breaking change.

zixia:
Yes, you are right. 

I did not consider the `browserify` case.

browser.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+module.exports = function () {
+  throw new Error(
```
shogowada: I've read https://github.com/websockets/ws/issues/1344, but I am personally not sure why it needs to throw when it can fallback. It looks like we are waiting for @3rd-Eden 's input?

Maybe it's already discussed at other places, but should this [fallback to browser's `WebSocket`](https://github.com/heineiuo/isomorphic-ws/blob/master/browser.js), while announcing that `WebSocket.Server` doesn't work on browser?

I think that would make many projects' dependency graph a bit cleaner and more straightforward when they need to implement a platform agnostic web socket client.

lpinca: @shogowada this has been discussed ad nauseam, we don't want to maintain a browser shim because there are so many differences/edge cases that makes it a pain.

Refs: https://github.com/websockets/ws/commit/e54d45fbab3b19c2940e9057ce1e7b8f105873e0

shogowada: Thanks for the additional context.

In that case, instead of throwing (that appears to break some), would a warning work?

Something like this:

```javascript
module.exports = function () {
  // Almost a copy & paste from the README + link to the ref you gave me for future people like me🙂
  console.warn('ws module does not work in the browser. Browser clients must use the native WebSocket object. To make the same code work seamlessly on Node.js and the browser, you can use one of the many wrappers available on npm, like isomorphic-ws. For more context, see https://github.com/websockets/ws/commit/e54d45fbab3b19c2940e9057ce1e7b8f105873e0.')
}
```

lpinca: I don't think it helps, it still breaks existing code. Throwing an error is the right thing to do imho.

lpinca:
Finally merged. Thank you.

zixia:
Glad to see that, cheers!

PR #1355, opened 2018-04-07
Improve docs and examples

lpinca:
- Improve documentation for the `port`, `server`, and `noServer` options.
- Remove Express example in favour of a general external server integration example.
- Add "noServer" mode example.


lpinca:
cc: @rgranger @kmathmann

rgranger:
Very nice ! Maybe add that the option { port } will automatically start the internal httpServer (unlike the option server and noServer that waits for the use to call .listen(port).

Or maybe add the functions .listen(port) and .close() to the WebsocketServer ? That would allow the user to close and open the server without having to create a new instance.

PR #1365, opened 2018-04-21
fix server param in sample

erlangp:


lpinca:
That is ES2015 shorthand syntax. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer.

PR #1372, opened 2018-04-29
[doc] "Multiple servers sharing...", add `request` to emit arguments

lukeavsec:
I spent 90-minutes trying to figure out why my `connection` handler `request` argument was undefined. I thought making the docs for the `noServer` configuration more idiomatic with the expected behavior of the `server` configuration might save others from frustration in the future.


lpinca:
Thanks.

PR #1402, opened 2018-06-18
Added options.maxPayload for websocket clients

Ban44n:
I came across a situation where a websocket client instance needs to be protected against large messages. This pull request makes the option.maxPayload option available for client instances.

### DIFF
--- a/bench/speed.js
+++ b/bench/speed.js
@@ -54,7 +54,9 @@ if (cluster.isMaster) {
 
   const runConfig = (useBinary, roundtrips, size, cb) => {
     const data = randomBytes.slice(0, size);
-    const ws = new WebSocket(`ws://localhost:${port}`);
+    const ws = new WebSocket(`ws://localhost:${port}`, {
+      maxPayload: 600 * 1024 * 1024
+    });
     var roundtrip = 0;
     var time;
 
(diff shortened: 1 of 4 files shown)


lpinca:
Can you please elaborate a bit? I think that if a server sends frames so big that can't be handled by clients, it's a server problem. There is no DoS in this case.

Ban44n:
I am using websockets as a network protocol in an untrusted peer-to-peer network. Each peer in this network has a server and client role. Because peers do not trust other peers in this network, there must be a DoS protection on the websocket client interface and hence this pull.

lpinca:
Ok thanks for the explanation. The option needs to be documented in https://github.com/websockets/ws/blob/master/doc/ws.md and at least one test should be added before this can be merged.

lpinca:
I also think this is a breaking change. The connection should not be closed if a frame bigger than 100 MiB is received on the client without bumping the major version.

Ban44n:
I have included the requested testcase for option.maxPayload. Furthermore, perMessageDeflate in client mode now also works with options.maxPayload.

Ban44n:
I did not find a section on maxPayload sizes in the ietf for v13:

https://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-13

I guess we do not have to bump the protocolVersion for this change.

lpinca:
`maxPayload` is not part of the spec. What I meant is that this change is a [semver](https://semver.org/) major change and as such it can't go in a `ws@5.x` release. It requires `ws@6`.

lpinca:
Thank you.

PR #1430, opened 2018-08-07
Support Promise for handleProtocols

recitativo:
To enable asynchronous handling of protocols, support Promise for return value from handleProtocols.

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -209,7 +209,7 @@ class WebSocketServer extends EventEmitter {
             return abortHandshake(socket, code || 401, message, headers);
           }
 
-          this.completeUpgrade(extensions, req, socket, head, cb);
+          this.handleProtocols(extensions, req, socket, head, cb);
         });
         return;
       }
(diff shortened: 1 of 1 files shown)


lpinca:
What is the use case of this? Any reason for making the subprotocol selection async? I would prefer to not use promises as it will set a precedent. If we add support for promises in `handleProtocols` people will start requesting promise support everywhere.

3rd-Eden:
I agree with that sentiment. If we start supporting Promises, we should go through the effort of making _everything_ `async` `await` compatible.

recitativo:
In my use case, http request is required in subprotocol selection. In the other motivation, libraries that only support asynchronous processing are increasing recently, I think it is better to promote overall Promise support. Thanks.

lpinca:
I suggest to use something like this:

```js
const http = require('http');

server.on('upgrade', async (request, socket, head) => {
  let protocol = req.headers['sec-websocket-protocol'] || '';
  protocol = protocol.trim().split(/ *, */);

  try {
    protocol = await selectProtocol(protocol);
  } catch (e) {
    socket.destroy();
    return
  }

  wss.on('headers', () => {
    headers.push(`Sec-WebSocket-Protocol: ${protocol}`);
  });
  wss.handleUpgrade(request, socket, head, (ws) => {
    wss.emit('connection', ws, request);
  });
});

  // ...
});
```

I think in most cases subprotocol selection does not require to be async as it's just a matter of selecting the first subprotocol from the ones sent by the client. Usually a simple whitelist is kept for this on the server.

recitativo:
Thank you for your suggestion. It's nice idea to override some functions. I have plan to handle each protocol in its plugin, including message processing closely related with authentication. Your suggestion seems to satisfy my requirements. I will try it after summer vacation. Thanks again!

PR #1453, opened 2018-10-03
Make `WebSocketServer#close()` emit the `'close'` event

### LINKED ISSUE #1375: getter for httpServer internal
- [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description
There isn't 'close' event for server, it may be useful the 'listening' event  and 'close' event  of httpServer. I think we should make a httpServer getter of  Server that is actually private.

lpinca:
As titled. Fixes #1375.


rupamkhaitan:
Dint understand this feature.. was the close event not emitted all this while?

3rd-Eden:
Correct.

PR #1454, opened 2018-10-07
Tip for other people in documentation

techsin:


README.md
```diff
@@ -112,7 +112,7 @@ const wss = new WebSocket.Server({
     // Other options settable:
     clientNoContextTakeover: true, // Defaults to negotiated value.
     serverNoContextTakeover: true, // Defaults to negotiated value.
-    clientMaxWindowBits: 10,       // Defaults to negotiated value.
+    clientMaxWindowBits: 10,       // Defaults to negotiated value. Right now, setting this causes conection to fail in FF.
```
lpinca: I think it's better to completely remove `clientMaxWindowBits` from the example. There is nothing wrong with FF, it's the configuration that makes the offer sent by FF unacceptable.

We already have this sentence "See docs for more options." before the example so that should be sufficient.

techsin: i agree removing is better.

lpinca:
Did you close this by mistake? You can update this PR if you want.

techsin:
updated it

lpinca:
Thank you.

Issue #1226, opened 2017-10-31
Clear the send queue when the other peer closes the connection


lpinca:
#### Description

permessage-deflate is enabled and one of the peer has queued data waiting for it to be compressed, framed and sent over the connection. Meanwhile, the other peer closes the connection. What should we do?

Currently the data processing keeps going as nothing happened and the [callback](https://github.com/websockets/ws/blob/48b0496879899f35602856d80460926a4a6c299d/doc/ws.md#websocketsenddata-options-callback) of each enqueued send operation, if any, is invoked with an error when trying to write to the closed socket (after the `'close'` event). This a waste of computational power. There is no reason to compress, frame, and try to send the data if we know beforehand that the socket is closed.

One way to work around the issue is to remove the enqueued send operations after invoking their callback with an error and before emitting the `'close'` event.
The issue happens when a message is already being compressed so the [queue](https://github.com/websockets/ws/blob/48b0496879899f35602856d80460926a4a6c299d/lib/Sender.js#L37) should be cleared when compression [completes](https://github.com/websockets/ws/blob/48b0496879899f35602856d80460926a4a6c299d/lib/Sender.js#L333).

Thoughts?

#### Reproducible in:

version: >=0.6.0
Node.js version(s): all
OS version(s): all

#### Steps to reproduce:

```js
const crypto = require('crypto');
const WebSocket = require('ws');

const wss = new WebSocket.Server({
  perMessageDeflate: { threshold: 0 },
  port: 3000
}, () => {
  const ws = new WebSocket('ws://localhost:3000');

  ws.on('open', () => {
    setTimeout(() => ws.terminate(), 10);
  });
});

wss.on('connection', (ws) => {
  ws.on('close', () => console.log('close'));

  const data = crypto.randomBytes(1024);

  for (let i = 0; i < 1024; i++) {
    ws.send(data, (err) => {
      if (err) console.log(err);
    });
  }
});
```

PR #1464, opened 2018-10-23
Do not waste time compressing when socket is closed

Evertras:
Addresses #1226 - I agree with @lpinca 's approach in general.  This has been tested with the offending code listed in the issue and it correctly short circuits while still providing a relevant error to each pending request.  I've also added a unit test that fails without the added code as it tries to write/compress too much.

Not sure how I feel about having to add the extra `on: () => {}` to the Sender unit tests for the socket object, but these are mocks anyway.  Open to suggestions if this isn't preferred.

### DIFF
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -28,7 +28,7 @@ class Sender {
     this._deflating = false;
     this._queue = [];
 
-    this._socket.once('close', () => {
+    this._socket.prependOnceListener('close', () => {
       const err = new Error(
         `WebSocket is not open: readyState ${this._socket.readyState} `
       );
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -30,7 +30,7 @@ class Sender {
 
     this._socket.once('close', () => {
       const err = new Error(
-        `WebSocket is not open: readyState CLOSING`
+        `WebSocket is not open: readyState 2 (CLOSING)`
       );
 
       while (this._queue.length) {
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -28,23 +28,19 @@ class Sender {
     this._deflating = false;
     this._queue = [];
 
-    this._socket.once('close', () => {
-      const err = new Error(
-        `WebSocket is not open: readyState 2 (CLOSING)`
-      );
+    if (this._extensions[PerMessageDeflate.extensionName]) {
+      this._socket.once('close', () => {
+        const err = new Error('WebSocket is not open: readyState 2 (CLOSING)');
 
-      while (this._queue.length) {
-        const params = this._queue.shift();
+        while (this._queue.length) {
+          const params = this._queue.shift();
+          const cb = params[params.length - 1];
 
-        this._bufferedBytes -= params[1].length;
-
-        const cb = params[params.length - 1];
-
-        if (typeof cb === 'function') {
-          cb(err);
+          this._bufferedBytes -= params[1].length;
+          if (typeof cb === 'function') cb(err);
         }
-      }
-    });
+      });
+    }
   }
 
   /**
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -38,7 +38,11 @@ class Sender {
 
         this._bufferedBytes -= params[1].length;
 
-        params[params.length - 1](err);
+        const cb = params[params.length - 1];
+
+        if (typeof cb === 'function') {
+          cb(err);
+        }
       }
     });
   }
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -30,7 +30,7 @@ class Sender {
 
     this._socket.once('close', () => {
       const err = new Error(
-        `WebSocket is not open: readyState ${this._socket.readyState} `
+        `WebSocket is not open: readyState CLOSING`
       );
 
       while (this._queue.length) {
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -363,7 +363,6 @@ class Sender {
       const params = this._queue.shift();
 
       this._bufferedBytes -= params[1].length;
-
       params[0].apply(this, params.slice(1));
     }
   }
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -28,7 +28,7 @@ class Sender {
     this._deflating = false;
     this._queue = [];
 
-    this._socket.on('close', () => {
+    this._socket.once('close', () => {
       const err = new Error(
         `WebSocket is not open: readyState ${this._socket.readyState} `
       );
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -28,7 +28,7 @@ class Sender {
     this._deflating = false;
     this._queue = [];
 
-    this._socket.prependOnceListener('close', () => {
+    this._socket.once('close', () => {
       const err = new Error(
         `WebSocket is not open: readyState ${this._socket.readyState} `
       );
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -27,6 +27,20 @@ class Sender {
     this._bufferedBytes = 0;
     this._deflating = false;
     this._queue = [];
+
+    this._socket.on('close', () => {
+      const err = new Error(
+        `WebSocket is not open: readyState ${this._socket.readyState} `
+      );
+
+      while (this._queue.length) {
+        const params = this._queue.shift();
+
+        this._bufferedBytes -= params[1].length;
+
+        params[params.length - 1](err);
+      }
+    });
   }
 
   /**
@@ -349,6 +363,7 @@ class Sender {
       const params = this._queue.shift();
 
       this._bufferedBytes -= params[1].length;
+
       params[0].apply(this, params.slice(1));
     }
   }
--- a/test/sender.test.js
+++ b/test/sender.test.js
@@ -6,17 +6,13 @@ const PerMessageDeflate = require('../lib/permessage-deflate');
 const Sender = require('../lib/sender');
 
 class MockSocket {
-  constructor ({ write, on, once, prependOnceListener } = {}) {
+  constructor ({ write, once } = {}) {
     if (write) this.write = write;
-    if (on) this.on = on;
     if (once) this.once = once;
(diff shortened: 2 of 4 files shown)


test/sender.test.js
```diff
@@ -42,7 +42,8 @@ describe('Sender', function () {
         write: (data) => {
           assert.strictEqual(data[0] & 0x40, 0x40);
           if (++count === 3) done();
-        }
+        },
+        on: () => {}
```
lpinca: To avoid adding this every time, we can use something like this

```js
class FakeSocket {
  constructor({ write, on } = {}) {
    if (write) this.write = write;
    if (on) this.on = on;
  }

  write() {}
  on() {}
}
```

and use it like

```js
const socket = new FakeSocket({
  write: (data) => {
    assert.strictEqual(data[0] & 0x40, 0x40);
    if (++count === 3) done();
  }
});
const sender = new Sender(socket, { 'permessage-deflate': perMessageDeflate });
```

lpinca:
Thank you for the PR. Some thoughts in addition to the comments:

1. Each callback should ideally be called before the `'close'` event is emitted on the `WebSocket`. The current approach relies to the fact that the `'close'` event listener is added to the `net.Socket` before the [`'close'` listener](https://github.com/websockets/ws/blob/d2317b1d8dda7d36f1ba952f533ff45e2464ae52/lib/websocket.js#L750) that actually emit `'close'` on the `WebSocket`. This is good and should work but there might be a chunk that is already being compressed when the `'close'` event is emitted on the `net.Socket` and this is not caught because it is no longer in the queue. As a result the callback for this message could be invoked after the `'close'` event is emitted on the `WebSocket` and after all other callback that are short circuited.
1. Each callback should find the ready state to `CLOSING`. If I'm not wrong when the queue is cleared the `WebSocket` ready state is still `OPEN` so the user might surprised to get the error and have the ready state set to `OPEN`.

The second point should be easy to fix, it should be sufficient to set the ready state to `CLOSING` in the `'close'` listener added in the `Sender` before clearing the queue.

The first point is harder and I'm not sure how to fix that.

test/sender.test.js
Evertras: Sounds good, will switch them over.

lib/sender.js
```diff
@@ -27,6 +27,20 @@ class Sender {
...
+      const err = new Error(
+        `WebSocket is not open: readyState ${this._socket.readyState} `
+      );
+
+      while (this._queue.length) {
+        const params = this._queue.shift();
+
+        this._bufferedBytes -= params[1].length;
+
+        params[params.length - 1](err);
```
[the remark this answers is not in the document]
Evertras: Ah, thought it was required.  Adding a check.

Evertras:
For the second point, my testing is showing the ready state as "closed".  Is "closing" a better state here?

lib/sender.js
```diff
@@ -27,6 +27,20 @@ class Sender {
     this._bufferedBytes = 0;
     this._deflating = false;
     this._queue = [];
+
+    this._socket.on('close', () => {
+      const err = new Error(
+        `WebSocket is not open: readyState ${this._socket.readyState} `
```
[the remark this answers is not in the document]
Evertras: Good point.  Would it be better to have a simple generic message here?  The original intent was to provide consistency with what the user would get as an error regardless of whether their send attempt was queued or not, but since it was already inconsistent a generic message wouldn't break anything here.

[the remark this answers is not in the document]
lpinca: Maybe we can pass the `Sender` a reference of the `WebSocket` so we can:

1. Set the ready state to `CLOSING`
2. Use it to make error message consistent.

What do you think?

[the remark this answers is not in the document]
Evertras: I've been trying to avoid that because I don't like how it hard links Sender to WebSocket.  A Sender doesn't need a WebSocket to function.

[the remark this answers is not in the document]
lpinca: I agree, the same is also true for the `Receiver` but we add a WebSocket reference to it in order to use it in the event listeners.

lib/sender.js
```diff
@@ -27,6 +27,24 @@ class Sender {
     this._bufferedBytes = 0;
     this._deflating = false;
     this._queue = [];
+
+    this._socket.prependOnceListener('close', () => {
```
lpinca: Do we need `prependOnceListener()`? Or is it used only to ensure that the listener is added before the one added in `websocket.js`? In the latter case I would just add a comment in `WebSocket#setSocket()` to specify that the `Sender` must be instantiated before adding the `'close'` listener to the `net.Socket`.

Evertras: Turns out we do actually need it.  I was testing a different ready state in my initial pass.

lpinca: Hmm ok, that's strange as this is added before the other anyway. I'm not a fan of `prependOnceListener()` because it will add the listener before all other listeners including the ones added by Node.js itself.

lpinca:
> I'm not sure how to reproduce the hypothetical you listed, is there an example you could provide?

I'll try to create a test case when I get a chance. Thank you.

lib/sender.js
Evertras: Let me poke it again, I may end up blaming jet lag. :D

Evertras: Yep, going to blame jet lag.  Sorry for the confusion.  We do not need prepend, I'll switch it over to the comment instead.  The issue was I was switching `prepend` to `once` in the Sender but not in the MockSocket for the test.

lib/sender.js
```diff
@@ -27,6 +27,24 @@ class Sender {
     this._bufferedBytes = 0;
     this._deflating = false;
     this._queue = [];
+
+    this._socket.once('close', () => {
+      const err = new Error(
+        `WebSocket is not open: readyState CLOSING`
```
lpinca:
```suggestion
        `WebSocket is not open: readyState 2 (CLOSING)`
```

lpinca:
@Evertras here is the test case

```js
const EventEmitter = require('events');
const { randomBytes } = require('crypto');

const ee = new EventEmitter();

  perMessageDeflate: { threshold: 0 },
  port: 0
}, function () {
    perMessageDeflate: { threshold: 0 }
  });

  ws.on('open', function () {
    const pmd = ws._extensions['permessage-deflate'];
    const compress = pmd.compress;

    // Monkey patch compress to show that deflate ends after the `'close'` event
    // is emitted.
    pmd.compress = function (data, fin, callback) {
      compress.call(pmd, data, fin, function (err, buf) {
        console.log('deflate done');
        callback(err, buf);
      });
    };

    ws.send(randomBytes(1024 * 1024), function (err) {
      if (err) console.log(err);
    });

    // Make the other peer destory the socket while data is being compressed.
    ee.emit('destroy');
  });

  ws.on('close', function () {
    console.log('close');
  });
});

wss.on('connection', function (ws) {
  ee.on('destroy', function () {
    ws.terminate();
  });
});
```

Evertras:
I see.  The deflate has already started by the time the close event from the socket is received, so by the time we're even aware that the socket has closed then zlib has already started to do its work.  The closest I could get to an interruption is in [this callback](https://github.com/websockets/ws/blob/master/lib/permessage-deflate.js#L434) and at that point all the expensive work is done.

I'm not sure if it's even possible to completely remove this scenario.  Once it's passed to zlib, some compression may take place while the socket is closing since it's all batched up in the `flush` call.  I'm also not sure if it's worth trying to chase.

lpinca:
Yes, I think there is not much we can do about it, we could check if the socket is closed [here]( https://github.com/websockets/ws/blob/d2317b1d8dda7d36f1ba952f533ff45e2464ae52/lib/sender.js#L335) so at least we avoid framing that chunk but I'm not sure if it's worth the effort, it's really and edge case.

The only thing that bugs me is that when this happens the order of callbacks is no longer respected and that is why in the original issue I thought it was better to clear the send queue [here](https://github.com/websockets/ws/blob/d2317b1d8dda7d36f1ba952f533ff45e2464ae52/lib/sender.js#L335) but if we do that, callbacks are invoked after the `'close'` event is emitted.

lpinca:
It could be addressed by refactoring the already complex close procedure, for example by emitting the `'close'` event only when both the `Receiver` and `Sender` complete their processing.

That said, I think this PR is already a big improvement and could land as is, just give me a few days to review it again (sorry I'm kinda busy).

Sometimes I think adding permessage-deflate was a mistake, the amount of complexity it added is huge.

Thank you.

Evertras:
Agreed there are ways to improve it, but yeah, it'll be complicated.  One step at a time I think.  No rush on approval, thanks for all the input!

lpinca:
@Evertras I was thinking that maybe we can simplify everything by simply clearing the queue here https://github.com/websockets/ws/blob/7e061bce4c78affbd9f30a4ffeadb8fb669d9b01/lib/sender.js#L335 without even calling the callbacks.

For example, consider this example:

```js
const { createConnection, createServer } = require('net');
const { randomBytes } = require('crypto');

function callback(err) {
  console.error(err);
}

server.on('connection', function (socket) {
  const chunk = randomBytes(1024);

  socket.on('error', function (err) {
    console.error(err);
  });

  while (socket.write(chunk)) {}

  // Write some more.
  for (var i = 0; i < 16; i++) {
    socket.write(chunk, callback);
  }
});

server.listen(function () {
  const socket = createConnection(this.address().port);

  socket.on('connect', function () {
        setTimeout(function () {
          callback();
          socket.destroy();
        }, 100);
      }
    });

    socket.pipe(writable);
  });
});
```

queued callbacks are not called. My point is that even when not using permessage-delfate, there are already cases where callbacks are not called if the connection is not closed cleanly.

What do you think?

lpinca:
I'll merge this in the meanwhile and think about https://github.com/websockets/ws/pull/1464#issuecomment-435578571 a little more before taking any action.

lpinca:
Thank you!

Evertras:
Thanks for the merge!  I've been in zombie mode for the past week due to a nasty illness, sorry for not responding earlier.

PR #1471, opened 2018-11-10
Ignore callbacks when clearing the send queue

### LINKED ISSUE #1226: Clear the send queue when the other peer closes the connection

lpinca:
Do not invoke callbacks when clearing the send queue due to premature
socket closure.

Refs: https://github.com/websockets/ws/pull/1464#issuecomment-435578571
Fixes #1226

### DIFF
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -27,20 +27,6 @@ class Sender {
     this._bufferedBytes = 0;
     this._deflating = false;
     this._queue = [];
-
-    if (this._extensions[PerMessageDeflate.extensionName]) {
-      this._socket.once('close', () => {
-        const err = new Error('WebSocket is not open: readyState 2 (CLOSING)');
-
-        while (this._queue.length) {
-          const params = this._queue.shift();
-          const cb = params[params.length - 1];
-
-          this._bufferedBytes -= params[1].length;
-          if (typeof cb === 'function') cb(err);
-        }
-      });
-    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


lpinca:
cc: @Evertras

PR #1473, opened 2018-11-22
[feature] allow changing maxPayload of open socket

mvduin:
This enables using a small maxPayload by default and then switching to a large value after the client has been authenticated in some way.

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -192,8 +192,7 @@ class WebSocketServer extends EventEmitter {
     if (this.options.perMessageDeflate) {
       const perMessageDeflate = new PerMessageDeflate(
         this.options.perMessageDeflate,
-        true,
-        this.options.maxPayload
+        true
       );
 
       try {
@@ -302,7 +301,8 @@ class WebSocketServer extends EventEmitter {
     socket.write(headers.concat('\r\n').join('\r\n'));
     socket.removeListener('error', socketOnError);
 
-    ws.setSocket(socket, head, this.options.maxPayload);
+    ws.maxPayload = this.options.maxPayload;
+    ws.setSocket(socket, head);
 
     if (this.clients) {
       this.clients.add(ws);
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -47,6 +47,7 @@ class WebSocket extends EventEmitter {
     this._closeCode = 1006;
     this._extensions = {};
     this._isServer = true;
+    this._maxPayload = 0;
     this._receiver = null;
     this._sender = null;
     this._socket = null;
@@ -100,6 +101,23 @@ class WebSocket extends EventEmitter {
   /**
    * @type {Number}
    */
+  get maxPayload() {
+    return this._maxPayload;
+  }
+
+  set maxPayload(value) {
+    this._maxPayload = value | 0;
+
+    if (this._receiver) this._receiver._maxPayload = value | 0;
+
+    if (this._extensions[PerMessageDeflate.extensionName]) {
+      this._extensions[PerMessageDeflate.extensionName]._maxPayload = value | 0;
+    }
+  }
+
+  /**
+   * @type {Number}
+   */
   get bufferedAmount() {
     if (!this._socket) return 0;
 
@@ -121,14 +139,13 @@ class WebSocket extends EventEmitter {
    *
    * @param {net.Socket} socket The network socket between the server and client
    * @param {Buffer} head The first packet of the upgraded stream
-   * @param {Number} maxPayload The maximum allowed message size
    * @private
    */
-  setSocket(socket, head, maxPayload) {
+  setSocket(socket, head) {
     const receiver = new Receiver(
       this._binaryType,
       this._extensions,
-      maxPayload
+      this._maxPayload
     );
 
     this._sender = new Sender(socket, this._extensions);
@@ -457,6 +474,7 @@ function initAsClient(address, protocols, options) {
   }
 
   this._isServer = false;
+  this.maxPayload = options.maxPayload;
 
   var parsedUrl;
 
@@ -505,8 +523,7 @@ function initAsClient(address, protocols, options) {
   if (options.perMessageDeflate) {
     perMessageDeflate = new PerMessageDeflate(
       options.perMessageDeflate !== true ? options.perMessageDeflate : {},
-      false,
-      options.maxPayload
+      false
     );
     options.headers['Sec-WebSocket-Extensions'] = extension.format({
       [PerMessageDeflate.extensionName]: perMessageDeflate.offer()
@@ -606,6 +623,7 @@ function initAsClient(address, protocols, options) {
 
         if (extensions[PerMessageDeflate.extensionName]) {
           perMessageDeflate.accept(extensions[PerMessageDeflate.extensionName]);
+          perMessageDeflate._maxPayload = this._maxPayload;
           this._extensions[PerMessageDeflate.extensionName] = perMessageDeflate;
         }
       } catch (err) {
@@ -614,7 +632,7 @@ function initAsClient(address, protocols, options) {
       }
     }
 
-    this.setSocket(socket, head, options.maxPayload);
+    this.setSocket(socket, head);
   });
 }
 
--- a/test/websocket.test.js
+++ b/test/websocket.test.js
@@ -164,6 +164,75 @@ describe('WebSocket', function() {
       });
     });
 
+    describe('`maxPayload`', function() {
+      it('defaults to 100 MiB', function() {
+        const ws = new WebSocket('ws://localhost', {
+          agent: new CustomAgent()
+        });
+
+        assert.strictEqual(ws.maxPayload, 100 * 1024 * 1024);
+      });
+
+      it('can be overridden using the `maxPayload` option', function() {
+        const maxPayload = 123;
+        const ws = new WebSocket('ws://localhost', {
+          maxPayload,
+          agent: new CustomAgent()
+        });
+
+        assert.strictEqual(ws.maxPayload, maxPayload);
+      });
+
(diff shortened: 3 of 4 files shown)


lpinca:
Can you please elaborate on why this is useful? Isn't it sufficient to always use a large enough value?

mvduin:
For the same reason maxPayload is useful in the first place: to help protect the server. Having a 100MB maxPayload means an attacker can easily consume that amount of the server's memory per connection established. Having a small maxPayload means an attacker would instead have to resort to flooding messages or connections, both of which can be dealt with by other means.

lpinca:
Yes but why changing it dynamically? If you know your clients will say 1 MiB at max just always set it to 1 MiB no? What's the point of changing the value during the lifetime of connection?

mvduin:
Because it may be difficult to evaluate how large it needs to be. We trust our clients however, and they have to authenticate themselves before they are allowed to send any "real" messages, so all that matters is that connections are prevented from sending large payloads until this authentication exchange has completed.

lpinca:
Is there any reason to authenticate the clients after the WebSocket is opened? Can't it be done during the handshake? Or do need to allow both authenticated and unauthenticated connections?

mvduin:
Using websocket messages for the authentication exchange is a lot more flexible, and it's too late to change anyway without breaking compatibility with existing clients.

Is there any reason to disallow modification of maxPayload after socket creation? It seems like a completely unnecessary limitation to me.

lpinca:
I'm not super happy because it's not a standard attribute. It's a deviation from the WHATWG interface.

mvduin:
But isn't that true for maxPayload in general already?

lpinca:
Yes but it's not exposed `ws.maxPayload` :)

mvduin:
I don't see how it's a good thing to hide an attribute that exists, has a functional effect, and whose initialization is already part of the API.

Also, it is mainly important for connections of a websocket server, which isn't covered by WHATWG at all anyway.

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
       });
     });
 
+    describe('`maxPayload`', function() {
+      it('defaults to 100MB upon "open"', function(done) {
```
lpinca:
```suggestion
      it('defaults to 100 MiB upon "open"', function(done) {
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+          const ws = new WebSocket(`ws://localhost:${wss.address().port}`);
+
+          ws.onopen = () => {
+            assert.strictEqual(ws.maxPayload, 100 * 1024 * 1024);
+            wss.close(done);
+          };
+        });
+      });
+
+      it('can be overridden using options.maxPayload', function(done) {
```
lpinca:
```suggestion
      it('can be overridden using the `maxPayload` option', function(done) {
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+        const maxPayload = 123;
+
```
lpinca:
```suggestion
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
       });
     });
 
+    describe('`maxPayload`', function() {
+      it('defaults to 100MB upon "open"', function(done) {
+        const wss = new WebSocket.Server({ port: 0 }, () => {
+          const ws = new WebSocket(`ws://localhost:${wss.address().port}`);
+
+          ws.onopen = () => {
```
lpinca:
```suggestion
          ws.on('open', () => {
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+            assert.strictEqual(ws.maxPayload, 100 * 1024 * 1024);
+            wss.close(done);
+          };
```
lpinca:
```suggestion
          });
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+
+      it('can be overridden using options.maxPayload', function(done) {
+        const maxPayload = 123;
+
+        const wss = new WebSocket.Server({ port: 0 }, () => {
+          const ws = new WebSocket(`ws://localhost:${wss.address().port}`, {
+            maxPayload
+          });
+
+          ws.onopen = () => {
```
lpinca:
```suggestion
          ws.on('open', () => {
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+            assert.strictEqual(ws.maxPayload, maxPayload);
+            wss.close(done);
+          };
```
lpinca:
```suggestion
          });
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+        });
+      });
+
+      it('can be modified after "open"', function(done) {
+        const maxPayload = 123;
+
```
lpinca:
```suggestion
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+          });
+
+          ws.onopen = () => {
+            assert.strictEqual(ws.maxPayload, maxPayload);
+            wss.close(done);
+          };
+        });
+      });
+
+      it('can be modified after "open"', function(done) {
```
lpinca:
```suggestion
      it("can be modified after the 'open' event is emitted", function(done) {
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+
+        const wss = new WebSocket.Server(
+          {
+            perMessageDeflate: true,
+            port: 0
+          },
+          () => {
+            const ws = new WebSocket(`ws://localhost:${wss.address().port}`);
+
+            ws.onopen = () => {
```
lpinca:
```suggestion
            ws.on('open', () => {
```

test/websocket.test.js
```diff
@@ -164,6 +164,59 @@ describe('WebSocket', function() {
...
+              ws.maxPayload = maxPayload;
+              assert.strictEqual(ws.maxPayload, maxPayload);
+              assert.strictEqual(ws._receiver._maxPayload, maxPayload);
+              assert.strictEqual(
+                ws._receiver._extensions['permessage-deflate']._maxPayload,
+                maxPayload
+              );
+              wss.close(done);
+            };
```
lpinca:
```suggestion
            });
```

lpinca:
Please add documentation.

lpinca:
> I don't see how it's a good thing to hide an attribute that exists, has a functional effect, and whose initialization is already part of the API.

Private attributes/methods exist in many languages for a reason. The with JS is that they are not that private.

mvduin:
> Nits

Will fix. I didn't like the onopen either, but I just copy-pasted it from the existing testcode below the one I inserted. Whenever I edit code that's not mine I default to using whatever style the code in the immediate surroundings uses, but I see now it just happened to be one of the small minority of test cases that still use onopen.

> Please also add documentation.

Will fix!

> Private attributes/methods exist in many languages for a reason.

I'm not disputing that, but maxPayload is already user-provided, and I don't see any good reason to keep it private after initialization.

lpinca:
@websockets/admin @3rd-Eden what do you think?

mvduin:
My previous version didn't allow maxPayload to be get/set on client sockets while connecting (i.e. before _receiver is set) because the setting was being held in a closure at that time. While I don't care about the ability to modify on client sockets at all, let alone when they're connecting, this exception just seemed too weird and unsightly to leave it be, especially as I tried to document it.

I've fixed this by keeping the setting on the WebSocket itself and syncing it to _receiver (and deflate extension, if applicable), similar to how binaryType is handled.

lib/websocket.js
```diff
@@ -457,6 +474,7 @@ function initAsClient(address, protocols, options) {
   }
 
   this._isServer = false;
+  this.maxPayload = options.maxPayload;
```
lpinca:
```suggestion
  this._maxPayload = options.maxPayload;
```

lib/websocket-server.js
```diff
@@ -192,8 +192,7 @@ class WebSocketServer extends EventEmitter {
     if (this.options.perMessageDeflate) {
       const perMessageDeflate = new PerMessageDeflate(
         this.options.perMessageDeflate,
-        true,
-        this.options.maxPayload
```
lpinca: Unnecessary change.

lib/websocket.js
```diff
@@ -47,6 +47,7 @@ class WebSocket extends EventEmitter {
     this._closeCode = 1006;
     this._extensions = {};
     this._isServer = true;
+    this._maxPayload = 0;
```
lpinca: I'd use the same default value of the `maxPayload` option.

lpinca:
@mvduin another thing I was thinking about is that changing the value dynamically may have surprising effects if that is done while the payload of a frame is being decompressed or in the middle of a fragmented message. I think it's like giving people a loaded gun to shoot themselves in the foot with, so if anything changing it dynamically should be discouraged.

lpinca:
Closing as per https://github.com/websockets/ws/pull/1473#issuecomment-441112214 and https://github.com/websockets/ws/pull/1473#issuecomment-441778987.

I think that changing the value of `maxPayload` dynamically is not something that ws should officially support/endorse. It's still possible to achieve this by using the private `_maxPayload` property on the `Receiver` and `PerMessageDeflate` instances if really needed.

I'm happy to reopen this if someone can make me change my mind.

PR #1474, opened 2018-11-25
[feature] socketTimeout option

gonapps:
When you connect to the websocket server and don't send any data and idle, the socket on the server keeps being opened for default timeout (12 minutes)
`$ nc localhost 8080`
I think for some people default 12 minutes timeout feels too long.
And this timeout seems enough to exhaust available sockets on the server if some malicious user commits DDOS attack.
I think it would be great if socketTimeout option is supported.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -20,6 +20,7 @@ This class represents a WebSocket server. It extends the `EventEmitter`.
   - `clientTracking` {Boolean} Specifies whether or not to track clients.
   - `perMessageDeflate` {Boolean|Object} Enable/disable permessage-deflate.
   - `maxPayload` {Number} The maximum allowed message size in bytes.
+  - `socketTimeout` {Number} Timeout in milliseconds for idle sockets.
 - `callback` {Function}
 
 Create a new server instance. One of `port`, `server` or `noServer` must be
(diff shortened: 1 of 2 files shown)


lpinca:
Thank you for the pull request. Default timeout is 2 minutes not 12.
Why not using an external server instead of adding a new option?

```js
const createServer = require('http');

wss.on('connection', function () {
  // Do stuff.
});

server.setTimeout(60000).listen(8080);
```

gonapps:
Yes you are right, It seems like somehow I misread the 120000 as a 12 minutes.
But I think still 2 minutes timeout is too long and there will be some people who agree with it.
And also I knew that it is possible to set timeout using external server object, but I think most of people will be satisfied with this default callback method: 
```javascript
this._server = http.createServer((req, res) => {
        const body = http.STATUS_CODES[426];

        res.writeHead(426, {
          'Content-Length': body.length,
          'Content-Type': 'text/plain'
        });
        res.end(body);
});
```
So instead of repeat writing 'Upgrade required' callback, I thought supporting timeout option is more convenient.
I also believe socketTimeout option is needed as frequently as port option.

lpinca:
To be honest I'm not a fan of adding a new option when the same can be achieved with literally 5 line of code.

lpinca:
@gonapps sorry but I'm going to close this.

gonapps:
Okay thanks

PR #1482, opened 2019-01-01
Added support to send JSON Objects

piyushkantm:
This reduces repetition for backend dealing with JSON objects and makes the code elegant as we do not have to repeatedly use JSON.stringify everywhere..

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,24 @@ ws.on('message', function incoming(data) {
 });
 ```
 
+### Sending JSON data
+
+```js
+const WebSocket = require('ws');
+
+const ws = new WebSocket('ws://www.host.com/path');
+
+ws.on('open', function open() {
+  ws.sendJSON({
+    "something": "creative",
+    "and": [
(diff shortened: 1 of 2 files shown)


piyushkantm:
@lpinca I am not sure if i need to cover this change with tests too. Please suggest

lpinca:
Thank you for the PR but this is not something that should be in ws.
It can easily be done outside of ws with a dead simple function.

```js
function sendJSON(ws, data, options, callback) {
  ws.send(JSON.stringify(data), options, callback);
}
```

piyushkantm:
Ok, i get it. Thanks for your prompt reply.

JayBeavers:
Disagree, nodejs lives & breathes JSON and there's easy ways to disambiguate an object from a string and auto-call JSON.stringify.

As a new user of this library, was quite surprised it didn't do this automatically when you sister libraries, like expressjs, handle this elegantly by default.

PR #1490, opened 2019-01-19
Add ability to follow redirects

### LINKED ISSUE #812: HTTP Redirection at Handshake
Hi there !

Like it say in [the RFC 6455](https://tools.ietf.org/html/rfc6455)

```
   1.  If the status code received from the server is not 101, the
       client handles the response per HTTP [RFC2616] procedures.  In
       particular, the client might perform authentication if it
       receives a 401 status code; the server might redirect the client
       using a 3xx status code (but clients are not required to follow
       them), etc.  Otherwise, proceed as follows.
```

But when i try a redirect status 302.
The websocket throw an Error  'unexpected server response (302)'

``` js
// serv.js
var Net = require('net')
var Wss = require('ws').Server

var server = new Net.Server()

server.on('connection',socket=>{
  socket.write('HTTP/1.1 302 Found\nLocation: ws://localhost:8080\n\n')
  socket.end()
})

server.listen(80)

var wss = new Wss({port:8080})
```

``` js
// ws.js
var Ws = require('ws')

var ws = new Ws('ws://localhost') // throw Error: unexpected server response (302)
```

I misunderstand something or this is a divergence of the RFC 6455 ?

lpinca:
Known defects:

- The options object it rebuilt after every redirect. This could be optimized to only do it once and update only the options that need to be changed after every redirect.
- There might be issues if an agent is used and the protocol is changed from HTTPS to HTTP or vice versa since HTTPS and HTTP use different agents.

Fixes #812

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -46,12 +46,14 @@ class WebSocket extends EventEmitter {
     this._closeTimer = null;
     this._closeCode = 1006;
     this._extensions = {};
-    this._isServer = true;
     this._receiver = null;
     this._sender = null;
     this._socket = null;
 
     if (address !== null) {
+      this._isServer = false;
+      this._redirects = 0;
+
       if (Array.isArray(protocols)) {
         protocols = protocols.join(', ');
       } else if (typeof protocols === 'object' && protocols !== null) {
@@ -59,7 +61,9 @@ class WebSocket extends EventEmitter {
         protocols = undefined;
       }
 
-      initAsClient.call(this, address, protocols, options);
+      initAsClient(this, address, protocols, options);
+    } else {
+      this._isServer = true;
     }
   }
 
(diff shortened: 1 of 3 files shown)


piranna:
> * There might be issues if an agent is used and the protocol is changed from HTTPS to HTTP or vice versa since HTTPS and HTTP use different agents.

I would throw an error if changuing from HTTPS to HTTP to notify about security issues by default, and maybe add an option to disable it.

lpinca:
We can but afaik none of the most popular HTTP libs (request, got, node-fetch, axios, etc.) does it. Accepting redirects has a lot of security implications, like open redirect attacks, or passing authorization headers when the hostname changes.

It should be enabled only the server is trusted.

piranna:
> It should be enabled only when the server is trusted.

So do you think just lefting redirections as an opt-in is just enough?

lpinca:
I don't know if it is enough, but it is the reason why I chose to not follow redirects by default.

piranna:
Great! Thank you :-)

lib/websocket.js
```diff
@@ -449,136 +462,159 @@ function initAsClient(address, protocols, options) {
...
+      }
+
+      req.abort();
+
+      const addr = url.URL
+        ? new url.URL(location, address)
+        : url.resolve(address, location);
 
-    abortHandshake(this, req, `Unexpected server response: ${res.statusCode}`);
+      initAsClient(websocket, addr, protocols, options);
```
ad-m: I think the redirect address should be emitted in some way to allow tracking.

PR #1494, opened 2019-01-23
Handle close frame which spans multiple packets

### LINKED ISSUE #1493: Close frame spanning multiple packets is not handled correctly
      issue.

#### Description
A close frame which has been split over multiple TCP packets, for example <header><payload> is not handled correctly.  The payload is never read and `conclude` event not emitted.

#### Reproducible in:

- version: 6.1.2
- Node.js version(s): 10.12
- OS version(s): Linux; Fedora 29

#### Steps to reproduce:

1. Establish a websocket session

2. Send a close frame from the server as two separate packets

#### Expected result:

1. Close frame is parsed correctly and `conclude` event is emitted from receiver

#### Actual result:

1. Close frame is not parsed and `conclude` is never emitted

mwm-twx:
Fixes #1493

### DIFF
--- a/lib/receiver.js
+++ b/lib/receiver.js
@@ -62,7 +62,7 @@ class Receiver extends stream.Writable {
    * @param {Function} cb Callback
    */
   _write(chunk, encoding, cb) {
-    if (this._opcode === 0x08) return cb();
+    if (this._opcode === 0x08 && this._state == GET_INFO) return cb();
 
     this._bufferedBytes += chunk.length;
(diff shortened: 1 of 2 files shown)


lpinca:
I will cut a new release as soon as I can.

mwm-twx:
Excellent.  Thanks for the quick response.

bradennapier:
Yep, just ran into this one :-D . Good to see already resolved!  Updating :-)

PR #1514, opened 2019-02-22
if the cb exist, and when we wrapping ws with Promise, then

lygstate:
the err would trigger twice. on from cb, and the other one
comes from throw


lpinca:
The `return` statement prevent the error from being thrown.

lygstate:
I am wrong, sorry

PR #1539, opened 2019-03-26
Add summary with links to the doc

ttous:
Thanks to The Nacho Man and his [nGitHubTOC](https://imthenachoman.github.io/nGitHubTOC/).

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -1,5 +1,52 @@
 # ws
 
+## Summary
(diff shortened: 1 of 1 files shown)


lpinca:
The structure of the documentation is similar to one used by Node.js. For example see https://nodejs.org/api/http.html. Can we use the same structure also for the ToC? Thank you.

lpinca:
Thank you.

PR #1581, opened 2019-06-08
docs: Add missing dependency in code snippet

mgolebiowski:
Hey!
In the "Multiple servers sharing a single HTTP/S server" usage example URL module is used but it's not included in the snippet. This PR is adding 
```javascript
const url = require('url');
```
line to make the snippet valid.


lpinca:
Thank you.

Issue #1583, opened 2019-06-12
WebSocket objects not assignable to TypeScript WebSocket type because they do not implement the `dispatchEvent` method


brainkim:
- [x] I've searched for any related issues and avoided creating a duplicate
      issue.

#### Description

The WebSocket objects which ws uses are not assignable to the typescript WebSocket type because the `dispatchEvent` method is missing from the implementation.

#### Reproducible in:

- version: 7.0.0
- Node.js version: 10.16.0
- TypeScript version: 3.4.5
- OS version(s): MacOS 10.14.4

#### Steps to reproduce:

1. Create a ws server

2. Listen for the connection event

3. Pass the websocket passed to the connection handler to a typescript function which expects a WebSocket object.

#### Expected result:

No typechecking errors occur
#### Actual result:

```
    src/connection/__tests__/socket.ts:104:33 - error TS2345: Argument of type 'WebSocket' is not assignable to parameter of type 'Socket'.
      Property 'dispatchEvent' is missing in type 'WebSocket' but required in type 'WebSocket'.

    104         resolve(new SocketProxy(socket, storage));
                                        ~~~~~~

      ../../node_modules/typescript/lib/lib.dom.d.ts:5226:5
        5226     dispatchEvent(event: Event): boolean;
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        'dispatchEvent' is declared here.
```

lpinca:
There is no TypeScript on this project and there is no `dispatchEvent()` because we don't fully implement the `EventTarget` interface.

brainkim:
> There is no TypeScript on this project
Understood! I’m using the types from https://www.npmjs.com/package/@types/ws

> there is no dispatchEvent() because we don't fully implement the EventTarget interface.
A stub method which throws a not implemented error would be enough to satisfy the type definition.

lpinca:
> A stub method which throws a not implemented error would be enough to satisfy the type definition.

Please file a PR if you are interested in having it.

Seairth:
I encountered this issue as well.  In my case, I have an interface that contains:
```Typescript
ws: WebSocket;
```
By default, TS assumes you mean the definition in _lib.dom.d.ts_.  Since the only `WebSocket` I cared about was the class declaration defined in the **@types/ws** package, I added the following at the top of the file containing my interface:
```Typescript
import WebSocket from 'ws';
```
[...]

PR #1584, opened 2019-06-12
add stub dispatchEvent method to WebSocket (#1583)

brainkim:
- [ ] Update `@types/ws` to include the dispatchEvent method

### DIFF
--- a/lib/event-target.js
+++ b/lib/event-target.js
@@ -164,6 +164,15 @@ const EventTarget = {
         this.removeListener(method, listeners[i]);
       }
     }
+  },
+
+  /**
+   * Dispatches event. This method will always throw an error.
+   *
+   * @public
+   */
+  dispatchEvent() {
+    throw new Error('The dispatchEvent() method is not implemented');
   }
 };
 
--- a/test/websocket.test.js
+++ b/test/websocket.test.js
@@ -1774,6 +1774,13 @@ describe('WebSocket', () => {
       assert.strictEqual(ws.listenerCount('foo'), 0);
     });
 
+    it('has a stub `dispatchEvent` method', () => {
+      const ws = new WebSocket('ws://localhost', { lookup() {} });
+      assert.throws(() => {
+        ws.dispatchEvent({ type: 'close' });
+      }, /^Error: The dispatchEvent\(\) method is not implemented$/);
+    });
+
     it('wraps text data in a `MessageEvent`', (done) => {
       const wss = new WebSocket.Server({ port: 0 }, () => {
         const ws = new WebSocket(`ws://localhost:${wss.address().port}`);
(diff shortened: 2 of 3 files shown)


3rd-Eden:
You probably want to include a test to ensure that they are not removed in future refactors

lib/event-target.js
```diff
@@ -164,6 +164,15 @@ const EventTarget = {
...
     }
+  },
+
+  /**
+   * Dispatches event. This method will always throw an error.
+   *
+   * @public
+   */
+  dispatchEvent() {
+    throw new Error('Not Implemented');
```
lpinca:
```suggestion
    throw new Error('The dispatchEvent() method is not implemented');
```

brainkim:
Added test and updated error message

test/websocket.test.js
```diff
@@ -1774,6 +1774,21 @@ describe('WebSocket', () => {
       assert.strictEqual(ws.listenerCount('foo'), 0);
     });
 
+    it('has a stub `dispatchEvent` method', (done) => {
+      const wss = new WebSocket.Server({ port: 0 }, () => {
```
lpinca: We can simplify this test, there is no need to establish a connection:

```js
const ws = new WebSocket('ws://localhost', {
  lookup() {}
});

assert.throws(
  ws.dispatchEvent(),
  /^Error: The dispatchEvent\(\) method is not implemented$/
)
```

brainkim:
tests are hanging one sec

test/websocket.test.js
brainkim: The `WebSocket` in this scope is a WebSocket from the jsdom lib. I think this test would fail?

brainkim: Never mind, this works. What I found earlier was that passing a fake event would cause another test to fail for some reason.

lpinca:
Thank you.

brainkim:
Thanks for the quick turnaround! May I leave the related issue open until I get a @types/ws definitions updated?

brainkim:
Never mind. I’m looking into the typings now and there are way more incompatibilities between the ws WebSocket type and the typescript lib WebSocket type 😓 including but not limited to the `Event` object lacking properties like `Bubble`, etc. Probably just gonna continue using type assertions at the end of the day.

lpinca:
Hmm ok then I think it's better to revert this as well.

lpinca:
15 min passed so I guess it's ok to force push. On my way.

brainkim:
Yeah if supporting the typescript websocket definition is desired, then I think supporting the Event/EventTarget interface more fully is required and I’m not sure if that’s what this project wants. Sorry for this.

lpinca:
No problem.

PR #1587, opened 2019-06-14
Setting the servername to an empty string '' should disable sending the SNI extension.

holgerkoser:
The option `servername` in [`tls.connect(options[, callback])`](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback) is the name of the host being connected to, and must be a host name, and not an IP address. If the server is not a multi-homed server and I want to connect with the IP address it is not possible to disable sending the SNI extension by setting the servername to empty string `''`. This PR implements the defaulting of the `servername` in the same way as it is done in the [node](https://github.com/nodejs/node/blob/master/lib/_http_agent.js#L161-L162). 

This PR is related to #1347 .

**Example**
In node v12 with the following code
```js
new WebSocket('wss://127.0.0.1', {
  ca: fs.readFileSync('ssl/ca.pem'),
  servername: '',
})
```
I get the deprecation warning [`[DEP0123]`](https://nodejs.org/api/deprecations.html#deprecations_dep0123_setting_the_tls_servername_to_an_ip_address)
```bash
(node:71346) [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address 
is not permitted by RFC 6066. This will be ignored in a future version.
```

### DIFF
--- a/test/websocket.test.js
+++ b/test/websocket.test.js
@@ -6,6 +6,7 @@ const assert = require('assert');
 const crypto = require('crypto');
 const https = require('https');
 const http = require('http');
+const tls = require('tls');
 const fs = require('fs');
 const { URL } = require('url');
 
@@ -2040,6 +2041,29 @@ describe('WebSocket', () => {
         });
       });
     }).timeout(4000);
+
+    describe('tls connect options', () => {
+      const tlsConnect = tls.connect;
+      let tlsConnectOptions;
+
+      beforeEach(() => {
+        tlsConnectOptions = {};
+        tls.connect = (options) => {
+          Object.assign(tlsConnectOptions, options);
+        };
+      });
+
+      afterEach(() => {
+        tls.connect = tlsConnect;
+      });
+
+      it('allows to use empty string to disable sending the SNI extension', () => {
+        const ws = new WebSocket('wss://127.0.0.1', {
+          servername: ''
+        });
+        assert.strictEqual(tlsConnectOptions.servername, '');
+      });
+    });
   });
 
   describe('Request headers', () => {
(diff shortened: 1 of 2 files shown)


lpinca:
Can we add a test?

holgerkoser:
Yes, of course. I already tried to add a test in the [`SSL`](https://github.com/websockets/ws/blob/master/test/websocket.test.js#L1922) but I not found an easy way to intercept the `tls.connect` call. I normally use [`proxyquire`](https://github.com/thlorenz/proxyquire) for this but don't wanted to add a new dependency for such a small change. I try to find an alternative simple solution for a test.

lpinca:
I think you can "copy" the test included in https://github.com/nodejs/node/commit/98e9de7db930e505884ca8d0ca6588afbe43f127.

holgerkoser:
Thanks for the hint. I have seen it too late and already added a test. If you think it is better to "copy" the kind of test from node I will do that.

test/websocket.test.js
```diff
@@ -2040,6 +2041,29 @@ describe('WebSocket', () => {
...
+          Object.assign(tlsConnectOptions, options);
+        };
+      });
+
+      afterEach(() => {
+        tls.connect = tlsConnect;
+      });
+
+      it('allows to use empty string to disable sending the SNI extension', () => {
+        const ws = new WebSocket('wss://127.0.0.1', {
```
lpinca: Can you please mock `tls.connect()` here and remove the `describe` wrapper and the `beforeEach()` and `afterEach()` hook? We will add them if/when needed. Thank you.

lpinca:
I prefer your test.

lpinca:
Thank you.

holgerkoser:
@lpinca Couldn't answer any more, because I had two days off and since today I'm back in office again. Many thanks for your fast reply and your help.

PR #1589, opened 2019-06-16
Add utility to wrap a `WebSocket` in a `Duplex` stream

lpinca:
I'm not sure if it make sense to add this but reasons for doing it are:

- No additional dependencies.
- Functionality is strictly bound to the ws internals.
- Already existing modules are broken. The most popular, [websocket-stream](https://github.com/maxogden/websocket-stream)
  - [Loses data](https://github.com/maxogden/websocket-stream/issues/139)
  - [Does not properly handle backpressure](https://gist.github.com/lpinca/114a348f6108acc1e4236fed05522f73)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ can use one of the many wrappers available on npm, like
   - [Multiple servers sharing a single HTTP/S server](#multiple-servers-sharing-a-single-https-server)
   - [Server broadcast](#server-broadcast)
   - [echo.websocket.org demo](#echowebsocketorg-demo)
+  - [Use the Node.js streams API](#use-the-nodejs-streams-api)
   - [Other examples](#other-examples)
 - [FAQ](#faq)
   - [How to get the IP address of the client?](#how-to-get-the-ip-address-of-the-client)
@@ -69,7 +70,8 @@ necessarily need to have a C++ compiler installed on your machine.
 
 ## API docs
 
-See [`/doc/ws.md`](./doc/ws.md) for Node.js-like docs for the ws classes.
+See [`/doc/ws.md`](./doc/ws.md) for Node.js-like documentation of ws classes and
+utility functions.
 
 ## WebSocket compression
 
@@ -300,6 +302,21 @@ ws.on('message', function incoming(data) {
     ws.send(Date.now());
   }, 500);
 });
+```
+
+### Use the Node.js streams API
+
+```js
+const WebSocket = require('ws');
+
+const ws = new WebSocket('wss://echo.websocket.org/', {
+  origin: 'https://websocket.org'
+});
+
+const duplex = WebSocket.createWebSocketStream(ws, { encoding: 'utf8' });
+
+duplex.pipe(process.stdout);
+process.stdin.pipe(duplex);
 ```
 
 ### Other examples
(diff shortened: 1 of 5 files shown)


stream.js
```diff
@@ -0,0 +1,150 @@
...
+
+  duplex._write = function(chunk, encoding, callback) {
+    if (ws.readyState === ws.CONNECTING) {
+      ws.once('open', function open() {
+        duplex._write(chunk, encoding, callback);
+      });
+      return;
+    }
+
+    ws.send(chunk, callback);
```
lpinca: When `readyState` is `CLOSING` or `CLOSED` and error is emitted on the duplex stream. This can be annoying when using `readable.pipe()` as the other peer can send a close frame at any time but I think it is the right thing to do. What do you think?

3rd-Eden: Do we want to wrap this write in a try catch statement to ensure that nothing is thrown when we attempt to write?

README.md
```diff
@@ -302,6 +304,22 @@ ws.on('message', function incoming(data) {
 });
 ```
 
+### Use the Node.js streams API
+
+```js
+const WebSocket = require('ws');
+const createWebSocketStream = require('ws/stream');
```
3rd-Eden: What was the reasoning behind putting it in a different file instead of as a separate export? e.g.

```js
const { createWebSocketStream } = require('ws')
```
```

File: stream.js (hunk 1 above)
Comment (lpinca) [reply stream.js]: As of 5d751fbd4c0ab3478a6de4194d4d06908bc8ac00 if no longer throw synchronously. Or well it does but only when `readyState` is `CONNECTING` and in that case this is unreachable.

The callback is called with an error when `readyState` is `CLOSING` or `CLOSED`.

File: README.md (hunk 2 above)
Comment (lpinca) [reply README.md]: I did not want to pollute the `WebSocket` class with spurious attributes but we already do that for `Server`, `Receiver`, and `Sender` so I'm fine exposing it as a property of the `WebSocket` class if that is preferred.

File: stream.js (hunk 1 above)
Comment (lpinca) [reply stream.js]: Actually it didn't event before https://github.com/websockets/ws/commit/5d751fbd4c0ab3478a6de4194d4d06908bc8ac00 when a callback was used.

File: README.md (hunk 2 above)
Comment (3rd-Eden) [reply README.md]: Yep, thats why I wondered if there was reason why this was exposed as separate file instead of added to the exports like the rest. For the sake of consistency it makes the most sense to add it as export.

Comment (lpinca) [reply README.md]: Ok, to be honest I would like to remove both `Receiver` and `Sender` from the `WebSocket` class and tell people to require `ws/lib/receiver` if they need to. I think Removing `Server` is too disruptive.

File: stream.js (hunk 1 above)
Comment (3rd-Eden) [reply stream.js]: Right, good point, not sure how I missed that CONNECTING was already above.

File: README.md (hunk 2 above)
Comment (3rd-Eden) [reply README.md]: I agree with that sentiment as I consider both of those classes internals. But that would however be a breaking change. But for things like `Server` I do think it makes sense to expose it as a named export instead of a separate file as it would be a public API endpoint.

Comment (lpinca) [reply README.md]: Agreed.

**Comment by lpinca** [comment]
@3rd-Eden PTAL.

**Comment by chainhead** [comment]
How do we read the stream _into_ `ws.send()`?

This option keeps hitting the event emitter limit - not sure why.


```js
duplex.on('data', (chunk) => {
  ws.send(chunk.toString())
}
```

PR #1590, opened 2019-06-17
Fix Server Broadcast example

amitnovick:
Tried to use the *Server Broadcast* example but it didn't work.

Removing some parts made it work.


README.md
```diff
@@ -254,20 +254,11 @@ const WebSocket = require('ws');
...
-    }
-  });
-};
-
 wss.on('connection', function connection(ws) {
   ws.on('message', function incoming(data) {
     // Broadcast to everyone else.
     wss.clients.forEach(function each(client) {
-      if (client !== ws && client.readyState === WebSocket.OPEN) {
+      if (client.readyState === WebSocket.OPEN) {
```
lpinca: This is not ok as it now broadcasts to ALL clients including the sender of the message.

lpinca:
The example shows two different scenarios.

- Broadcast to all.
- Broadcast a message received from a client to all other clients (this is whan happen when the example is run).

Sorry but this change removes the original purpose of the example.

README.md
amitnovick: Oh I see.
I couldn't figure it out from the code having looked at it for the first time.
In this context `client` is actually a client's web socket object and that wasn't obvious to me.

Got no obvious suggestion to make it more readable I'm afraid. :confused:

amitnovick:
> The example shows two different scenarios:
> 
>     * Broadcast to all.
> 
>     * Broadcast a message received from a client to all other clients (this is whan happen when the example is run).
> 
> 
> Sorry but this change removes the original purpose of the example.
Thanks so much for replying! 

I Couldn't find a documentation on the `broadcast` method of the WebSocket.Server object.

Not sure why it's needed in this example:
- The `wss.on('message', ...)` part is already broadcasting messages sent from clients to all other clients
- The wss.broadcast` sends messages to clients at **which phase? i.e. which event triggers it?**

Kind regards

lpinca:
>I Couldn't find a documentation on the broadcast method of the `WebSocket.Server` object.

Yes, because there is no official broadcast method. It is a convenience method defined in the example itself.

> The `wss.broadcast` sends messages to clients at which phase? i.e. which event triggers it?

It is not used. it is just there to show the difference.

Perhaps we can split the example in two. Once that broadcast to all and another that broadcast to all but the sender.

amitnovick:
> > I Couldn't find a documentation on the broadcast method of the `WebSocket.Server` object.
> 
> Yes, because there is no official broadcast method. It is a convenience method defined in the example itself.

Oh, yea I see what the intent was now.
That said, coming from React ecosystem, where functional programming style is the norm, this mutability just seems like a code smell by now, and my assumption was that the underlying object would detect the mutation and do something with the new property.

> Perhaps we can split the example in two. Once that broadcast to all and another that broadcast to all but the sender.

That seems like a good suggestion! went ahead and did just that. :+1: 

Please let me know if something seems off about this to you.

README.md
```diff
@@ -249,23 +251,35 @@ server.listen(8080);
 
 ### Server broadcast
 
+1. A client web socket broadcasting to all connected web socket clients
```
lpinca:
```suggestion
A client WebSocket broadcasting to all connected WebSocket clients, including itself.
```

README.md
```diff
@@ -249,23 +251,35 @@ server.listen(8080);
...
+      if (client.readyState === WebSocket.OPEN) {
+        client.send(data);
+      }
+    });
   });
-};
+});
+```
+
+2. A client web socket broadcasting to every other connected web socket,
```
lpinca:
```suggestion
A client WebSocket broadcasting to every other connected WebSocket clients, excluding itself.
```

README.md
```diff
@@ -21,26 +21,28 @@ can use one of the many wrappers available on npm, like
...
-  - [echo.websocket.org demo](#echowebsocketorg-demo)
-  - [Other examples](#other-examples)
-- [FAQ](#faq)
-  - [How to get the IP address of the client?](#how-to-get-the-ip-address-of-the-client)
-  - [How to detect and close broken connections?](#how-to-detect-and-close-broken-connections)
-  - [How to connect via a proxy?](#how-to-connect-via-a-proxy)
-- [Changelog](#changelog)
-- [License](#license)
+- [ws: a Node.js WebSocket library](#ws-a-nodejs-websocket-library)
+  - [Table of Contents](#table-of-contents)
```
lpinca: I'm not sure if it makes sense to add the table of contents itself in the table of contents. Anyway no strong opinion.

amitnovick: > I'm not sure if it makes sense to add the table of contents itself in the table of contents. Anyway no strong opinion.

Apologies for my blunder! :sob: 
Had a run-in with my formatter applying that automatically without me noticing.

I believe it's restored now after the commit I made.

Thanks for catching that! :+1:

lpinca: Seems good can you please apply the two suggested changes above? The numbered list is not rendered correctly anyway. Other than that LGTM.

amitnovick: > Seems good can you please apply the two suggested changes above? The numbered list is not rendered correctly anyway. Other than that LGTM.

Should be good now.
The suggestion violated a code style rule, which I addressed

lpinca:
Thank you.

PR #1600, opened 2019-06-27
Ensure that concat() never returns uninitialized data

ChALkeR:
_There are no known to me occurrences of returning uninitialized data, I am adding this as a safeguard._

---

This is just a safeguard that ensures that for cases if totalLength is by some mistake gets miscalculated, no uninitialized memory would be leaked.

This should not introduce any performance impact as in a normal case (which should be all the cases) this is just a single comparison.

### DIFF
--- a/lib/buffer-util.js
+++ b/lib/buffer-util.js
@@ -23,6 +23,8 @@ function concat(list, totalLength) {
     offset += buf.length;
   }
 
+  if (offset < totalLength) return target.slice(0, offset);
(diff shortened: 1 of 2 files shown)


lpinca:
Tests should already cover the bad case and catch the bug in the implementation if exists as the returned buffer length would be bigger than the expected one. Anyway I'm fine with the additional safety guard. Can you add a test for it? Thank you.

Issue #1564, opened 2019-05-12
 Cannot set property 'readyState' of undefined


mccoysc:
websocket.readyState = WebSocket.CLOSING;
                       ^

TypeError: Cannot set property 'readyState' of undefined
    at TLSSocket.socketOnClose (/node_modules/ws/lib/websocket.js:826:24)
    at TLSSocket.emit (events.js:187:15)
    at _handle.close (net.js:606:12)
    at Socket.done (_tls_wrap.js:386:7)
    at Object.onceWrapper (events.js:273:13)
    at Socket.emit (events.js:182:13)
    at TCP._handle.close (net.js:606:12)

mccoysc:
@lpinca i found that as a server:
if 2 or more clients(some app run in browser) are connected to this server,and if one client closed,the error would be occoured in server side.

mccoysc:
https://careeron.ecspool.com
you can browse this url in 2 browser tab.
refresh one tab,the another tab's websocket connection would be in "CLOSED" state(in fact,the server side is errored at this moment)

lpinca:
I can't see how this can happen:

1. The [`socketOnClose()`](https://github.com/websockets/ws/blob/6.2.1/lib/websocket.js#L166) listener is added by [`WebSocket#setSocket()`](https://github.com/websockets/ws/blob/6.2.1/lib/websocket.js#L140-L173) after [`socket[kWebSocket] = this;`](https://github.com/websockets/ws/blob/6.2.1/lib/websocket.js#L150)
1. How can [`websocket`](https://github.com/websockets/ws/blob/6.2.1/lib/websocket.js#L826) be `undefined` when the listener is called?

Please create a minimal test case, using only `ws`, to reproduce the issue.

mccoysc:
i found that for one same websocket,the socketOnClose function is called twice:for the first,you have make "this[kWebSocket]=undefined",and so the seconed,this[kWebSocket] is undefined

mccoysc:
@lpinca

lpinca:
If the `'close'` event is emitted more than once on the `net.Socket` or `tls.Socket` then there is a bug in Node.js.

Also the listener removes itself after the it is called. See https://github.com/websockets/ws/blob/6.2.1/lib/websocket.js#L823

mccoysc:
but how can i resolve this problem?

lpinca:
[...]

PR #1606, opened 2019-07-12
if this[kWebSocket] = undefined , then must check it exists else where

amark:
`ws` is fantastic, and devs/users of my Open Source library have been using my library with `ws` in it for years in production, they currently have like ~14M monthly active users! None of these devs/users are paying me commercially though, so this patch is on my own OSS time for their sake, but I've been very thankful that `ws` is also OSS like my library is, it is very much appreciated.

I'm issuing this patch because I cannot catch these errors that happen when I (1) run a websocket server on localhost:8765, (2) connect browser to websocket server (3) connect websocket server to a non-existent localhost:8080 & retry every 2 seconds (4) refresh browser & reconnect to localhost:8765 . Sometimes I may need a `dgram` socket running in the same instance, but it also crashes when I disable that (though maybe flooding the router with a bunch of UDP packets may be necessary to replicate).

```
node_modules/ws/lib/websocket.js:837
  websocket.readyState = WebSocket.CLOSING;
                       ^
TypeError: Cannot set property 'readyState' of undefined
    at Socket.socketOnClose
```

Either way, setting `this[kWebSocket] = undefined` seems dangerous (when most of the listeners start with a `this[kWebSocket]`) unless it is **impossible** for the listeners to be called again (which clearly is not the case, since I'm getting a crash where `this[kWebSocket]` is `undefined` which can only be possible if `socketOnClose` was called (and clearly `this.removeListener('close', socketOnClose);` is not working correctly).

I suspect that somehow the listeners are still active, so that should separately be fixed (but I'm not sure how), but the `if(!websocket){ return }` are safeguards. If the rest of the code works as intended, these lines will not break anything. If the code does not work as intended, these lines at least **prevent** an un-catch-able crash on production servers, which is really really important for me - so I hope this can be pulled & published as the solution, or pulled & published until another fix comes along.

Thanks again!

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -743,6 +743,7 @@ function sendAfterClose(websocket, data, cb) {
  */
 function receiverOnConclude(code, reason) {
   const websocket = this[kWebSocket];
+  if(!websocket){ return }
 
   websocket._socket.removeListener('data', socketOnData);
   websocket._socket.resume();
@@ -761,7 +762,9 @@ function receiverOnConclude(code, reason) {
  * @private
  */
 function receiverOnDrain() {
-  this[kWebSocket]._socket.resume();
+  const websocket = this[kWebSocket];
+  if(!websocket){ return }
+  websocket._socket.resume();
 }
 
 /**
@@ -772,6 +775,7 @@ function receiverOnDrain() {
  */
 function receiverOnError(err) {
   const websocket = this[kWebSocket];
+  if(!websocket){ return }
 
   websocket._socket.removeListener('data', socketOnData);
 
@@ -787,7 +791,9 @@ function receiverOnError(err) {
  * @private
  */
 function receiverOnFinish() {
-  this[kWebSocket].emitClose();
+  const websocket = this[kWebSocket];
+  if(!websocket){ return }
+  websocket.emitClose();
 }
 
 /**
@@ -797,7 +803,9 @@ function receiverOnFinish() {
  * @private
  */
 function receiverOnMessage(data) {
-  this[kWebSocket].emit('message', data);
+  const websocket = this[kWebSocket];
+  if(!websocket){ return }
+  websocket.emit('message', data);
 }
 
 /**
@@ -808,6 +816,7 @@ function receiverOnMessage(data) {
  */
 function receiverOnPing(data) {
   const websocket = this[kWebSocket];
+  if(!websocket){ return }
 
   websocket.pong(data, !websocket._isServer, NOOP);
   websocket.emit('ping', data);
@@ -820,7 +829,9 @@ function receiverOnPing(data) {
  * @private
  */
 function receiverOnPong(data) {
-  this[kWebSocket].emit('pong', data);
+  const websocket = this[kWebSocket];
+  if(!websocket){ return }
+  websocket.emit('pong', data);
 }
 
 /**
@@ -834,6 +845,7 @@ function socketOnClose() {
   this.removeListener('close', socketOnClose);
   this.removeListener('end', socketOnEnd);
 
+  if(!websocket){ return }
   websocket.readyState = WebSocket.CLOSING;
 
   //
@@ -850,7 +862,7 @@ function socketOnClose() {
   websocket._receiver.end();
 
   this.removeListener('data', socketOnData);
-  this[kWebSocket] = undefined;
+  this[kWebSocket] = undefined; // if we clear this out, we must check it exists else where.
 
   clearTimeout(websocket._closeTimer);
 
@@ -872,7 +884,9 @@ function socketOnClose() {
  * @private
  */
 function socketOnData(chunk) {
-  if (!this[kWebSocket]._receiver.write(chunk)) {
+  const websocket = this[kWebSocket];
+  if(!websocket){ return }
+  if (!websocket._receiver.write(chunk)) {
     this.pause();
   }
 }
@@ -884,6 +898,7 @@ function socketOnData(chunk) {
  */
 function socketOnEnd() {
   const websocket = this[kWebSocket];
+  if(!websocket){ return }
 
   websocket.readyState = WebSocket.CLOSING;
   websocket._receiver.end();


amark:
It does not seem to like the code format - I leave it to you to run whatever prettifier you want on the code/patch (I have no clue what configs you have. Why doesn't it just re-format automatically? Seems silly to leave it up to clumsy humans).

lpinca:
I'm not going to merge this without a way to reproduce the issue. This is masking a possible but, not fixing it.

amark:
I tried to explain in my post why this isn't a mask, would you be willing to re-read it again? Thanks!

It is not a matter of a mask, the problem is doing a `this[kWebSocket] = undefined` in code that frequently uses `this[kWebSocket]` without checking if it exists is dangerous. @lpinca for instance, would you consider this code to be buggy?

```
var object = {foo: {bar: 1}};
setTimeout(function(){
  object.foo = undefined;
}, Math.random() * 1000);

setTimeout(function(){
  var a = object.foo.bar * 2;
}, Math.random() * 1000);
```

Figuring out why `removeListener` seems to not be working should be handled by some other unit test (that is what the bug is, and my code is not masking that or preventing it). My code would not mask any actual unit test for this. All it does is provide safety **type/existence** checks in case anything else fails or a race condition happens.

To reproduce, please try steps (1) ~ (4).

Would you be willing to put yourself into my shoes? My OSS user's are using this in production with millions of users (I'm not getting paid for this, and I assume you aren't either, so you can at least empathize?), and they need a fix that will work now/ASAP - I am doing this for them, I took the time to handle it, and I'm certainly not an expert in your code - I think these checks should exist even if the other unit tests pass, but I'm totally fine if you delete my code after that, but until then, this is crashing stuff and it currently cannot be caught by me on my end.

Or at least could you make this catch-able in the meanwhile? So it won't crash servers?

Thanks so much.

lpinca:
> It is not a matter of a mask, the problem is doing a this[kWebSocket] = undefined in code that frequently uses this[kWebSocket] without checking if it exists is dangerous.

It is done by design. None of the listeners for which you added the guard should be called after `this[kWebSocket] = undefined` is set. If they do it's a bug and this patch is masking it.

The `'close'` event on the `{net,tls}.Socket` should not be emitted more than once. Furthermore we remove the listener itself after it is called. So the real question is why the `socketOnClose` is called more than once? What is causing it?

> Would you be willing to put yourself into my shoes? My OSS user's are using this in production with millions of users (I'm not getting paid for this, and I assume you aren't either, so you can at least empathize?), and they need a fix that will work now/ASAP

I'm also not payed to maintain ws, and this is completely irrelevant. gun.js is only one of the thousands of projects that use ws.

> I think these checks should exist even if the other unit tests pass

No, read above.

amark:
> None of the listeners for which you added the guard should be called 

@lpinca , yet it is. Isn't a bug defined as "something that happens when it should not"?

 > thousands of projects that use ws.

Right, this benefits all of them.

 > I'm also not payed to maintain ws, and this is completely irrelevant.

Exactly, which is why I made the patch for you!

I don't expect you have the time to work on this bug.

I don't have time to figure out why `removeListener` fails  (which again: my code doesn't mask this).

Now this is just a lose-lose-lose situation. You have an issue left open, I + others are stuck with a server that crashes, and who else is gonna fix except you or me (but my fix is not allowed)?

Sigh. I tried. Rather than just complaining. Have a good weekend. Let me know if/when/ever you'll think this'll get fixed. Bye.

lpinca:
> Let me know if/when/ever you'll think this'll get fixed.

Probably never if the underlying issue is not identified and explained. I don't want to repeat myself by in my opinion this is a blind fix that does not actually fixes the underlying issue. It masks it.

We received only one bug report with the same stack trace and that user was using guess what? gun.js https://github.com/websockets/ws/issues/1564. I have no clue why the issue happens only with gun.js.

If you really need this patch you can monkey patch ws in gun.js or temporarily fork ws and use it until a proper fix is found.

I can't merge this, sorry.

amark:
Did you see @nooop3 's comment https://github.com/websockets/ws/issues/1564#issuecomment-506588851 on that thread?

lpinca:
Yes, but I don't understand it? What is its meaning?

nooop3:
prevent call ws close multi times may be helpful.

jadbox:
@amark @lpinca
Maybe the reason why it seems that removeListener is not working is because the same listener is actually being added multiple times (which is allowable by node)?

"removeListener will remove, at most, one instance of a listener from the listener array. **If any single listener has been added multiple times to the listener array for the specified eventName, then removeListener must be called multiple times to remove each instance**"

lpinca:
It is not added multiple times, at least not by ws.

https://github.com/websockets/ws/blob/dd42c8bc3ac215668bbfe1788b3dab197219df2a/lib/websocket.js#L166

jadbox:
Is it possible that `req.on('upgrade', (res, socket, head) => {` is being triggered multiple times, which calls setSocket which adds the listeners? I'm not sure the flow here- when does the 'upgrade' event happen on `req` object and should it recall setSocket every time?

amark:
I do see a double emit on connection, but I'm guessing that piece is my fault - however that should cause it to register as separate sockets (in which case the close event is only fired once), but it seems like they're on the same socket, that is causing OnClose to be called a 2nd time.

The fact it is coded to support multiple listeners is cool, but assuming the opposite, that OnClose won't be called multiple times, seems like a bug waiting to happen.

lpinca:
> Is it possible that req.on('upgrade', (res, socket, head) => { is being triggered multiple times, which calls setSocket which adds the listeners? I'm not sure the flow here- when does the 'upgrade' event happen on req object and should it recall setSocket every time?

The `'upgrade'` event is emitted only once. It is emitted when the `Upgrade: websocket` and `Connection: Upgrade` are received. It requires a whole HTTP request/response roundtrip.

lpinca:
`'connection'` is not emitted multiple times for the same socket. There is a 1-1 correspondence between `'connection'` events and `WebSocket` instances and another 1-1 correspondence between `WebSocket` instances and `{net,tls}.Socket` instances.

If user code changes this behavior all kind of issues may arise.

lpinca:
> To reproduce, please try steps (1) ~ (4).

I'm trying to create a test case based on this.

> (3) connect websocket server to a non-existent localhost:8080 & retry every 2 seconds

What does this mean?

```js
'use strict';

const http = require('http');

const data1 = `<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <script>
      (function () {
        var ws = new WebSocket('ws://localhost:8080');

        ws.onopen = function () {
          console.log('open');
        };
      })();
    </script>
  </body>
</html>`;

const data2 = `<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <script>
      (function () {
        function connect() {
          var ws = new WebSocket('ws://localhost:3000');

          ws.onclose = function() {
            setTimeout(connect, 2000);
          };
        }

        connect();
      })();
    </script>
  </body>
</html>`;

server.on('request', (req, res) => {
  if (req.url === '/foo') {
    res.writeHead(200, { 'Content-Type': 'text/html' });
    res.end(data1);
  } else if (req.url === '/bar') {
    res.writeHead(200, { 'Content-Type': 'text/html' });
    res.end(data2);
  } else {
    res.writeHead(404, { 'Content-Type': 'text/plain' });
    res.end(http.STATUS_CODES[404]);
  }
});

server.listen(8080, () => {
  console.log('Listening on *:8080');
});
```

Is this code correct?

I open /foo and establish a WebSocket connection. (1) and (2)
I open /bar in another tab and watch it fails to connect several. times (3)
I refresh the /foo tab. (4) I also tried closing the whole browser and reopened /foo again.

No error is thrown.

amark:
@lpinca I nailed it!

Turns out it was not a race condition with pinging another server, its a shared socket on receiving servers.

Here are options for a fix:

 - (What I hate, but guessing you'll like best) Crash on 2+ WebSocket Servers being attached to the same port.
 -  (What I recommend) Use type & existence checks to safe guard your code.
 - (What is probably the most "pure" option) Don't use shared references.
 - (What is probably the most correct option) Rewrite your removeListeners to remove all listeners from all servers that share same socket.

Test to instantly replicate:

```

function server(){
	return require('http').createServer().listen(8765);
}
var opt = {server: server()};

function sockets(){
	var wss = new WebSocket.Server(opt);

		ws.onmessage = function incoming(data) {
		  console.log('1msg', data);
		};

		ws.onerror = function(err){
			console.log('1err', ''+err);
		};

		ws.onclose = function(){
			console.log('1bye');
			ws.close();
		};

	});

	return wss;
}

var wsServer1 = sockets();
var wsServer2 = sockets();

setTimeout(function(){

	var ws = new WebSocket('ws://localhost:8765');
	ws.onclose = function(){
		console.log("bye");
	};
	ws.onerror = function(err){
		console.log("err", ''+err);
	};
	ws.onopen = function(){
		console.log("open");
	}
	ws.onmessage = function(msg){
		console.log("msg", msg);
	};

}, 100);
```

lpinca:
The example above is attaching the same HTTP server on two different `WebSocketServer`. This is **not** ok. It is invalid usage as the `'upgrade'` event will be handled by both `WebSocketServer`.

If you need a shared HTTP server between multiple `WebSocketServer` use this https://github.com/websockets/ws#multiple-servers-sharing-a-single-https-server.

lpinca:
Closing as we now have an explanation and it's not an issue in ws. Feel free to submit a patch that detects this invalid usage and throws an error. I'm more than happy to merge it.

lpinca:
For reference https://github.com/websockets/ws/pull/885

amark:
@lpinca 

(A) saying it is **not** OK but the code being OK with it is a bug. 

(B) it being documented as allowed but it then being possible to crash because no safe guards exists on those shared sockets is a bug.

If I were to submit a patch for (A) to throw on two socket servers with same http server, then it would contradict your documented use case of such (which is very valid!).

Please consider that (B) is a very reasonable case of exactly why safeguards should be added, so the use case is possible **and** doesn't crash.

It is your project, so you can do as you wish. Just please recognize you claimed my bug/patch was not a concern because your code only adds a listener once, yet it actually will add it more than once, and you said my fix is a mask to a deeper issue, but that deeper issue is actually a documented feature. It is fine to reject, just please note to others that either `ws` theory does not match to implementation, or that documented features can produce known unstable buggy side-effects in production.

lpinca:
> If I were to submit a patch for (A) to throw on two socket servers with same http server, then it would contradict your documented use case of such (which is very valid!).

What does it contradict specifically?

```js

const wss1 = new WebSocket.Server({ server });
const wss2 = new WebSocket.Server({ server });
```

This is invalid usage, not documented anywhere and we should throw an error if the user does this.

```js
const wss1 = new WebSocket.Server({ noServer: true });
const wss2 = new WebSocket.Server({ noServer: true });

  const pathname = url.parse(request.url).pathname;

  if (pathname === '/foo') {
    wss1.handleUpgrade(request, socket, head, function done(ws) {
      wss1.emit('connection', ws, request);
    });
    wss2.handleUpgrade(request, socket, head, function done(ws) {
      wss2.emit('connection', ws, request);
    });
  } else {
    socket.destroy();
  }
});
```

This is valid usage and does not crash anything if used.

> Just please recognize you claimed my bug/patch was not a concern because your code only adds a listener once, yet it actually will add it more than once, and you said my fix is a mask to a deeper issue, but that deeper issue is actually a documented feature.

Please show me where we document this usage:

```js

const wss1 = new WebSocket.Server({ server });
const wss2 = new WebSocket.Server({ server });
```

Yes, exactly your patch would have exactly masked the invalid usage. In my opinion broken code should crash and not keep running in a broken state.

amark:
I missed the `new WebSocket.Server({ server });` vs `new WebSocket.Server({ noServer: true });`, fair enough point.

Where we disagree: It is **not** OK to crash other people's in-production servers. *Maybe* if you throw them an error they can catch, then at least they can opt-out. But this bug was not emit to the `.onerror` callback (what is the point of that callback then?) **and** it cannot be caught. That forces everybody else to suffer simply because you have a particular *preference* - I personally have experienced sleepless nights while my parents were visiting from out of town, trying to get this salvaged on behalf of my dev community so it is fixed before their launches go out. Some of these machines I have limited access to and therefore cannot update after-the-fact, or have millions of users, or are running in an offline self-sustaining mesh network that is a template for refugee/disaster survival relief camps. This hurts other people, real world humans, that have no clue about code or tech, and they **especially** don't care about my (or yours) theoretical desire for code "purity".

The analogy here is code is like a building people use. The philosophy that code should crash rather than running in a broken state, is literally akin to planting layers of C4 bombs to explode *if* any 1 of 10 elevators goes out of service, or if a toilet gets clogged. A crash blows up the entire building and everyone in it and all the session/memory, it brings the whole server down, potentially indefinitely, on machines I may or may not have control to auto-restart. That sucks for me, but I can debug it, what really sucks is for them, they're the ones who suffer the most.

lpinca:
For what is worth here is the patch I was referring to https://github.com/websockets/ws/commit/364126620bf3fe4e6e0042d3b74ec53b5ccbbb08. No footgun (pun intended) for other developers.

Yes, we disagree. Patches like this are not useful and if you keep adding them, in the long run they will make your code impossible to debug.

Furthermore this was affecting only a small percentage of ws users (only gun users?) and this made me think from the beginning that there was something wrong with how ws was used (https://github.com/websockets/ws/issues/1564#issuecomment-491608878).

This wasn't an emergency patch to fix production servers from crashing. This was a blind patch to fix an invalid usage of a dependency in gun and the blind patch was submitted upstream...

That said even if not documented, the invalid usage is allowed and that is my fault. I hope https://github.com/websockets/ws/commit/364126620bf3fe4e6e0042d3b74ec53b5ccbbb08 will help in this regard.

Issue #377, opened 2014-10-16
Pass verifyClient result to `connection` event


pwnall:
I'm doing some expensive work in `verifyClient` and I'd like to reuse the result in the `connection` event handler.

I'm currently using the fact that the undocumented `WebSocket` property `upgradeReq` is the same request as `info.req` in `verifyClient`, and I'm modifying the request. This feels dirty though.

Will you please consider allowing any truthy `verifyClient` result, and passing it into the `connection` event handler?

If this seems reasonable, I'd be glad to prepare a pull request.

luin:
+1

rosendi:
+1

ChrisZieba:
This would be a nice addition. I'm using jwt's in my `verifyClient` code and it would be nice to cleanly save the decoded result fo use in the connection handler. Something like this:


```js
  const wss = new WebSocketServer({
    server: server,
    verifyClient: function(info, done) {
      let query = url.parse(info.req.url, true).query;
      jwt.verify(query.token, config.jwt.secret, function(err, decoded) {
        if (err) return done(false, 403, 'Not valid token');
   
        // Saving the decoded JWT on the client would be nice 
        done(true);
      });
    }
  });

  wss.on('connection', ws => {
   // get decoded JWT here?
  });
```

stellanhaglund:
@pwnall do you store the jwt as a cookie or do you use it as part of the url?

I'm implementing jwt into my code but I'm not quite sure how to deliver the token.

Any suggestions?

stellanhaglund:
oh and +1 ;)

pwnall:
@stefanocudini I'm not using jwt. Here is what I'm doing.

https://github.com/pwnall/w3gram-server/blob/70a3024527e72f184cfb4d0139de218f96690848/src/ws_connection.coffee#L44
https://github.com/pwnall/w3gram-server/blob/70a3024527e72f184cfb4d0139de218f96690848/src/ws_connection.coffee#L58
https://github.com/pwnall/w3gram-server/blob/70a3024527e72f184cfb4d0139de218f96690848/src/ws_connection.coffee#L22

I hope this helps.

dweremeichik:
Perhaps #1099  makes this functionality a little more public? Out of curiosity, how are you handling the invalid token on the client side? Using the 1006 error?

cjhowe7:
@ChrisZieba I ended up verifying the JWT in my connection handler as well. Did you find a better solution?

dmax1:
[...]

PR #1612, opened 2019-08-01
[feature] attach result of verifyClient to ws

adrianhopebailie:
This change attaches the result of `verifyClient` to the `WebSocket` instance that is created by the server.

This allows servers to provide an authentication function as the `verifyClient` server option and then subsequently have access to the authenticated client when handling the connection.

Example:
```javascript

const wss = new WebSocket.Server({ port: 8080, verifyClient: (info) => {
	const client = authenticateWebSocketHandshake(info.req)
	return client;
}});

  // This client is the same value returned in the `verifyClient` callback above
  const client = ws.client

  ws.on('message', function incoming(message) {
    console.log('received from %s: %s', client, message);
  });

  ws.send('something');
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ can use one of the many wrappers available on npm, like
   - [Simple server](#simple-server)
   - [External HTTP/S server](#external-https-server)
   - [Multiple servers sharing a single HTTP/S server](#multiple-servers-sharing-a-single-https-server)
+  - [Authentication on the server (sync)](#authentication-on-the-server-sync)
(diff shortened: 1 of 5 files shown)


lpinca:
https://github.com/websockets/ws/issues/377#issuecomment-462152231

lpinca:
@adrianhopebailie sorry but I'm :-1: on this. I'm closing. Discussion can continue if needed.

adrianhopebailie:
Np. Wasn't aware of the history here. Have submitted #1613 to update some docs to save someone doing this too.

Issue #1220, opened 2017-10-20
Compile with warnings


albandaft:
Hi guys
I have integrated the tool inside my project and I get this warning while compiling:

This issue is coming from the https://github.com/zalmoxisus/remote-redux-devtools/issues/102, but is a WS library issue.
Basically I notice that the BufferUtil.js check if the library is already present to use it, other use it implements it. The problem is that this will cause `require` to fire an warning.

    WARNING in ./node_modules/ws/lib/BufferUtil.js
    Module not found: Error: Can't resolve 'bufferutil' in '/Users/alban/projects/frontend-base/node_modules/ws/lib'
     @ ./node_modules/ws/lib/BufferUtil.js 35:21-42
     @ ./node_modules/ws/lib/Receiver.js
     @ ./node_modules/ws/index.js
     @ ./node_modules/socketcluster-client/lib/sctransport.js
     @ ./node_modules/socketcluster-client/lib/scsocket.js
     @ ./node_modules/socketcluster-client/index.js
     @ ./node_modules/remote-redux-devtools/lib/devTools.js
     @ ./node_modules/remote-redux-devtools/lib/index.js
     @ ./src/store/configureStore.js
     @ ./src/index.server.js
     @ multi ./src/index.server.js
    
    WARNING in ./node_modules/ws/lib/Validation.js
    Module not found: Error: Can't resolve 'utf-8-validate' in '/Users/alban/projects/frontend-base/node_modules/ws/lib'
     @ ./node_modules/ws/lib/Validation.js 10:22-47
     @ ./node_modules/ws/lib/Receiver.js
     @ ./node_modules/ws/index.js
     @ ./node_modules/socketcluster-client/lib/sctransport.js
     @ ./node_modules/socketcluster-client/lib/scsocket.js
     @ ./node_modules/socketcluster-client/index.js
     @ ./node_modules/remote-redux-devtools/lib/devTools.js
     @ ./node_modules/remote-redux-devtools/lib/index.js
     @ ./src/store/configureStore.js
     @ ./src/index.server.js
     @ multi ./src/index.server.js

Wouldn't it be better to just include these libraries in the `package.json` or remove totally them from loading?

lpinca:
This is done on purpose as both `bufferutil` and `utf-8-validate` have a native component which can fail to compile. Originally they were included in `package.json` but they have been removed to prevent installation from failing when compilation was not successful.
Removing them completely is not an option as they greatly improve the library performances.

[...]

PR #1626, opened 2019-09-06
fix: List optional peer dependencies

eps1lon:
Listing packages mentioned in https://github.com/websockets/ws#opt-in-for-performance-and-spec-compliance as optional peer dependencies.

Additional context:
* [optional peer dependencies](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md) (implemented in `yarn@1.13.0`, `npm@6.11.0` but not [yet](https://github.com/nodejs/node/pull/29273) bundled with node and `pnpm@?`

I'm currently not using this feature myself until node bundles `npm@6.11`.  Just creating this PR now so that we can discuss any concerns you have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,18 @@
   "dependencies": {
     "async-limiter": "^1.0.0"
   },
+  "peerDependencies": {
+    "bufferutil": "^4.0.1",
+    "utf-8-validate": "^5.0.2"
+  },
+  "peerDependenciesMeta": {
+    "bufferutil": {
+      "optional": true
+    },
+    "utf-8-validate": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "benchmark": "^2.1.4",
     "bufferutil": "^4.0.1",


lpinca:
Hmm, what happens wit yarn < 1.13.0 and npm < 6.11.0. I guess `peerDependenciesMeta` is just ignored right? So bufferutil and utf-8-validate would be treated as normal peer dependencies right?

eps1lon:
>  So bufferutil and utf-8-validate would be treated as normal peer dependencies right?

Yep. For those package managers it would create warnings. However, this should be preferred over creating incomplete dependency trees.

lpinca:
I think it makes more sense to add them as [optionalDependencies](https://docs.npmjs.com/files/package.json#optionaldependencies) if build failures do not cause installation to fail as per description. See discussion in https://github.com/websockets/ws/issues/1220.

lpinca:
The main difference is that peer dependencies are not installed by default which is nice. I like peer dependencies. The only thing I don't like is the warning that is raised with older versions of npm/yarn.

eps1lon:
Yeah I can't help you with deciding what strategy to use since I only encountered this package as a transitive dependency. 

>  The only thing I don't like is the warning that is raised with older versions of npm/yarn.

Heard this fairly often but IMO I'd rather have a warning (with a fix by upgrading the package manager) that does no harm instead of a dependency tree that is potentially broken. Especially with the amount of dependencies in the JS ecosystem you do not want to debug your dependency tree ever. That's what package managers are for but they require information from the packages.

lpinca:
I'll wait until `npm@6.11` is bundled in Node.js 10.

lpinca:
Thank you.

PR #1636, opened 2019-09-25
Update websocket.js to remove bad syntax

PatriciaReyes:
Remove ... which are causing errors on compilation


lpinca:
This is not bad syntax you need Node.js >= 8 or use ws@6 or older.

Issue #1640, opened 2019-09-29
createWebSocketStream - objectMode


tadeuszwojcik:
Hi, forgive me please if it's silly question, but is there a reason why duplex stream created via `createWebSocketStream` doesn't have `objectMode` set to true? `ws.onmessage` clearly produces/pushes separate messages not continuous stream of bytes, am I missing something?
I'm trying to run smth like this:
```js
const WebSocket = require('ws')
const ws = new WebSocket('wss://www.bitmex.com/realtime')
const messageStream = WebSocket.createWebSocketStream(ws, { encoding: 'utf8' })

ws.once('open', function open() {
  ws.send(JSON.stringify({ op: 'subscribe', args: ['orderBookL2'] }))
})

async function run() {
  for await (let message of messageStream) {
    console.log(message)
  }
}
await run()
```
Is such pattern fine? Seems to be working and each message is correct string that matches json object provided by the server, but could that break as `objectMode` is not set?
I also tried to send message via stream, but couldn't sort it out quickly.

Would you perhaps accept PR that would change https://github.com/websockets/ws/blob/master/lib/stream.js#L65:
to:
```js
const duplex = new Duplex({
    autoDestroy: false,
    emitClose: false,
    objectMode: false,
    readableObjectMode: false,
    writableObjectMode: false,
    ...options
  });
```
so I could provide objectMode flag via options?

lpinca:
Because it does not play well with backpressure handling. In object mode backpressure is handled by counting the number of objects and not the actual bytes of buffered messages.

Also it does not make much sense. As you noticed the WebSocket protocol is message oriented. What is pushed onto the readable queue of the wrapper stream is a full message. This means that every chunk you get from the `'data'` listeners of stream returned by `createWebSocketStream()`  is the full message itself.

The pattern you are using is ok and will not break unless you use `duplex.read()` directly.

tadeuszwojcik:
All makes sense, thanks a lot!

tadeuszwojcik:
I'm not 100% sure why, but I found the issue where stream returned by `createWebSocketStream` returns two messages when iterating via `for await` at once which obviously breaks stuff. When using standard 'message' callback messages are returned separately as they should. Also when I set `readableObjectMode` to true as suggested in this issue it fixes the issue. So perhaps object mode should be set to on by default? 
Here's small snippet that reproduces that:
```js
const WebSocket = require('ws')
const ws = new WebSocket('wss://api-pub.bitfinex.com/ws/2')
[...]

PR #1647, opened 2019-10-11
[minor] allow overriding default duplex stream options for createWebSocketStream

tadeuszwojcik:
closes https://github.com/websockets/ws/issues/1640

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -63,12 +63,9 @@ function createWebSocketStream(ws, options) {
   }
 
   const duplex = new Duplex({
+    ...options,
     autoDestroy: false,
-    emitClose: false,
-    objectMode: false,
-    readableObjectMode: false,
-    writableObjectMode: false,
-    ...options
+    emitClose: false
   });
 
   ws.on('message', function message(msg) {
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -63,12 +63,12 @@ function createWebSocketStream(ws, options) {
   }
 
   const duplex = new Duplex({
-    ...options,
     autoDestroy: false,
     emitClose: false,
     objectMode: false,
     readableObjectMode: false,
-    writableObjectMode: false
+    writableObjectMode: false,
+    ...options
   });
 
   ws.on('message', function message(msg) {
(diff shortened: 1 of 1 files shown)


lib/stream.js
```diff
@@ -63,12 +63,12 @@ function createWebSocketStream(ws, options) {
...
 
   const duplex = new Duplex({
-    ...options,
     autoDestroy: false,
     emitClose: false,
     objectMode: false,
     readableObjectMode: false,
-    writableObjectMode: false
+    writableObjectMode: false,
+    ...options
```
lpinca: Please keep it as it originally was and just remove `objectMode`, `writableObjectMode`, `readableObjectMode` whose value is `false` by default. `autoDestroy`, and `emitClose` must never be set to `true`.

Also only `readableObjectMode` should be allowed to be changed so perhaps just remove it?

tadeuszwojcik: Makes sense, done.

lib/stream.js
```diff
@@ -65,10 +65,7 @@ function createWebSocketStream(ws, options) {
   const duplex = new Duplex({
     ...options,
     autoDestroy: false,
-    emitClose: false,
-    objectMode: false,
```
lpinca: Should users be allowed to change this? If so why?

lib/stream.js
```diff
@@ -65,10 +65,7 @@ function createWebSocketStream(ws, options) {
...
-    readableObjectMode: false,
-    writableObjectMode: false
```
lpinca: Ditto.

lib/stream.js
tadeuszwojcik: Changed it as you requested in previous comment. In practice I'd like to be able to set `readableObjectMode` and `encoding`.

lpinca:
Thank you.

tadeuszwojcik:
Thanks!

tadeuszwojcik:
Do you have rough idea when those changes will be published to NPM? Currently using version from #master but would like to plan migration to stable v7.x, thanks!

lpinca:
ws@7.2.0 was released a few minutes ago.

tadeuszwojcik:
Great! Thank you!

PR #1670, opened 2019-12-13
Make server accept connections from multiple paths

hmak3d:
Currently, `path` option is single string.  This PR allows it to be RegExp or string[] in the [style of Express](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5bd5b28/types/express-serve-static-core/index.d.ts#L138).

This would allow Apollo Server to use single WebSocket server to service multiple paths.

It creates the WebSocket server via:
https://github.com/apollographql/apollo-server/blob/apollo-server-express@2.8.1/packages/apollo-server/src/index.ts#L117
-> https://github.com/apollographql/apollo-server/blob/apollo-server-express@2.8.1/packages/apollo-server-core/src/ApolloServer.ts#L602-L612
-> https://github.com/apollographql/subscriptions-transport-ws/blob/v0.9.16/src/server.ts#L135

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,25 @@ server.on('upgrade', function upgrade(request, socket, head) {
 server.listen(8080);
 ```
 
+### Server that accepts connections only at whitelisted paths
+
+```js
+const WebSocket = require('ws');
+
+const path = '/foo';
+// const path = ['/foo', '/bar'];
+// const path = /^\/(foo|bar)$/;
+const wss = new WebSocket.Server({ port: 8080, path: path });
+
+wss.on('connection', function connection(ws) {
+  ws.on('message', function incoming(message) {
+    console.log('received: %s', message);
+  });
+
+  ws.send('something');
+});
+```
+
 ### Client authentication
 
 ```js
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -169,21 +169,22 @@ class WebSocketServer extends EventEmitter {
    */
   shouldHandle(req) {
     const whitelist = this.options.path;
+    let isMatch = true;
 
     if (whitelist) {
       const index = req.url.indexOf('?');
       const pathname = index !== -1 ? req.url.slice(0, index) : req.url;
 
       if (whitelist instanceof RegExp) {
-        if (!whitelist.test(pathname)) return false;
+        isMatch = whitelist.test(pathname);
       } else if (whitelist instanceof Array) {
-        if (!whitelist.includes(pathname)) return false;
+        isMatch = whitelist.includes(pathname);
       } else {
-        if (pathname !== whitelist) return false;
+        isMatch = pathname === whitelist;
       }
     }
 
-    return true;
+    return isMatch;
   }
 
   /**
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -30,7 +30,7 @@ class WebSocketServer extends EventEmitter {
    * @param {String} options.host The hostname where to bind the server
    * @param {Number} options.maxPayload The maximum allowed message size
    * @param {Boolean} options.noServer Enable no server mode
-   * @param {String} options.path Accept only connections matching this path
+   * @param {(String | String[] | RegExp)} options.path Accept only connections matching this path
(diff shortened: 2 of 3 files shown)


lpinca:
Override [`WebSocketServer.prototype.shouldHandle()`](https://github.com/websockets/ws/blob/master/doc/ws.md#servershouldhandlerequest) instead. It exists exactly for this reason. If you need a custom logic override it.

hmak3d:
Thanks for taking time to look at this PR.  Changing prototype is certainly an option.  I was hoping to avoid changing the Apollo Server code since it just passes along the `path` option I give to the ApolloServer instance.  The prototype approach you suggested would effectively make the `path` option ignored.  Why not tweak that option instead?

Any chance the PR could be modified in a way that would make it acceptable?  Thanks.

lpinca:
> The prototype approach you suggested would effectively make the `path` option ignored. Why not tweak that option instead?

It depends on the implementation. The developer could use the `path` option or any other options in the custom implementation of `server.shouldHandle()`.

> Any chance the PR could be modified in a way that would make it acceptable?

No I'm sorry, this has already been proposed in the past

- https://github.com/websockets/ws/pull/1062
- https://github.com/websockets/ws/pull/1290

and it defeats the purpose of `server.shouldHandle()`.

hmak3d:
Hmm, I was hoping that the `ws` package could take the change.  Supporting multiple paths would allow the `WebSocket.Server` endpoint to change with zero down time (e.g., from `/old_api` to `/new_api`).  i.e., during a transitional period, `options.path` would be `string[]`/`RegExp` instead of `string`.  Also, as mentioned at the start of the conversation, this matches nicely with Express's multi-path support for [app.use(string | string[] | RegExp, ...)](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5bd5b2884fcbd6eb3e64f559f2db784672250ba5/types/express-serve-static-core/index.d.ts#L1054).

FYI, I don't manage the [apollo-server code that creates](https://github.com/apollographql/subscriptions-transport-ws/blob/e5d4ffe01cf0e920be98f1ca7caf57051cfff5d3/src/server.ts#L135) the `WebSocket.Server` instance.  I suppose I can post a PR for apollo-server but I'm _not_ sure they would be okay prototype patching `WebSocker.Server` as you've suggested.

Thanks for pointing out the other two PRs.  I understand your desire to keep the codebase "lean".  Perhaps If there were more such feature requests in the future, it would make sense to extend the functionality of the `path` option.  i.e., instead of duplicating prototype patches everywhere [that `new WebSocket.Server()` is called from], we can centralize the useful path matching logic into the official `WebSocket.Server.shouldHandle()` itself.  Once again, I appreciate your consideration.

lpinca:
There is no need to change `subscriptions-transport-ws` code. You can require `ws` (the same `ws` required by `subscriptions-transport-ws`) and define your custom `shouldHandle()` method.

```js
const { Server } = require('ws');

Server.prototype.shouldHandle = function() {
  // ...
}
```

This should just work.

If you don't like this, the `SubscriptionServer` also accepts an instance of [`WebSocket.Server`](https://github.com/apollographql/subscriptions-transport-ws/blob/a1976837c17fdf65918fc49009b17061d984b266/src/server.ts#L113) so you can do something like this.

```js
const http = require('http');
const { SubscriptionServer } from 'subscriptions-transport-ws';

const wsServer = new WebSocket.Server({ server/*,  ... */ });

wsServer.shouldHandle = function () {
  // ...
};

const subscriptionServer = new SubscriptionServer({ /* ... */ }, wsServer);

// ...

server.listen(0, 'localhost', function() {
  console.log('Server listening on http://localhost:%d', this.address().port);
});
```

One more option is to add the custom `shouldHandle()` on the `WebSocket.Server` internally created by the `SubscriptionServer` constructor:

```js
const { SubscriptionServer } from 'subscriptions-transport-ws';

const subscriptionServer = new SubscriptionServer({ /* ... */ }, { /* ... */ });

subscriptionServer.wsServer.shouldHandle = function () {
  // ...
};
```

I'm closing this now. Thank you anyway.

PR #1680, opened 2020-01-17
[fix] Prevent type error in Duplex stream implementation on Node >=12

peterkooijmans:
Recent versions of Node have added getters for some of the internal stream
properties and throws this state away when the stream is being ended. This
commit adds a polyfill for the internal properties that are now exposed
through getters, and checks for availability of properties that do not have
a corresponding getter before using them.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,6 +1,26 @@
 'use strict';
 
-const { Duplex } = require('stream');
+const { Duplex, Readable, Writable } = require('stream');
+
+/* istanbul ignore next */
+if (!Object.getOwnPropertyDescriptor(Readable.prototype, 'readableEnded')) {
+  Object.defineProperty(Readable.prototype, 'readableEnded', {
+    enumerable: false,
+    get() {
+      return this._readableState ? this._readableState.endEmitted : false;
+    }
+  });
+}
+
+/* istanbul ignore next */
+if (!Object.getOwnPropertyDescriptor(Writable.prototype, 'writableFinished')) {
+  Object.defineProperty(Writable.prototype, 'writableFinished', {
+    enumerable: false,
+    get() {
+      return this._writableState ? this._writableState.finished : false;
+    }
+  });
+}
 
 /**
  * Emits the `'close'` event on a stream.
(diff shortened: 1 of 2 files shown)


lpinca:
What problem is this trying to solve? I prefer to not use those getters.

peterkooijmans:
I'm attempting to solve the problem where on the Node version that I'm on (12.13.1) the internal fields of the Node stream that are used currently are set to undefined earlier in newer versions. This causes my server code to crash with a `TypeError` when I'm trying to end the web socket stream created with `createWebsocketStream()`:
``` 
../node_modules/ws/7.2.1/lib/stream.js:112
    if (ws._socket._writableState.finished) {
                                  ^

TypeError: Cannot read property 'finished' of undefined
    at Duplex.duplex._final (../node_modules/ws/7.2.1/lib/stream.js:112:35)
    at callFinal (_stream_writable.js:627:10)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)```

peterkooijmans:
> Can you show how and where this happens in Node.js core? AFAIK the _{readable,writable}State properties are never nulled.

When I first looked at the stack trace and saw those Node.js internal properties being undefined I turned to the documentation which showed that in newer Node versions there are getters for some of those properties. The implementation of those getters first checks to see if the internal property even exists. This led me to believe that these could become undefined at any time, which is why I proposed this patch.

> Also, can you create a minimal code example to reproduce the issue?

In attempting to create a minimal code example I looked into the issue a bit further, and it turns out the problem is in my code. I implement a custom HTTPS agent which I pass to the constructor options of the websocket. This custom HTTPS agent returns a wrapped TLS socket from its `connect()` method that doesn't copy the `_readableState` and `_writableState` properties and thus causes the crash.

Although I think it is better to use the officially exposed Node.Js properties where possible instead of the internal once, the decision to do this is ultimately upto you of course. Do you still want this pull request to be merged? If so I will attempt to fix it so it works on all supported Node.js versions. If not I will just close this pull request.

lpinca:
> The implementation of those getters first checks to see if the internal property even exists

True but IIRC those checks were added because the user can subclass a stream class and try to use those getters in the subclass constructor before calling the super constructor. This is never done by ws and arguably a user error.

> Do you still want this pull request to be merged?

No, thanks anyway.

PR #1682, opened 2020-01-21
[fix] Allow close code 1014

zed0:
IANA has assigned 1014 as "Bad Gateway"

https://www.iana.org/assignments/websocket/websocket.xhtml#close-code-number
https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent

ws support for 1012, and 1013 was added in b58f688bf0fa0e87a3cef87c2f5d01bf43590668


lpinca:
Thank you.

PR #1722, opened 2020-03-30
op speed

LongTengDao:
1. cache static methods of bufferutil exports.
2. from `if/else` to `?:`, which more faster (by <https://jsperf.com/switch-vs-two-if>), and reduce source code repeat.

### DIFF
--- a/lib/buffer-util.js
+++ b/lib/buffer-util.js
@@ -120,19 +120,17 @@ function viewToBuffer(view) {
 
 try {
   const bufferUtil = require('bufferutil');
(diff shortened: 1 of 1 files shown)


lpinca:
Thanks but I'm not interested in this. It hinders readability and the performance is insignificant. Furthermore this is not an hot path.

LongTengDao:
> Thanks but I'm not interested in this. It hinders readability and the performance gain is insignificant. Furthermore this is not an hot path.

@lpinca ok~ I thought each message recieved from client need to be unmask, so it's a hot code.

Issue #1674, opened 2020-01-05
Client authentication example doesn't work


uAliFurkanY:
```
const http = require('http');
const WebSocket = require('ws');

const server = http.createServer();
const wss = new WebSocket.Server({ noServer: true });

wss.on('connection', function connection(ws, request, client) {
  ws.on('message', function message(msg) {
    console.log(`Received message ${msg} from user ${client}`);
  });
});

server.on('upgrade', function upgrade(request, socket, head) {
  authenticate(request, (err, client) => {
    if (err || !client) {
      socket.destroy();
      return;
    }

    wss.handleUpgrade(request, socket, head, function done(ws) {
      wss.emit('connection', ws, request, client);
    });
  });
});
```
Gives output: Received message a from user undefined

lpinca:
I find it hard to believe because if you run that code it will simply crash as `authenticate()` is not implemented.

uAliFurkanY:
That is the example given in the readme

lpinca:
Yes exactly it is an example. The `authenticate()` implementation is a up to you. If you are looking for a working example check the `express-session-parse` example under the `examples` folder which uses the same logic.

uAliFurkanY:
Ok. My point is that `client` is never passed to `function connection(ws, request, client)`

lpinca:
It is, see `wss.emit('connection', ws, request, client);` inside the `authenticate()` function. You can check its value before emitting that event.

uAliFurkanY:
Oh, I thought it was always passed to `function connection(ws, request, client)`. Thanks for informing.

Issue #1629, opened 2019-09-17
README: "authenticate" function not defined


nahuel:
Here the `authenticate` function is not defined: https://github.com/websockets/ws/blob/master/README.md#client-authentication

It will be nice to show how to implement it, how to return a 401 error code and how to get the ws client if the authentication succeds

lpinca:
This is done on purpose as the implementation depends on the use case. You can click the link below the snippet for a full working example.

nahuel:
Ok, please close this issue and sorry for the inconvenience.

lpinca:
No problem.

PR #1741, opened 2020-04-16
[doc] improve authentication example

RoTerentev:
1. The remind about 'authenticate' function implementing
2. Sending 401 HTTP response instead socket destroying

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,10 +266,7 @@ wss.on('connection', function connection(ws, request, client) {
 });
 
 server.on('upgrade', function upgrade(request, socket, head) {
-  /**
-   * NOTE:
-   *  The `authenticate` function not implemented! Please replace that by your auth logic!
-   */
+  // This function is not defined on purpose. Implement it with your own logic.
   authenticate(request, (err, client) => {
     if (err || !client) {
       socket.end('HTTP/1.1 401 Unauthorized');
--- a/README.md
+++ b/README.md
@@ -269,7 +269,8 @@ server.on('upgrade', function upgrade(request, socket, head) {
   // This function is not defined on purpose. Implement it with your own logic.
   authenticate(request, (err, client) => {
     if (err || !client) {
-      socket.end('HTTP/1.1 401 Unauthorized');
+      socket.write('HTTP/1.1 401 Unauthorized\r\n\r\n');
+      socket.destroy();
       return;
     }
 
--- a/README.md
+++ b/README.md
@@ -266,9 +266,13 @@ wss.on('connection', function connection(ws, request, client) {
 });
 
 server.on('upgrade', function upgrade(request, socket, head) {
+  /**
+   * NOTE:
+   *  The `authenticate` function not implemented! Please replace that by your auth logic!
+   */
   authenticate(request, (err, client) => {
     if (err || !client) {
-      socket.destroy();
+      socket.end('HTTP/1.1 401 Unauthorized');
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -266,9 +266,13 @@ wss.on('connection', function connection(ws, request, client) {
 });
 
 server.on('upgrade', function upgrade(request, socket, head) {
+  /**
+   * NOTE:
+   *  The `authenticate` function not implemented! Please replace that by your auth logic!
```
lpinca, line 271: I don't think this is needed. If the example is copy-pasted an error will be thrown as the function is not defined.

README.md
```diff
@@ -266,9 +266,13 @@ wss.on('connection', function connection(ws, request, client) {
...
+   */
   authenticate(request, (err, client) => {
     if (err || !client) {
-      socket.destroy();
+      socket.end('HTTP/1.1 401 Unauthorized');
```
lpinca, line 275: The user might want to reply with something different so I don't think this is better than current example where the socket is just destroyed. Also it is better to destroy the socket to prevent it from being kept indefinitely open if the other peer does not close it.

lpinca:
Thanks for the PR but I don't think it actually improves the example.

README.md
RoTerentev: Sure,  I thought about it because of https://github.com/websockets/ws/issues/1674 and https://github.com/websockets/ws/issues/1629 (maybe there are similar issues).

README.md
RoTerentev: I didn't know it before, thank you (I thought that 'end' will also destroy the socket). The problem I want to solve is sending a http response back to the browser, when the WebSocket can't be established.

Is this snippet better ?

```javascript
socket.end('HTTP/1.1 401 Unauthorized');
socket.destroy();

// or
socket.end('HTTP/1.1 401 Unauthorized');
setTimeout(function () {
  socket && socket.destroy();
},  500);
```

lpinca:
```js
socket.write('HTTP/1.1 401 Unauthorized\r\n\r\n');
socket.destroy();
```

RoTerentev: Thanks

lpinca: Can you please apply this? I think it is reasonable since the function name is `authenticate`.

README.md
```diff
@@ -266,9 +266,13 @@ wss.on('connection', function connection(ws, request, client) {
 });
 
 server.on('upgrade', function upgrade(request, socket, head) {
+  /**
+   * NOTE:
+   *  The `authenticate` function not implemented! Please replace that by your auth logic!
+   */
```
lpinca, line 272:
```suggestion
  // This function is not defined on purpose. Implement it with your own logic.
```

README.md
```diff
@@ -266,9 +266,10 @@ wss.on('connection', function connection(ws, request, client) {
 });
 
 server.on('upgrade', function upgrade(request, socket, head) {
+  // This function is not defined on purpose. Implement it with your own logic.
   authenticate(request, (err, client) => {
     if (err || !client) {
-      socket.destroy();
+      socket.end('HTTP/1.1 401 Unauthorized');
```
lpinca, line 272:
```suggestion
      socket.write('HTTP/1.1 401 Unauthorized\r\n\r\n');
      socket.destroy();
```

lpinca:
Merged. Thank you.

Issue #1749, opened 2020-04-27
[Documentation/Feature] Socket identifier or someway to save information against a socket


barkermn01:
- [x ] I've searched for any related issues and avoided creating a duplicate
      issue.

#### Description

There seems to be a lack of information about how we can get a socket identifier. the reason for this is as your example shows there should be an authentication system. but then I need to save information about that authentication into my application.

E.G
Client Connection (socket)
Authentication (App) - needs to save what they are authenticated for
Server Upgrades / Accepts

I tried saving information against the socket E.G
```js
on('connection', (sock) => { sock.id = "something"; })
on('close', (sock) => {console.log("Socket closing", sock.uuid);})
```

If there is a method could it be documented, if there's not would it be possible to provide some method so it could be E.G a socket identifier against the socket on connection or reconnection?

lpinca:
> [x] I've searched for any related issues and avoided creating a duplicate issue

Not really :)

This is out of the scope of the project if you search through closed issues there are multiple suggestions on how to do this. You can add an ID as show in your example (https://github.com/websockets/ws/issues/859) and/or keep a map ID -> websocket.

barkermn01:
I'm sorry I socket identifier is out of the scope of this project then close the project because the whole point about WebSockets is to give a TCP like a stack, every TCP socket system implemented gives a way to identify the socket, you said my example works but it does not so clearly that's a bug, because as my example says it does not work the console.log gives an undefined.

![image](https://user-images.githubusercontent.com/37368/80484654-f7246280-894f-11ea-97fa-43e0a266c6b5.png)

lpinca:
I'm not sure what you are talking about.

```js
const uuid = require('uuid');
const WebSocket = require('ws');

const server = new WebSocket.Server({ port: 0 });

server.on('connection', function(ws) {
  ws.id = uuid.v4();

  console.log('open', ws.id);

  ws.on('close', function() {
    console.log('close', ws.id);
  });
  ws.close();
});

server.on('listening', function() {
  const { port } = server.address();
  const ws = new WebSocket(`ws://localhost:${port}`)
});
```

Run this and you'll see that it works as expected.

[Section 4.2.2](https://tools.ietf.org/html/rfc6455#section-4.2.2) of the RFC describes the server's opening handshake and has nothing to do with what you want. It's what the server should verify and return in the HTTP response sent to the client.

barkermn01:
yeah,

[...]

PR #1751, opened 2020-04-28
Update websocket.js

barkermn01:
Added support for id to be saved against the socket, allow a unique identifier to be specified.

This is due to class security, as an object can't have data written to it from outside the class for a persistent state.

For example
class obj{
this._id = "something"; // inside the object is valid
}
let o = new obj();
o._id = "something"; only updates o for the scope o is in not the actuall object's internal data

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -56,6 +56,7 @@ class WebSocket extends EventEmitter {
     this._receiver = null;
     this._sender = null;
     this._socket = null;
+    this._id = null;
 
     if (address !== null) {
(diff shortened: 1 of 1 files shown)


barkermn01:
This could be updated so that it uses the UUID module as a submodule and auto-populates the id if a flag to auto-populate is active,

lpinca:
Thanks for the PR but I'm not interested.

As discussed in https://github.com/websockets/ws/issues/1749#issuecomment-620713661 you can add the `id` property directly without using a getter/setter if wanted.

PR #1754, opened 2020-05-05
Add support for the EventTarget `once` option

Sceat:
As per the documentation this is something not browser related and should be supported to allow transparent polyfill usage

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

```js
import WebSocket from 'ws'

new WebSocket().addEventListener('message', () => {}, { once: true })
```

### DIFF
--- a/lib/event-target.js
+++ b/lib/event-target.js
@@ -109,13 +109,16 @@ const EventTarget = {
   /**
    * Register an event listener.
    *
-   * @param {String} method A string representing the event type to listen for
+   * @param {String} type A string representing the event type to listen for
    * @param {Function} listener The listener to add
-   * @param {Object} [options] An options object specifies characteristics
-   * about the event listener
+   * @param {Object} options An options object specifies characteristics about
+   *     the event listener
+   * @param {Boolean} options.once A `Boolean`` indicating that the listener
+   *     should be invoked at most once after being added. If `true`, the
+   *     listener would be automatically removed when invoked.
    * @public
    */
-  addEventListener(method, listener, options = {}) {
+  addEventListener(type, listener, options) {
     if (typeof listener !== 'function') return;
 
     function onMessage(data) {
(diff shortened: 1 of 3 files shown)


lpinca:
Well, the whole `EventTarget` is browser related. Also it doesn't make much sense to use `once` for the `'open'`, `'error'`, and `'close'` events as they are emitted only once.

Do you really need this feature? In Node.js you can use `ws.once()`.

Sceat:
well to build an hybrid node/web client you wouldn't know if you can use `.once` as it doesn't exist in the browser :/ this option allow to be transparent and not care about which WebSocket object you use

lib/event-target.js
```diff
@@ -111,13 +111,15 @@ const EventTarget = {
    *
    * @param {String} method A string representing the event type to listen for
    * @param {Function} listener The listener to add
+   * @param {Object} [options] An options object specifies characteristics about the event listener
```
lpinca, line 114: Please keep the limit of 80 chars per line.

lib/event-target.js
```diff
@@ -111,13 +111,15 @@ const EventTarget = {
...
    * @public
    */
-  addEventListener(method, listener) {
+  addEventListener(method, listener, options = {}) {
     if (typeof listener !== 'function') return;
 
     function onMessage(data) {
       listener.call(this, new MessageEvent(data, this));
+      if (options.once) this.removeListener(method, listener);
```
lpinca, line 122: I think it's better to add it back also for other events to ensure that the listener is removed as documented. This should also be supported for custom events (line 180).

Not sure if it's better to remove it before the listener is called or after as it is currently done. Does the browser removes it before or after?

lpinca:
Also this need tests.

lib/event-target.js
Sceat: no idea of what the browser do but it seems indeed better to remove it first to avoid any same tick problem

lpinca:
The documentation should also be updated https://github.com/websockets/ws/blob/master/doc/ws.md#websocketaddeventlistenertype-listener.

I think it does not worth the effort as it is only useful for the `'message'` event and the same can be achieved by removing the listener from the inside of the listener itself but I have no objections.

Anyway if you want to add it we have to do it properly.

Sceat:
> Also this need tests (also for `removeEventListener` to make it sure that is works for `once` events added with `addEventListener`).

Resolved the previous issues, but i'm kinda not very good at testing stuff :fearful:

Sceat:
rip coverage

lib/event-target.js
```diff
@@ -134,44 +137,38 @@ const EventTarget = {
...
+      onClose._listener = listener;
+      this[method](type, onClose);
+    } else if (type === 'error') {
+      onError._listener = listener;
+      this[method](type, onError);
+    } else if (type === 'open') {
+      onOpen._listener = listener;
+      this[method](type, onOpen);
+    } else {
+      this[method](type, listener);
```
Sceat: is this faster ? or just less verbose

lpinca: You mean `if...else` instead of `switch`? Last time I checked `switch` was not faster and I find an `if...else` chain easier to read. Furthermore it was already written as an `if...else` chain.

lpinca:
Thank you.

PR #1773, opened 2020-06-19
update api doc to clarify WebSocket options

sa-linetco:
clarify which of the mentioned http options can not be used in the WebSocket constructor but can be provided with the address parameter

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -255,6 +255,13 @@ This class represents a WebSocket. It extends the `EventEmitter`.
     depending on the `protocolVersion`.
   - `maxPayload` {Number} The maximum allowed message size in bytes.
   - Any other option allowed in [http.request()][] or [https.request()][].
+    - Exceptions are: `createConnection`, `socketPath`, `hostname`, `protocol`,
+      `timeout`, `method`, `auth`, `host`, `path` and `port` where the following
+      can be provided via the `address` parameter:
+      - `auth`: `ws://user:password@host`
+      - `host`: `ws://host`
+      - `path`: `ws://host/path`
+      - `port`: `ws://host:port`
 
 `perMessageDeflate` default value is `true`. When using an object, parameters
 are the same of the server. The only difference is the direction of requests.
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -242,7 +242,8 @@ This class represents a WebSocket. It extends the `EventEmitter`.
 
 - `address` {String|url.URL} The URL to which to connect.
 - `protocols` {String|Array} The list of subprotocols.
-- `options` {Object}
+- `options` {Object} Options given do not have any effect if parsed from the URL
+  given with the `address` parameter.
   - `followRedirects` {Boolean} Whether or not to follow redirects. Defaults to
     `false`.
   - `handshakeTimeout` {Number} Timeout in milliseconds for the handshake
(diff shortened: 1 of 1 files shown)


doc/ws.md
```diff
@@ -255,6 +255,13 @@ This class represents a WebSocket. It extends the `EventEmitter`.
     depending on the `protocolVersion`.
   - `maxPayload` {Number} The maximum allowed message size in bytes.
   - Any other option allowed in [http.request()][] or [https.request()][].
+    - Exceptions are: `createConnection`, `socketPath`, `hostname`, `protocol`,
```
lpinca, line 258: I don't think it makes sense to document this as it is an implementation detail. These options are allowed but silently ignored or overridden.

- `protocol` is replaced by the URL protocol.
- `auth` works but the URL userinfo takes precedence over it.
- `hostname` / `host` are replaced by the URL hostname.
- `port` is replaced by the URL port.
- `path` is replaced by the URL path and URL search.
- `socketPath` is specified when using  [UNIX Domain Sockets]
(https://github.com/websockets/ws/blob/master/doc/ws.md#unix-domain-sockets)
- `timeout` is replaced by the `handshakeTimeout` option.
- `method` cannot be changed, only GET can be used.

They are basically all options that are taken from the connection URL for obvious reasons.

sa-linetco: I understand that it makes sense to take these parameters from the connection URL. However the documentation currently gives wrong information on how to use the constructor options parameter.

In my specific case I needed to provide basic authorization, so I followed the documentation and used the `auth` option for that.
However, using the `auth` option does not lead to a `Authorization` header field being added.
And this is why I thought it would make sense to clarify that some options of `http` or `https` can not be used (opposed to the documentation currently saying so) but have to be given via the `address` parameter.

Server
```
const ws = require('ws');
const wss = new ws.Server({ port: 8182 });

wss.on('connection', function connection(ws, req) {
    console.log(req.headers);
});
```

Server output
```
{
  'sec-websocket-version': '13',
  'sec-websocket-key': 'aJheST4cnuU4iJfSlIIwFA==',
  connection: 'Upgrade',
  upgrade: 'websocket',
  'sec-websocket-extensions': 'permessage-deflate; client_max_window_bits',
  host: 'localhost:8182'
}
```

Client
```
const websocket = new WebSocket(_uri, {
        auth: 'user:password',
});
```

lpinca: This has been fixed in version 7.2.4, see https://github.com/websockets/ws/commit/eaf2ad1335ff7b502a7b119c576ea3c611c7a70c.

sa-linetco: Thanks for the input regarding that specific fix.

I still think it would be a good idea to clarify which of the options available in `http` or `https` are also usable in `ws`, should be provided in another way (e.g. URL) or can not be used.

lpinca: I disagree, we moved away from listing every single option in https://github.com/websockets/ws/commit/9e152f920a1818d8a23e0a113b4625c83f90d30d because keeping them in sync was a bit of a pain.

All options are accepted, what ws does with them is an implementation detail.

I would be ok documenting that options that are extracted from the connection URL have no effect (without listing them) because the URL is a required parameter.

sa-linetco: I agree with your proposal to document that options given with the `options` parameter do not have any effect if they were parsed from the URL given with the `address` parameter.
I will adapt the changes to your proposal.

What do you think of a note on the `address` parameter that, if a string is given, it has to comply with the WHATWG URL Standard? This way a user could follow the documentation and look up how individual fields should be represented in the `address` string.

lpinca: > What do you think of a note on the address parameter that, if a string is given, it has to comply with the WHATWG URL Standard?

I think it doesn't make much sense as URIs that follow this specification https://tools.ietf.org/html/rfc3986#section-3 would be valid URLs. The parser or the validation of parameters will then throw an error is something is wrong.

doc/ws.md
```diff
@@ -242,7 +242,8 @@ This class represents a WebSocket. It extends the `EventEmitter`.
 
 - `address` {String|url.URL} The URL to which to connect.
 - `protocols` {String|Array} The list of subprotocols.
-- `options` {Object}
+- `options` {Object} Options given do not have any effect if parsed from the URL
```
lpinca, line 245: Would you mind moving this note after "Any other option allowed in `http.request()` or `https.request()`" (the latest element in the list)? Thank you.

sa-linetco: done

lpinca:
Thank you.

PR #1778, opened 2020-07-24
Server clients can have a url

jackschmidt:
It took a little bit to find how to get the URL of a WebSocket.Server created client. The solution is simple, but it seems like it could easily be done by default. This makes the documentation a little bit shorter. I didn't see any differences when running tests.

`ws.url = req.url`

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -302,6 +302,8 @@ class WebSocketServer extends EventEmitter {
       }
     }
 
+    ws.url = req.url;
+
     if (extensions[PerMessageDeflate.extensionName]) {
       const params = extensions[PerMessageDeflate.extensionName].params;
       const value = format({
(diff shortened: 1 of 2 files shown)


lib/websocket-server.js
```diff
@@ -302,6 +302,8 @@ class WebSocketServer extends EventEmitter {
       }
     }
 
+    ws.url = req.url;
```
lpinca, line 305: `req.url` is only the the URL `path` so this is not acceptable.

lpinca: The problem is also getting the original URL `protocol` and `host` when server is behind a proxy. I don't think it's worth the added complexity that this will result in.

Users can already access `req.url`.

jackschmidt: That makes sense. A full solution is more trouble than it is worth. The simple `req.url` was enough for my use (behind a proxy). It seemed simple enough to add to the server code, but this would just be a different incompatibility between the WebSocket objects, not an improvement.

PR #1798, opened 2020-09-16
Add 401 response and some comments

ehpc:


examples/express-session-parse/index.js
```diff
@@ -52,13 +52,17 @@ app.delete('/logout', function (request, response) {
 // Create HTTP server by ourselves.
 //
 const server = http.createServer(app);
-const wss = new WebSocket.Server({ clientTracking: false, noServer: true });
+const wss = new WebSocket.Server({
+  clientTracking: false,
+  noServer: true // Manually handle upgrade
```
lpinca, line 57: I'm not very happy with this. I find it inconsistent. There is a comment only for the `noServer` option. There should a comment for every option or no comment.

I would prefer the user to refer to https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketserveroptions-callback.

examples/express-session-parse/index.js
```diff
@@ -52,13 +52,17 @@ app.delete('/logout', function (request, response) {
 // Create HTTP server by ourselves.
 //
 const server = http.createServer(app);
```
lpinca, line 54:
```suggestion
// Create an HTTP server.
//
const server = http.createServer(app);
```

examples/express-session-parse/index.js
```diff
@@ -52,13 +52,17 @@ app.delete('/logout', function (request, response) {
...
+//
+// Create WebSocket server without explicit HTTP server.
+//
```
lpinca, line 57:
```suggestion

//
// Create a WebSocket server completely detached from the HTTP server.
//
```

lpinca:
Thank you.

examples/express-session-parse/index.js
```diff
@@ -49,16 +49,21 @@ app.delete('/logout', function (request, response) {
...
+// Create a WebSocket server completely detached from the HTTP server.
+//
 const wss = new WebSocket.Server({ clientTracking: false, noServer: true });
 
 server.on('upgrade', function (request, socket, head) {
   console.log('Parsing session from request...');
 
   sessionParser(request, {}, () => {
     if (!request.session.userId) {
+      socket.write('HTTP/1.1 401 Unauthorized\r\n\r\n');
```
vtjnash, line 66: This example may be unreliable and incorrect because the `socket.destroy` call on the next line is allowed to discard this write. If you want to write data, you either need to wait for the write callback to call `destroy`, or call `end`. Per the documentation for 'destroy':

> Previous calls to write() may not have drained, and may trigger an ERR_STREAM_DESTROYED error. Use end() instead of destroy if data should flush before close, or wait for the 'drain' event before destroying the stream.

It will often work, since `socket` probably hasn't had anything written yet, so it will have flushed the write buffer immediately, but that is not dependable.

lpinca: Yes this is known. Using `socket.end(data)` would be the proper way to ensure that data is actually sent without calling `socket.destroy()`. The example uses `socket.write()` + `socket.destory()` to ensure that socket is closed immediately, even if the other peer is keeping the connection open on purpose.

vtjnash: I am confused then, and I would say this deserves a comment in the code. Since regardless, I don't believe this combination quite achieves your stated intent. The kernel will not close this connection immediately, because it needs the other end to ACK the FIN packet first, regardless of the presence of this write call. I don't think that that time delay is impacted by whether the peer sends that ACK first or the local kernel reaches SO_TIMEOUT first. Therefore the kernel's behavior upon of doing this correctly (with 'end') vs. the result of relying on undefined behavior (with this PR) I think are likely to be similar.

Note that the 'end' event is emitted as soon as this 'write' is able to be queued in the kernel, and thus it can be caused to wait, if the peer has not yet read prior data (due to the resulting backpressure in the kernel), but otherwise would be done immediately (it will not wait for the peer to ACK this write).

PR #1799, opened 2020-09-19
Indicate optional params in JSDoc

Jokero:
To avoid warnings in IDE:
image

### DIFF
--- a/lib/event-target.js
+++ b/lib/event-target.js
@@ -111,9 +111,9 @@ const EventTarget = {
    *
    * @param {String} type A string representing the event type to listen for
    * @param {Function} listener The listener to add
-   * @param {Object} options An options object specifies characteristics about
+   * @param {Object} [options] An options object specifies characteristics about
    *     the event listener
-   * @param {Boolean} options.once A `Boolean`` indicating that the listener
+   * @param {Boolean} [options.once] A `Boolean`` indicating that the listener
    *     should be invoked at most once after being added. If `true`, the
    *     listener would be automatically removed when invoked.
    * @public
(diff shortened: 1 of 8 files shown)


lpinca:
I'm not very happy with this because:

1. AFAIK there is no proper way to define the position of an optional arguments. For example in the `websocket.close([code[, reason]])` signature `reason` is optional but if specified it must be the second argument.
1. If we do this we should do it everywhere and not only in a couple of files.

Jokero:
1. I actually always thought that order of optional parameters is strict and if some function has such description
```js
/**
 * @param {number} [a]
 * @param {string} [b]
 * @param {boolean} [c]
 */
function test(a, b, c) {}
```
it can be used only this way
```js
test(1);
test(1, 'b');
test(1, 'b', true);
```
I wouldn't expect this will be correct
```js
test('b');
```
But I can't find any proof or disproof, jsdoc spec is quite poor on examples and comments

2. Maybe. But small improvements are still improvements

lpinca:
Assume that in your example the `a` argument is defined as `@param {string} [a]`. `test('b')` would then be correct.

lpinca:
Anyway if you want to get this merged, please do it for all files. It shouldn't take a lot of time and I can help if needed.

Thank you.

Jokero:
> test('b') would then be correct if b can be used as first argument when a is not provided.

I would never allow in real code to skip arguments as it's so not obvious and only confuses.

Ok, if you are not against these changes, I'll try to do it in all files

lpinca:
@Jokero are you interested in finishing this off?

Jokero:
I can try to finish it next week

lpinca:
@Jokero I went ahead and finished it, I hope you don't mind.

lpinca:
Thank you.

Jokero:
I don't mind 😄 Thanks for your work

PR #1808, opened 2020-10-13
Add Windows named pipe support for client

hyurl:
The `WebSocket` client has been supporting Unix socket for a long time, but lacks the support of Windows named pipe, 
because in the previous versions, the URL constructor used to parse the address cannot parse a path in windows style, 
and the following code will throw a TypeError:

```js
const ws = new WebSocket("ws+unix://\\\\?\\pipe\\C:\\example.sock");
```

To implement full support of IPC, I added two more steps for parsing the address and setting the socket path for connection,
now the above code can work as expected.

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -466,10 +466,32 @@ function initAsClient(websocket, address, protocols, options) {
   }
 
   let parsedUrl;
+  let isWinPipe;
 
   if (address instanceof URL) {
     parsedUrl = address;
     websocket.url = address.href;
+  } else if (isWinPipe = /^ws\+unix\:\/\/\\\\[\.\?]\\pipe\\/.test(address)) {
+    let pathname = address.slice(10);
+    let index = pathname.indexOf("?", 9);
+    let search = "";
+
+    if (index !== -1) {
+      search = pathname.slice(index);
+      pathname = pathname.slice(0, index);
+    }
+
+    parsedUrl = {
+      protocol: "ws+unix:",
+      hostname: "",
+      port: "",
+      host: "",
+      pathname,
+      search,
+      hash: "",
+      username: "",
+      password: ""
+    };
   } else {
     parsedUrl = new URL(address);
     websocket.url = address;
@@ -528,7 +550,12 @@ function initAsClient(websocket, address, protocols, options) {
     opts.auth = `${parsedUrl.username}:${parsedUrl.password}`;
   }
 
-  if (isUnixSocket) {
+  if (isWinPipe) {
+    const index = opts.path.indexOf(":", 11);
+
+    opts.socketPath = index === -1 ? opts.path : opts.path.slice(0, index);
+    opts.path = index === -1 ? "" : opts.path.slice(index + 1);
+  } else if (isUnixSocket) {
     const parts = opts.path.split(':');
 
     opts.socketPath = parts[0];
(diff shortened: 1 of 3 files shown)


lib/websocket.js
```diff
@@ -466,10 +466,33 @@ function initAsClient(websocket, address, protocols, options) {
   }
 
   let parsedUrl;
+  let isWinPipe;
 
   if (address instanceof URL) {
     parsedUrl = address;
     websocket.url = address.href;
+  } else if ((isWinPipe = /^ws\+unix:\/\/\\\\[.?]\\pipe\\/.test(address))) {
```
lpinca, line 474: I'm not happy to skip the URL parser which is why there is no IPC support on Windows. This is not a valid URL.

lib/websocket.js
```diff
@@ -528,7 +551,12 @@ function initAsClient(websocket, address, protocols, options) {
     opts.auth = `${parsedUrl.username}:${parsedUrl.password}`;
   }
 
-  if (isUnixSocket) {
+  if (isWinPipe) {
+    const index = opts.path.indexOf(':', 11);
```
lpinca, line 555: Wouldn't `index` be incorrectly set to `-1` is `opts.path` is something like `'\\.\pipe\x:/foo'`, `'\\.\pipe\x'` is a valid named pipe.

lpinca:
@hyurl how about something like this:

```diff
diff --git a/lib/websocket.js b/lib/websocket.js
index ce6f34c..7ab8677 100644
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -475,9 +475,9 @@ function initAsClient(websocket, address, protocols, options) {
     websocket.url = address;
   }
 
-  const isUnixSocket = parsedUrl.protocol === 'ws+unix:';
+  const isIPCUrl = parsedUrl.protocol === 'ws+unix:';
 
-  if (!parsedUrl.host && (!isUnixSocket || !parsedUrl.pathname)) {
+  if (!parsedUrl.host && (!isIPCUrl || !parsedUrl.pathname)) {
     throw new Error(`Invalid URL: ${websocket.url}`);
   }
 
@@ -528,11 +528,18 @@ function initAsClient(websocket, address, protocols, options) {
     opts.auth = `${parsedUrl.username}:${parsedUrl.password}`;
   }
 
-  if (isUnixSocket) {
-    const parts = opts.path.split(':');
+  if (isIPCUrl) {
+    if (opts.path.startsWith('\\\\.\\pipe\\', 1)) {
+      const [socketPath, path] = opts.path.split(':/');
 
-    opts.socketPath = parts[0];
-    opts.path = parts[1];
+      opts.socketPath = socketPath.slice(1);
+      opts.path = path ? `/${path}` : '/';
+    } else {
+      const [socketPath, path] = opts.path.split(':');
+
+      opts.socketPath = socketPath;
+      opts.path = path;
+    }
   }
 
   let req = (websocket._req = get(opts));

```

Limitations:

1.  `\\?\pipe\foo` cannot be used but using `\\.\pipe\foo` to connect to `\\?\pipe\foo` works anyway so I think it's not a problem.
2. An additional `/` must be added  after the protocol. For example: `ws+unix:///\\.\pipe\foo:/pathname?bar=baz`.

hyurl:
@lpinca I was thinking about something like `ws+unix:pathname?foo=bar` (no `//`), which is valid for the URL constructor, for example:

```js
const ws = new WebSocket("ws+unix:some.sock"); // this syntax should work both on Windows and Unix

const ws = new WebSocket("ws+unix:C:\\some.sock"); // this syntax only works on Windows

const ws = new WebSocket("ws+unix://some.sock"); // this syntax only works on Unix
```

All we have to do is add `\\?\pipe\` or `\\.\pipe\` internally for the user.

PR #1820, opened 2020-12-02
Update websocket.js

stultuss:
When **protList** is an array of length greater than 2, it always return 'Server sent an invalid subprotocol'。


lpinca:
Sorry but this does not make sense. `protList` should be an array, not a string. If the subprotocol sent by the server is not included in the `protList` array you will get the "Server sent an invalid subprotocol" error.

PR #1834, opened 2021-01-14
Fixed code issue in docs for broadcasting

RIAEvangelist:
in the example 

```js

      if (client !== ws && client.readyState === WebSocket.OPEN) {
        client.send(data);
      }
    });
  });
});
```

`ws` is not defined and throws an error. It should be `wss`. Just had this error myself and contribute to fix the docs.


RIAEvangelist:
I was implementing this for a server, I see here this is different.

Issue #1669, opened 2019-12-13
Decompression does not work on Node.js >= 12 if the BFINAL bit is set to 1


jeffrson:
#### Description

I want ws to connect to a self written WebSocket Server based on https://github.com/sta/websocket-sharp.
While the test program works when using NodeJS 10.17.0, it doesn't with v12 or v13 - although the client connects and can send data successfully to the server, it does not show any received data. Since it does not seem to respond on "ping", the server library closes the connection after some time.

The server cannot be too wrong, since another short test with https://github.com/theturtle32/WebSocket-Node succeeds.

#### Reproducible in:

- version: 7.2.0
- Node.js version(s): 13.3.0, 12.13.1
- OS version(s): Windows 10

#### Attachments:
```

const WebSocket = require('ws');
const ws = new WebSocket('ws://127.0.0.1:12345/test');

ws.on('open', () => {
  ws.on('message', data => console.log(data))
  ws.on('close', error => console.log('close'))
  ws.on('error', error => console.log(`Error: ${error}`))
  ws.on('unexpected-response', error => console.log('unexpected-response'))
  ws.on('upgrade', error => console.log('upgrade'))
  ws.on('ping', () => console.log('ping'))
  ws.on('pong', data => console.log('pong'))

  let login = {
      name: "TEST",
  }

  ws.send(JSON.stringify(login))
})

```

lpinca:
I don't think it's an issue in the library because given ws popularity this problem would have flooded the issue tracker.

Add a listener for `'data'` event on the internal socket and log the raw data.

```js
ws.on('open', function() {
  ws._socket.prependListener('data', console.log);
})
```

jeffrson:
> I don't think it's an issue in the library because given ws popularity this problem would have flooded the issue tracker.

Yes, sure - it's strange, however, that it works when moving to node10.

I followed your advice and added the listener to the internal socket. Data arrives - and it's the same for Node10 and Node12. Again, Node10 sees the data, Node12 doesn't.
Then I tried another websocket server, python based, and everything goes well. I compared data, and there is a bit of a difference. So I was poking through the source of websocket-sharp and it looks like it's related to compression.
I disabled compression (perMessagedeflate: false) and it works!

Now there are two things I don't understand here. Why's there a difference when using compression (websocket-sharp is using .Net's DeflateStream) and why does it work with Node10.

lpinca:
[...]

PR #1840, opened 2021-02-02
reset compressor/decompressor instead of re-initialize

cTn-dev:
Use `zlib.reset()` to (as per nodejs documentation `Reset the compressor/decompressor to factory defaults. Only applicable to the inflate and deflate algorithms.`)

Recently we have introduced/enabled `server_no_context_takeover` on one of our canary production servers which i monitor with nodejs with the help of this ws library (approx 250 connections on a single nodejs process connecting to the server).

With the introduction of `server_no_context_takeover` i have noticed significantly increased CPU usage on the client of approx 2x (8% -> 16%) to what it was before the parameter was added).

I have traced it down to significantly higher GC activity, which went from ~**100ms** total time/minute to ~**500ms**.
When it comes to actual GC counts, it went from ~**25 GC**/minute to ~**150**.
Minor GCs almost stopped and majority of the time was spent in the major GCs and incremental GCs.

The root cause of this overhead appears to be the constant destruction/re-initialization of the `zlib.createInflateRaw`.
The attached changes eliminate this overhead, and brings the performance back in line to where it was before the `server_no_context_takeover` parameter was introduced.
The changes are holding up fine in production/i haven't encountered any issues so far.

I don't know if there are any potential side-effects of the below changes, the changes do pass all the tests currently available for the library (although i am not sure the current tests would catch compressor/decompressor "corruption" resulting from the changes in this PR, at least i wasn't able to run into such case so far). 

Running the available speed.js points towards a very minor throughput improvement for the 5000 roundtrips part of the test.
Each benchmark run fluctuates quite a bit on my work machine so i opted for not posting the results here.

I am attaching a set of graphs showing the performance 
1. before `server_no_context_takeover` was introduced on the server side
2. with the current WS library version (v7.4.2)
3. and the performance after the patch below was pushed

Node version: LTS v14.15.4
Platform: Linux x64

image
image

### DIFF
--- a/lib/permessage-deflate.js
+++ b/lib/permessage-deflate.js
@@ -377,12 +377,11 @@ class PerMessageDeflate {
       );
 
       if (fin && this.params[`${endpoint}_no_context_takeover`]) {
-        this._inflate.close();
-        this._inflate = null;
-      } else {
-        this._inflate[kTotalLength] = 0;
-        this._inflate[kBuffers] = [];
+        this._inflate.reset();
       }
+
+      this._inflate[kTotalLength] = 0;
+      this._inflate[kBuffers] = [];
 
       callback(null, data);
     });
@@ -449,12 +448,11 @@ class PerMessageDeflate {
       this._deflate[kCallback] = null;
 
       if (fin && this.params[`${endpoint}_no_context_takeover`]) {
-        this._deflate.close();
-        this._deflate = null;
-      } else {
-        this._deflate[kTotalLength] = 0;
-        this._deflate[kBuffers] = [];
+        this._deflate.reset();
       }
+
+      this._deflate[kTotalLength] = 0;
+      this._deflate[kBuffers] = [];
 
       callback(null, data);
     });
--- a/lib/permessage-deflate.js
+++ b/lib/permessage-deflate.js
@@ -376,12 +376,12 @@ class PerMessageDeflate {
         this._inflate[kTotalLength]
       );
 
+      this._inflate[kTotalLength] = 0;
+      this._inflate[kBuffers] = [];
+
       if (fin && this.params[`${endpoint}_no_context_takeover`]) {
         this._inflate.reset();
       }
-
-      this._inflate[kTotalLength] = 0;
-      this._inflate[kBuffers] = [];
 
       callback(null, data);
     });
@@ -447,12 +447,12 @@ class PerMessageDeflate {
       //
       this._deflate[kCallback] = null;
 
+      this._deflate[kTotalLength] = 0;
+      this._deflate[kBuffers] = [];
+
       if (fin && this.params[`${endpoint}_no_context_takeover`]) {
         this._deflate.reset();
       }
-
-      this._deflate[kTotalLength] = 0;
-      this._deflate[kBuffers] = [];
 
       callback(null, data);
     });


lpinca:
> I don't know if there are any potential side-effects of the below changes, the changes do pass all the tests currently available for the library (although i am not sure the current tests would catch compressor/decompressor "corruption" resulting from the changes in this PR, at least i wasn't able to run into such case so far).

I don't know either and no, I don't think existing tests cover this condition but the patch seems good. We should just verify that resetting has the same behavior of re-instantiating.

lib/permessage-deflate.js
```diff
@@ -377,13 +377,12 @@ class PerMessageDeflate {
...
-        this._inflate.close();
-        this._inflate = null;
-      } else {
-        this._inflate[kTotalLength] = 0;
-        this._inflate[kBuffers] = [];
+        this._inflate.reset();
       }
 
+      this._inflate[kTotalLength] = 0;
+      this._inflate[kBuffers] = [];
```
lpinca, line 384: Nit: can you move this before the `if` statement?

lib/permessage-deflate.js
```diff
@@ -449,13 +448,12 @@ class PerMessageDeflate {
...
-        this._deflate.close();
-        this._deflate = null;
-      } else {
-        this._deflate[kTotalLength] = 0;
-        this._deflate[kBuffers] = [];
+        this._deflate.reset();
       }
 
+      this._deflate[kTotalLength] = 0;
+      this._deflate[kBuffers] = [];
```
lpinca, line 455: Ditto.

lpinca (approved):

lib/permessage-deflate.js
cTn-dev: done

lpinca:
Thank you. I'll cut a new release later and eventually revert if side effects will be reported.

cTn-dev:
Thank you as well, ill keep an eye on the issues list once the new build is out.
I don't expect any issues but i am indeed curious if there is some edge case i haven't thought of while testing this manually (will see).

schoudhary21:
ws_issue
with the latest version "7.4.3" i am getting this issue, it is just crashing.

cTn-dev:
@schoudhary21 I see you are running into this while using nw.js (it might be worth listing the version for it as well), are you able to reproduce the same crash in a regular node.js ?
Are you connecting to a specific url/have some specific repro could that you could share?

schoudhary21:
@cTn-dev i am using https://dl.nwjs.io/v0.51.1/nwjs-sdk-v0.51.1-win-x64.zip version of nw.js, for connecting localhost websocket server using 127.0.0.1
I never had a issue with ws module. after recent update of ws it stopped working.
Now i forced lower version using 'npm install ws@7.4.1' and this working perfectly fine for me.

it will be great if we can fix it as other npm modules 'engine.io' etc are using it so; many components with different versions of ws are going to give more problem.

thank you :)

lpinca:
@schoudhary21 are you able to consistently reproduce the issue and write a test case?

schoudhary21:
yes, every time.

lpinca:
Ok, would you be so kind to write some code here, using only `ws` that reproduce the issue? Thank you.

schoudhary21:
i did another test;
i installed the latest ws 7.4.3 and the app started crashing then i went to the ws package location and under lib folder in the 'permessage-deflate.js'  i change following line ...

**> if (fin && this.params[`${endpoint}_no_context_takeover`]) {
>         this._inflate.reset();
>       }
>** 

with this ....

**>       if (fin && this.params[`${endpoint}_no_context_takeover`]) {
>        this._inflate.close();
>         this._inflate = null;
>       }**

and now the app started working fine without any issue.

lpinca:
@schoudhary21 it would be great if you could write something like this:

```js

const server = new WebSocket.Server(
  {
    perMessageDeflate: true,
    port: 0
  },
  function () {
    const ws = new WebSocket(`ws://localhost:${server.address().port}`, {
      perMessageDeflate: {
        clientNoContextTakeover: true,
        threshold: 0
      }
    });

    ws.on('open', function () {
      ws.send('A'.repeat(50));
    });
  }
);

server.on('connection', function (ws) {
  ws.on('message', console.log);
  ws.on('error', console.error);
});
```

with the missing bits that make it crash.

schoudhary21:
ok i will try for that;

another key point it connects to my websocket server and don't crash and stay connected for long time but as soon as i try to send msg which is compression enabled then it crashes.

lpinca:
@jeffrson can you please log the handshake response headers and the data received? You can do that like this:

```js
const ws = new WebSocket('ws://127.0.0.1:8888/svc');

ws.on('headers', function(res) {
  console.log(res.headers);
}); 

ws.onopen = function() {
  ws._socket.on('data', console.log);
};
```

jeffrson:
No problem

```
<Buffer c1 0b 33 34 32 36 31 35 33 b7 00 00 00>
<Buffer c1 0b 33 34 32 36 31 35 33 b7 00 00 00>
```

lpinca:
What about the headers?

lpinca:
I used the wrong event in the above example, sorry. It's the `'upgrade'` event not `'headers'`.

jeffrson:
Yes, I copied the snippet from the email notification... never mind!

```
{
  server: 'websocket-sharp/1.0',
  upgrade: 'websocket',
  connection: 'Upgrade',
  'sec-websocket-accept': 'XNWJgYkWYdm162Df815xmcnJnos=',
  'sec-websocket-extensions': 'permessage-deflate; client_no_context_takeover; server_no_context_takeover'
}
```

lpinca:
Thank you. Will test it in a bit.

jeffrson:
BTW, if you need to test and have a Linux version of dotnet (https://docs.microsoft.com/de-de/dotnet/core/install/linux-debian), you may create a project like so:

```
dotnet new console
dotnet add package websocketsharp --prerelease
<replace Program.cs with the file from the repo>
dotnet run
```

This will work on Windows too with .Net SDK only (I strongly assume, but I have an installation of Visual Studio)

lpinca:
I can reproduce. Here is a test case:

```js
const http = require('http');

server.on('upgrade', (req, socket) => {
  console.log(req.headers);

  const key = crypto
    .createHash('sha1')
    .update(req.headers['sec-websocket-key'] + GUID)
    .digest('base64');

  socket.on('data', function () {
    socket.write(Buffer.from('c10b33343236313533b7000000', 'hex'));
    socket.write(Buffer.from('c10b33343236313533b7000000', 'hex'));
  });

  socket.write(
    [
      'Upgrade: websocket',
      'Connection: Upgrade',
      'Sec-Websocket-Extensions: permessage-deflate; ' +
        'client_no_context_takeover; server_no_context_takeover',
      '\r\n'
    ].join('\r\n')
  );
});

server.listen(function () {
  const ws = new WebSocket(`ws://localhost:${server.address().port}`);

  ws.on('open', function () {
    ws.send('foo');
  });

  ws.on('message', console.log);
});
```

One thing I noticed is that our server behaves slightly differently

```js

const server = new WebSocket.Server(
  {
    perMessageDeflate: {
      clientNoContextTakeover: true,
      serverNoContextTakeover: true,
      threshold: 0
    },
    port: 0
  },
  function () {
    const ws = new WebSocket(`ws://localhost:${server.address().port}`);

    ws.on('open', function () {
      ws._socket.on('data', console.log);
      ws.send('foo');
    });

    ws.on('message', console.log);
  }
);

server.on('connection', function (ws) {
  ws.on('error', console.error);
  ws.on('message', function (message) {
    console.log(message);

    ws.send('12345678');
    ws.send('12345678');
  });
});
```

The frame with `ws` server is `c1 0a 32 34 32 36 31 35 33 b7 00 00` and not `c1 0b 33 34 32 36 31 35 33 b7 00 00 00`.

I will dig deeper later.

lpinca:
Here is a test case that reproduces the issue using only the Node.js `zlib` module:

```js
const zlib = require('zlib');

const TRAILER = Buffer.from([0x00, 0x00, 0xff, 0xff]);
const data = Buffer.from([
  0x33,
  0x34,
  0x32,
  0x36,
  0x31,
  0x35,
  0x33,
  0xb7,
  0x00,
  0x00,
  0x00
]);

const inflate = zlib.createInflateRaw({
  windowBits: zlib.Z_DEFAULT_WINDOWBITS
});

inflate.resume();

inflate.on('error', console.error);

inflate.write(data);
inflate.write(TRAILER);
inflate.flush(function () {
  inflate.reset();

  inflate.write(data);
  inflate.write(TRAILER);
  inflate.flush(function() {
    inflate.reset();
  })
});
```

If the second `inflate.reset()` is remove the following error is emitted

```
Error [ERR_STREAM_PUSH_AFTER_EOF]: stream.push() after EOF
    at new NodeError (node:internal/errors:329:5)
    at readableAddChunk (node:internal/streams/readable:276:30)
    at InflateRaw.Readable.push (node:internal/streams/readable:227:10)
    at Zlib.processCallback (node:zlib:553:10) {
  code: 'ERR_STREAM_PUSH_AFTER_EOF'
}
```

so it seems the stream already ended when the first `inflate.flush()` callback is called.

This is a bug in `ws`. We should check if the stream ended before writing more data to it.

lpinca:
Something like this

```diff
diff --git a/lib/permessage-deflate.js b/lib/permessage-deflate.js
index 74bf14a..a8974b9 100644
--- a/lib/permessage-deflate.js
+++ b/lib/permessage-deflate.js
@@ -376,11 +376,16 @@ class PerMessageDeflate {
         this._inflate[kTotalLength]
       );
 
-      this._inflate[kTotalLength] = 0;
-      this._inflate[kBuffers] = [];
+      if (this._inflate._readableState.endEmitted) {
+        this._inflate.close();
+        this._inflate = null;
+      } else {
+        this._inflate[kTotalLength] = 0;
+        this._inflate[kBuffers] = [];
 
-      if (fin && this.params[`${endpoint}_no_context_takeover`]) {
-        this._inflate.reset();
+        if (fin && this.params[`${endpoint}_no_context_takeover`]) {
+          this._inflate.reset();
+        }
       }
 
       callback(null, data);
```

should fix the issue but I have to add a regression test and check that it works on all supported Node.js versions.

jeffrson:
Oh - great work! I had only tested with some of zlib options of WebSocket.Server but could not reproduce.

cTn-dev:
@lpinca Is this an "expected" behavior of zlib triggered by the trailer shown above that some other WS libraries use ?
I assume RFC 7692 is "loose enough" to allow for a slightly different compressor/decompressor "reset" logic which this regression falls under?

lpinca:
@cTn-dev yes, see https://github.com/websockets/ws/issues/1669#issuecomment-565836942. See also https://github.com/nodejs/node/commit/28db96f31c2724b16d6b9bec19c600023365b7cc and https://github.com/nodejs/node/commit/0e89b64d6606d9c7948259e9090a46d185c5281f.

The test case above (https://github.com/websockets/ws/pull/1840#issuecomment-791659952) should work without errors on Node.js < 12.

lpinca:
@schoudhary21 @jeffrson the issue should be fixed in `ws@7.4.4`.

jeffrson:
Yes, it's fixed here. Thank you!

PR #1843, opened 2021-02-06
docs: Fixing `send` data type

rschristian:
The type for `data` within the `send()` function is not `any`, but instead `<string> | <Buffer>`.

Try to say pass an object, and you'll be met with: 

```
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
```

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -417,7 +417,8 @@ receives an `OpenEvent` named "open".
 
 ### websocket.ping([data[, mask]][, callback])
 
-- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The data to send in the ping frame.
+- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The
+  data to send in the ping frame.
 - `mask` {Boolean} Specifies whether `data` should be masked or not. Defaults to
   `true` when `websocket` is not a server client.
 - `callback` {Function} An optional callback which is invoked when the ping
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -456,7 +456,7 @@ Removes an event listener emulating the `EventTarget` interface.
 
 ### websocket.send(data[, options][, callback])
 
-- `data` {string | Buffer} The data to send.
+- `data` {number | array | arrayBuffer | buffer | object | string} The data to send.
 - `options` {Object}
   - `compress` {Boolean} Specifies whether `data` should be compressed or not.
     Defaults to `true` when permessage-deflate is enabled.
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -428,7 +428,8 @@ Send a ping.
 
 ### websocket.pong([data[, mask]][, callback])
 
-- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The data to send in the pong frame.
+- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The
+  data to send in the pong frame.
 - `mask` {Boolean} Specifies whether `data` should be masked or not. Defaults to
   `true` when `websocket` is not a server client.
 - `callback` {Function} An optional callback which is invoked when the pong
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -456,7 +456,7 @@ Removes an event listener emulating the `EventTarget` interface.
 
 ### websocket.send(data[, options][, callback])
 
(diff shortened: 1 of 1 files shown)


lpinca:
This is not correct, you can also pass a number, and anything accepted by `Buffer.from()`.

doc/ws.md
```diff
@@ -456,7 +456,7 @@ Removes an event listener emulating the `EventTarget` interface.
 
 ### websocket.send(data[, options][, callback])
 
-- `data` {Any} The data to send.
+- `data` {string | Buffer} The data to send.
```
lpinca, line 459: This is not correct, you can also pass a number, and anything accepted by `Buffer.from()`

rschristian: Apologies, you're totally right. I'll get that updated.

rschristian: Though objects are accepted by `Buffer.from()`, but not `socket.send()`.

lpinca: It does not matter. If possible, the `data` argument is converted to `Buffer` before being sent.

rschristian: Ah, yeah, I had something odd I was trying to use. I didn't look at the specifics for the implementations.

rschristian: Hopefully that update is (more) correct. It's `number` + all the `Buffer.from()` args. Apologies if it's not, still super new to the lib, but the `Any` threw me for a loop.

doc/ws.md
```diff
@@ -456,7 +456,7 @@ Removes an event listener emulating the `EventTarget` interface.
 
 ### websocket.send(data[, options][, callback])
 
-- `data` {Any} The data to send.
+- `data` {number | array | arrayBuffer | buffer | object | string} The data to send.
```
lpinca, line 459:
```suggestion
- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The data to send.
```

Can you please change it also for `websocket.ping()` and `websocket.pong()`?

rschristian: From what I understand, `DataView` is pretty much just an `ArrayBuffer`, though with a specific task, and `TypedArray` isn't actually an accessible property/constructor. Does it really make sense to add them?

Will add to `ping()` and `pong()`.

lpinca: > From what I understand, `DataView` is pretty much just an `ArrayBuffer`, though with a specific task, and `TypedArray` isn't actually an accessible property/constructor. Does it really make sense to add them?

Yes it does

```
> var buffer = new ArrayBuffer(2);
undefined
> var view = new DataView(buffer);
undefined
> Buffer.from(view)
<Buffer >
> Buffer.from(view.buffer)
<Buffer 00 00>
```

doc/ws.md
```diff
@@ -417,7 +417,7 @@ receives an `OpenEvent` named "open".
 
 ### websocket.ping([data[, mask]][, callback])
 
-- `data` {Any} The data to send in the ping frame.
+- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The data to send in the ping frame.
```
lpinca, line 420:
```suggestion
- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The
  data to send in the ping frame.
```

doc/ws.md
```diff
@@ -427,7 +427,7 @@ Send a ping.
 
 ### websocket.pong([data[, mask]][, callback])
 
-- `data` {Any} The data to send in the pong frame.
+- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The data to send in the pong frame.
```
lpinca, line 430:
```suggestion
- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The
  data to send in the pong frame.
```

doc/ws.md
```diff
@@ -456,7 +456,7 @@ Removes an event listener emulating the `EventTarget` interface.
 
 ### websocket.send(data[, options][, callback])
 
-- `data` {Any} The data to send.
+- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The data to send.
```
lpinca, line 459:
```suggestion
- `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The
  data to send.
```

lpinca (commented):
Not sure why Travis CI and the linter are not running.

lpinca:
Thank you.

PR #1901, opened 2021-06-14
Attach error codes to all receiver errors

### LINKED ISSUE #1892: Expose error status codes
- [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description

When ws throws errors due to received content, it associates a websocket status code with them, which specifies the status code that will be sent when closing the connection.

It attaches this to the error behind a symbol called `kStatusCode` ([used here](https://github.com/websockets/ws/blob/990306d1446faf346c76452409a4c11455690514/lib/receiver.js#L505), [defined here](https://github.com/websockets/ws/blob/990306d1446faf346c76452409a4c11455690514/lib/constants.js#L6)), keeping that status code as internal private state that can't be easily or reliably accessed by application code.

That is inconvenient, because in some cases this is useful information for applications too, e.g. to tracking metrics of websocket failure reasons, or to implement various advanced error handling scenarios that depend on type of protocol failure involved.

Would you be open to making this a non-symbol property, so it's available as a reliable part of the API for error handling?

pimterry:
Fixes #1892

~~This change doesn't change any messages or close codes, but does move the existing values into constants linked to the error code, to avoid repeating them at every point where the error is thrown.~~

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -45,7 +45,7 @@
   - [websocket.url](#websocketurl)
 - [WebSocket.createWebSocketStream(websocket[, options])](#websocketcreatewebsocketstreamwebsocket-options)
 - [WS Error Codes](#ws-error-codes)
-  - [WS_ERR_INVALID_DATA_PAYLOAD_LENGTH](#wserrinvaliddatapayloadlength)
+  - [WS_ERR_UNSUPPORTED_MESSAGE_LENGTH](#wserrunsupporteddatapayloadlength)
   - [WS_ERR_INVALID_CONTROL_PAYLOAD_LENGTH](#wserrinvalidcontrolpayloadlength)
   - [WS_ERR_INVALID_UTF8](#wserrinvalidutf8)
   - [WS_ERR_INVALID_OPCODE](#wserrinvalidopcode)
@@ -309,7 +309,8 @@ human-readable string explaining why the connection has been closed.
 
 - `error` {Error}
 
-Emitted when an error occurs. Errors may have a `.code` property, matching one of the string values defined below under [WS Error Codes](#ws-error-codes).
+Emitted when an error occurs. Errors may have a `.code` property, matching one
+of the string values defined below under [WS Error Codes](#ws-error-codes).
 
 ### Event: 'message'
 
@@ -506,11 +507,13 @@ given `WebSocket`.
 
 ## WS Error Codes
 
-Errors emitted by the websocket may have a `.code` property, describing the specific type of error that has occurred:
+Errors emitted by the websocket may have a `.code` property, describing the
+specific type of error that has occurred:
 
-### WS_ERR_INVALID_DATA_PAYLOAD_LENGTH
+### WS_ERR_UNSUPPORTED_MESSAGE_LENGTH
 
-A data frame with an unexpectedly long payload was received.
+A message was received with a length longer than the maximum supported length
+for this endpoint.
 
 ### WS_ERR_INVALID_CONTROL_PAYLOAD_LENGTH
 
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -46,6 +46,7 @@
 - [WebSocket.createWebSocketStream(websocket[, options])](#websocketcreatewebsocketstreamwebsocket-options)
 - [WS Error Codes](#ws-error-codes)
   - [WS_ERR_UNSUPPORTED_MESSAGE_LENGTH](#wserrunsupporteddatapayloadlength)
+  - [WS_ERR_UNSUPPORTED_DATA_PAYLOAD_LENGTH](#wserrunsupporteddatapayloadlength)
   - [WS_ERR_INVALID_CONTROL_PAYLOAD_LENGTH](#wserrinvalidcontrolpayloadlength)
   - [WS_ERR_INVALID_UTF8](#wserrinvalidutf8)
   - [WS_ERR_INVALID_OPCODE](#wserrinvalidopcode)
@@ -526,7 +527,7 @@ A control frame with an invalid payload length was received.
 
 ### WS_ERR_INVALID_UTF8
 
-A text frame was received containing invalid UTF-8 data.
+A text or close frame was received containing invalid UTF-8 data.
 
 ### WS_ERR_INVALID_OPCODE
 
@@ -538,11 +539,11 @@ A WebSocket close frame was received with an invalid close code.
 
 ### WS_ERR_UNEXPECTED_RSV_1
 
-A WebSocket frame was received with the RSV 1 bit set unexpectedly.
+A WebSocket frame was received with the RSV1 bit set unexpectedly.
 
 ### WS_ERR_UNEXPECTED_RSV_2_3
 
-A WebSocket frame was received with the RSV 2 or 3 bits set unexpectedly.
+A WebSocket frame was received with the RSV2 or RSV3 bit set unexpectedly.
 
 ### WS_ERR_EXPECTED_FIN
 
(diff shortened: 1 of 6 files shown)


lpinca:
Can you please only add the codes? I'm not sure I like the `errorDefinition` stuff on `lib/constant.js`.

lpinca:
Also can you prefix them with `WS_ERR_` instead of just `WS_` and document them in `doc/ws.md` similar to how it is done here https://nodejs.org/api/errors.html#errors_node_js_error_codes?

pimterry:
Fair enough, done! Docs updated too.

I've changed it for now, but I do think there's a benefit to commonizing the error details to handle the error cases explicitly and make them more maintainable in future. It means quite a bit of duplication of error params otherwise, where 90% of them are just "[msg], true, 1002, [code]", and it makes it harder to quickly check if the close codes and messages and error codes are used consistently or not for each case. Up to you of course though.

lpinca:
Maybe but they are all in one place at the moment so it's ok for now. It also keep the commit/PR focused at doing only one thing and the added benefit is that there is no need to open a new file to read the error message and close code.

lib/receiver.js
```diff
@@ -285,7 +346,8 @@ class Receiver extends Writable {
         RangeError,
         'Unsupported WebSocket frame: payload length > 2^53 - 1',
         false,
-        1009
+        1009,
+        'WS_ERR_INVALID_DATA_PAYLOAD_LENGTH'
```
lpinca, line 350: I think "invalid" is a bit misleading here. It's not invalid is not supported by `ws`.

lib/receiver.js
```diff
@@ -304,7 +366,13 @@ class Receiver extends Writable {
       this._totalPayloadLength += this._payloadLength;
       if (this._totalPayloadLength > this._maxPayload && this._maxPayload > 0) {
         this._loop = false;
-        return error(RangeError, 'Max payload size exceeded', false, 1009);
+        return error(
+          RangeError,
+          'Max payload size exceeded',
+          false,
+          1009,
+          'WS_ERR_INVALID_DATA_PAYLOAD_LENGTH'
```
lpinca, line 374: How about `WS_ERR_MESSAGE_TOO_BIG` or something like that? The payload length is not invalid per se.

lib/receiver.js
```diff
@@ -384,7 +452,13 @@ class Receiver extends Writable {
         this._messageLength += buf.length;
         if (this._messageLength > this._maxPayload && this._maxPayload > 0) {
           return cb(
-            error(RangeError, 'Max payload size exceeded', false, 1009)
+            error(
+              RangeError,
+              'Max payload size exceeded',
+              false,
+              1009,
+              'WS_ERR_INVALID_DATA_PAYLOAD_LENGTH'
```
lpinca, line 460: Ditto.

lib/receiver.js
pimterry: I've renamed the `WS_ERR_INVALID_DATA_PAYLOAD_LENGTH` code to `WS_ERR_UNSUPPORTED_MESSAGE_LENGTH`. That's a bit more consistent with the other codes (`MESSAGE_TOO_BIG` isn't a noun), but it's explicitly a message rather than a single frame, and it's just not supported in this connection rather than completely invalid. Does that work for you?

lpinca: `WS_ERR_UNSUPPORTED_MESSAGE_LENGTH` is perfect to address my previous comment where payload length > 2^53 - 1. However in this case it is not unsupported. The error in this case is that the message length exceeds the maximum allowed message length as defined by the `maxPayload` option.

lib/receiver.js
lpinca: `WS_ERR_UNSUPPORTED_DATA_PAYLOAD_LENGTH`?

lib/receiver.js
pimterry: > However in this case it is not unsupported. The error in this case is that the message length exceeds the maximum allowed message length as defined by the maxPayload option.

That's what I mean by 'unsupported'. Messages that go beyond it aren't "invalid" - imo invalid means against the specification, i.e. this will never work on any conformant websocket connection.

With 'unsupported' I'm trying to say that it may be valid, and it might work elsewhere, but it won't work on this connection right now (e.g. due to the configured maxPayload option). That matches standard use elsewhere - e.g. close code 1003 ("unsupported data" - valid data in a format that this server can't process) or HTTP 415 ("unsupported media type" - a valid request with a media type that this server refuses to handle).

I think that fits both this case (unsupported due to ws maxPayload config) and the other case (unsupported due to JS limitations).

I see what you mean about data frame size vs total message size though. How about we split the two cases into:

* `WS_ERR_UNSUPPORTED_MESSAGE_LENGTH` - overall message is too long (> maxPayload)
* `WS_ERR_UNSUPPORTED_DATA_PAYLOAD_LENGTH` - a single data frame is too long (> 2^53-1)

lpinca: > How about we split the two cases into

Works for me and that is what I prefer.

pimterry: Great, done :+1:

doc/ws.md
```diff
@@ -44,6 +44,17 @@
   - [websocket.terminate()](#websocketterminate)
   - [websocket.url](#websocketurl)
 - [WebSocket.createWebSocketStream(websocket[, options])](#websocketcreatewebsocketstreamwebsocket-options)
+- [WS Error Codes](#ws-error-codes)
+  - [WS_ERR_UNSUPPORTED_MESSAGE_LENGTH](#wserrunsupporteddatapayloadlength)
+  - [WS_ERR_INVALID_CONTROL_PAYLOAD_LENGTH](#wserrinvalidcontrolpayloadlength)
```
lpinca, line 49:
```suggestion
  - [WS_ERR_UNSUPPORTED_DATA_PAYLOAD_LENGTH](#wserrunsupporteddatapayloadlength)
  - [WS_ERR_INVALID_CONTROL_PAYLOAD_LENGTH](#wserrinvalidcontrolpayloadlength)
```

doc/ws.md
```diff
@@ -493,6 +505,57 @@ The URL of the WebSocket server. Server clients don't have this attribute.
...
+A data frame was received with a length longer the max supported length (2^53-1,
+due to JavaScript language limitations).
+
+### WS_ERR_INVALID_CONTROL_PAYLOAD_LENGTH
+
+A control frame with an invalid payload length was received.
+
+### WS_ERR_INVALID_UTF8
+
+A text frame was received containing invalid UTF-8 data.
```
lpinca, line 529:
```suggestion
A text or close frame was received containing invalid UTF-8 data.
```

doc/ws.md
```diff
@@ -493,6 +505,57 @@ The URL of the WebSocket server. Server clients don't have this attribute.
...
+
+A WebSocket frame was received with an invalid opcode.
+
+### WS_ERR_INVALID_CLOSE_CODE
+
+A WebSocket close frame was received with an invalid close code.
+
+### WS_ERR_UNEXPECTED_RSV_1
+
+A WebSocket frame was received with the RSV 1 bit set unexpectedly.
```
lpinca, line 541:
```suggestion
A WebSocket frame was received with the RSV1 bit set unexpectedly.
```

doc/ws.md
```diff
@@ -493,6 +505,57 @@ The URL of the WebSocket server. Server clients don't have this attribute.
...
+
+### WS_ERR_UNEXPECTED_RSV_2_3
+
+A WebSocket frame was received with the RSV 2 or 3 bits set unexpectedly.
```
lpinca, line 545:
```suggestion
A WebSocket frame was received with the RSV2 or RSV3 bit set unexpectedly.
```

lpinca (approved):

lpinca:
Thank you.

PR #1902, opened 2021-06-14
Simplify the closing handshake

lpinca:
- When the socket emits the `'end'` event, do not call `socket.end()`.
  End only the `receiver` stream.
- Do not wait for a close frame to be received and sent before calling
  `socket.end()`. Call it right after the close frame is written.
- When the `receiver` stream emits `'finish'`, send a close frame if no
  close frame is received.

The assumption is that the socket is allowed to be half-open. On the
server side this is always true (unless the user explicitly sets the
`allowHalfOpen` property of the socket to `false`). On the client side
the user might use an agent so we set `socket.allowHalfOpen` to `true`
when the `http.ClientRequest` object emits the `'upgrade'` event.

Refs: https://github.com/websockets/ws/pull/1899

### DIFF
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -151,6 +151,7 @@ class Sender {
       }),
       cb
     );
+    this._socket.end();
   }
 
   /**
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -939,7 +939,9 @@ function socketOnEnd() {
   // the closing handshake but unlike `WebSocket.CLOSING` it does not prevent
   // `WebSocket.prototype.close()` from sending a close frame.
   //
-  websocket._readyState = -2;
+  if (websocket._readyState === WebSocket.OPEN) {
+    websocket._readyState = -2;
+  }
   websocket._receiver.end();
 }
 
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -123,7 +123,7 @@ class WebSocket extends EventEmitter {
    * @type {Number}
    */
   get readyState() {
-    return this._readyState;
+    return this._readyState === -2 ? WebSocket.CLOSING : this._readyState;
   }
 
   /**
@@ -159,6 +159,7 @@ class WebSocket extends EventEmitter {
     receiver.on('conclude', receiverOnConclude);
     receiver.on('drain', receiverOnDrain);
     receiver.on('error', receiverOnError);
+    receiver.on('finish', receiverOnFinish);
     receiver.on('message', receiverOnMessage);
     receiver.on('ping', receiverOnPing);
     receiver.on('pong', receiverOnPong);
@@ -201,31 +202,25 @@ class WebSocket extends EventEmitter {
   /**
    * Start a closing handshake.
    *
-   *          +----------+   +-----------+   +----------+
-   *     - - -|ws.close()|-->|close frame|-->|ws.close()|- - -
-   *    |     +----------+   +-----------+   +----------+     |
-   *          +----------+   +-----------+         |
-   * CLOSING  |ws.close()|<--|close frame|<--+-----+       CLOSING
-   *          +----------+   +-----------+   |
-   *    |           |                        |   +---+        |
-   *                +------------------------+-->|fin| - - - -
-   *    |         +---+                      |   +---+
-   *     - - - - -|fin|<---------------------+
-   *              +---+
-   *
    * @param {Number} [code] Status code explaining why the connection is closing
    * @param {String} [data] A string explaining why the connection is closing
    * @public
    */
   close(code, data) {
-    if (this.readyState === WebSocket.CLOSED) return;
     if (this.readyState === WebSocket.CONNECTING) {
       const msg = 'WebSocket was closed before the connection was established';
       return abortHandshake(this, this._req, msg);
     }
 
-    if (this.readyState === WebSocket.CLOSING) {
-      if (this._closeFrameSent && this._closeFrameReceived) this._socket.end();
+    //
+    // `this._readyState` below is not a typo. The
+    // `WebSocket.prototype.readyState` getter returns `WebSocket.CLOSING` even
+    // if `this._readyState` is `-2`.
+    //
+    if (
+      this._readyState === WebSocket.CLOSING ||
+      this.readyState === WebSocket.CLOSED
+    ) {
       return;
     }
 
@@ -238,7 +233,6 @@ class WebSocket extends EventEmitter {
       if (err) return;
 
       this._closeFrameSent = true;
-      if (this._closeFrameReceived) this._socket.end();
     });
 
     //
@@ -610,6 +604,15 @@ function initAsClient(websocket, address, protocols, options) {
   });
 
   req.on('upgrade', (res, socket, head) => {
+    //
+    // `tls.connect()` does not support the `allowHalfOpen` option in Node.js <
+    // 12.9.0. Also, the socket creation is not always under our control as the
+    // user might use the `http{,s}.request()` `agent` option and by default
+    // `net.connect()` and `tls.connect()` create a socket not allowed to be
+    // half-open.
+    //
+    socket.allowHalfOpen = true;
+
     websocket.emit('upgrade', res);
 
     //
@@ -679,6 +682,7 @@ function initAsClient(websocket, address, protocols, options) {
  * @private
  */
 function netConnect(options) {
+  options.allowHalfOpen = true;
   options.path = options.socketPath;
   return net.connect(options);
 }
@@ -691,6 +695,7 @@ function netConnect(options) {
  * @private
  */
 function tlsConnect(options) {
+  options.allowHalfOpen = true;
   options.path = undefined;
 
   if (!options.servername && options.servername !== '') {
@@ -820,6 +825,17 @@ function receiverOnError(err) {
  * @private
  */
 function receiverOnFinish() {
+  const websocket = this[kWebSocket];
+
+  if (!websocket._closeFrameReceived) websocket.close();
+}
+
+/**
+ * A listener for the `Receiver` `'error'` or `'finish'` event.
+ *
+ * @private
+ */
+function receiverOnErrorOrFinish() {
   this[kWebSocket].emitClose();
 }
 
@@ -893,8 +909,8 @@ function socketOnClose() {
   ) {
     websocket.emitClose();
   } else {
-    websocket._receiver.on('error', receiverOnFinish);
-    websocket._receiver.on('finish', receiverOnFinish);
+    websocket._receiver.on('error', receiverOnErrorOrFinish);
+    websocket._receiver.on('finish', receiverOnErrorOrFinish);
   }
 }
 
@@ -918,9 +934,13 @@ function socketOnData(chunk) {
 function socketOnEnd() {
   const websocket = this[kWebSocket];
 
-  websocket._readyState = WebSocket.CLOSING;
+  //
+  // This is a special state. It indicates that the connection is going through
+  // the closing handshake but unlike `WebSocket.CLOSING` it does not prevent
+  // `WebSocket.prototype.close()` from sending a close frame.
+  //
+  websocket._readyState = -2;
   websocket._receiver.end();
-  this.end();
 }
 
 /**
--- a/test/create-websocket-stream.test.js
+++ b/test/create-websocket-stream.test.js
@@ -204,6 +204,8 @@ describe('createWebSocketStream', () => {
 
     it('reemits errors', (done) => {
       let duplexCloseEventEmitted = false;
+      let serverClientCloseEventEmitted = false;
+
       const wss = new WebSocket.Server({ port: 0 }, () => {
         const ws = new WebSocket(`ws://localhost:${wss.address().port}`);
         const duplex = createWebSocketStream(ws);
@@ -218,6 +220,7 @@ describe('createWebSocketStream', () => {
 
           duplex.on('close', () => {
             duplexCloseEventEmitted = true;
+            if (serverClientCloseEventEmitted) wss.close(done);
           });
         });
       });
@@ -225,10 +228,11 @@ describe('createWebSocketStream', () => {
       wss.on('connection', (ws) => {
         ws._socket.write(Buffer.from([0x85, 0x00]));
         ws.on('close', (code, reason) => {
-          assert.ok(duplexCloseEventEmitted);
           assert.strictEqual(code, 1002);
           assert.strictEqual(reason, '');
-          wss.close(done);
+
+          serverClientCloseEventEmitted = true;
+          if (duplexCloseEventEmitted) wss.close(done);
         });
       });
     });
--- a/test/sender.test.js
+++ b/test/sender.test.js
@@ -13,9 +13,10 @@ class MockSocket {
     if (write) this.write = write;
   }
 
+  end() {}
   cork() {}
-  write() {}
   uncork() {}
+  write() {}
 }
 
 describe('Sender', () => {
--- a/test/websocket-server.test.js
+++ b/test/websocket-server.test.js
@@ -400,7 +400,9 @@ describe('WebSocketServer', () => {
 
         server.on('upgrade', (req, socket, head) => {
           wss.handleUpgrade(req, socket, head, (ws) => {
-            ws.close();
+            process.nextTick(() => {
+              ws.close();
+            });
           });
           assert.throws(
             () => wss.handleUpgrade(req, socket, head, NOOP),
--- a/test/websocket.test.js
+++ b/test/websocket.test.js
@@ -430,6 +430,8 @@ describe('WebSocket', () => {
   describe('Events', () => {
     it("emits an 'error' event if an error occurs", (done) => {
       let clientCloseEventEmitted = false;
+      let serverClientCloseEventEmitted = false;
+
       const wss = new WebSocket.Server({ port: 0 }, () => {
         const ws = new WebSocket(`ws://localhost:${wss.address().port}`);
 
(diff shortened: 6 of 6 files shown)


lpinca:
cc: @pimterry

This is base on our discussion in https://github.com/websockets/ws/pull/1899. There are some small things to discuss but I'm a bit in hurry. I'll list them tomorrow.

lpinca:
> There are some small things to discuss but I'm a bit in hurry. I'll list them tomorrow.

- This branch https://github.com/websockets/ws/blob/074e6a8be7275a69a407f6c1fa2270c754d2834b/lib/websocket.js#L238 was not covered by tests. It now is, but it is expected. The error happens because [`server.close()`](https://github.com/websockets/ws/blob/074e6a8be7275a69a407f6c1fa2270c754d2834b/doc/ws.md#serverclosecallback) terminate all clients with `websocket.terminate()` (destroying the socket) and the remote peer now tries to send a close frame on a destroyed socket when its socket emits the `'end'` event. It might make sense to change `server.close()` behavior so that it closes all connections gracefully with `websocket.close()`. This should eventually be done independently of this change.
- The `_closeFrameSent` property was needed because `socket.end()` was called after a close frame was both received and sent. `socket.end()` is now called right after the close frame is sent so the `_closeFrameSent` property is no longer needed for that purpose. I did not remove it because its value is used to compute the value of the `wasClean` property of the `CLoseEvent` https://github.com/websockets/ws/blob/074e6a8be7275a69a407f6c1fa2270c754d2834b/lib/event-target.js#L63 I've read the HTML WebSocket specification and tried to find some clue in the Web Platform Test but it is not clear to me when the `wasClean` attribute should be set to `true`.
- This patch is completely untested in real world scenarios. I don't know if it works as expected with other implementations of the WebSocket specification, for example in environments where there is no support for half-open sockets.

pimterry:
This is really interesting and looks very promising imo! Great work. I've been playing around with similar changes myself as well, and testing the code you posted before for Chrome's behaviour, to try and confirm how everything works here.

I'd quite like to try and finish putting some tests together to check exactly how the previous version behaved so we can check how that will change with these updates. In theory we should be able to simulate each of the possible open/half-open cases in tests from within node ourselves, which would make it much easier to be confident about the real-world impact.

I don't think there's any hurry here and I'm a bit squeezed for time right now, so I'm going to try and flesh those out first to explore this in a bit more depth, and come back to this with some testing and a proper review in a few days. Hope that's ok!

lpinca:
Sure, there is no hurry. Take your time.

pimterry:
I've just opened [a PR](https://github.com/websockets/ws/pull/1906) with some extra tests to explore all this.

I haven't done thorough real world testing, just lots of playing around with the unit tests, but those tests clearly show broken behaviour on master, and working behaviour on this branch, which seems like a great sign to me :+1:

On your points above:

> It might make sense to change server.close() behavior so that it closes all connections gracefully with websocket.close().

That makes sense to me too, and I agree it can be done separately :+1:

> I don't know if it works as expected with other implementations of the WebSocket specification, for example in environments where there is no support for half-open sockets.

I haven't extensively tested other clients, but the tests here do some basic simulation of every combination of half/full-close supported peers, and at the very least they all seem to be able to shut down cleanly in simple cases, so we can be confident they're not _completely_ broken.

> I've read the HTML WebSocket specification and tried to find some clue in the Web Platform Tests but it is not clear to me when the wasClean attribute should be set to true.

The WebSocket RFC 7.1.4 says:

>  If the TCP connection was closed after the WebSocket closing handshake was completed, the WebSocket connection is said to have been closed _cleanly_.
> If the WebSocket connection could not be established, it is also said that _The WebSocket Connection is Closed_, but not _cleanly_.

I think that's probably the clearest definition around: a websocket is cleanly closed only if we completed the handshake before the socket was fully closed, which AFAICT just requires that we sent and received a close frame.

lpinca:
I amended the first commit with a minor change:

There is no need to wait for `sender.close()` callback before calling `socket.end()`. We can call it right after the close frame is written to the socket in `sender.doClose()`.

```diff
diff --git a/lib/sender.js b/lib/sender.js
index ad71e19..18f0cf4 100644
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -151,6 +151,7 @@ class Sender {
       }),
       cb
     );
+    this._socket.end();
   }
 
   /**
diff --git a/lib/websocket.js b/lib/websocket.js
index 1b652ca..fbe4a67 100644
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -233,7 +233,6 @@ class WebSocket extends EventEmitter {
       if (err) return;
 
       this._closeFrameSent = true;
-      this._socket.end();
     });
 
     //
```

lpinca:
I'm a bit torn on how to handle this change.

- It will certainly create issues if a ws client with this change connects to a ws server without this change, and vice versa.
- If added in next major version it will create even more issues as many people (for good reasons) do not upgrade to a major version or do it very slowly.
- I still don't know if it causes issues with non-ws clients or servers.

I think the best option is to include this in a minor release.

@cTn-dev, sorry for pinging you here, but If I remember correctly you were using permessage-deflate in production. Is there any chance you can help us testing this? Thank you.

pimterry:
I think we're actually being too aggressive with closure here I'm afraid. It's correct from a TCP POV to half-close as soon as we've sent everything, but the fully correct websockets behaviour is to close after a close frame is both sent + received, or to close immediately after sending a close frame only when we hit an error.

We do still want some of the previous behaviour! We don't want to remove that completely.

A bit of googling suggests some other servers will indeed fail in this case if we're connecting to them from node (e.g. when connecting to Akka servers, although they're quite apologetic about it: https://doc.akka.io/docs/akka-http/current/client-side/websocket-support.html#half-closed-client-websockets).

Relevant bits of the spec:

> Once an endpoint has both sent and received a Close control frame, that endpoint SHOULD _Close the WebSocket Connection_ 

(i.e. _after_ we have both close frames we should start half-closing to initiate TCP shutdown)

> If _The WebSocket Connection is Established_ prior to the point where the endpoint is required to _Fail the WebSocket Connection_, the endpoint SHOULD send a Close frame with an appropriate status code (Section 7.4) before proceeding to _Close the WebSocket Connection_.

(i.e. after an error we should send a close frame and only then _immediately_ half-close to initiate TCP shutdown)

Sorry, I think this is my fault :grimacing:. My explanation in #1899 wasn't clear and I think I've confused myself slightly en route too. This change is still required, but only for the case where we send a close frame after an error I think. Helpfully that does reduce the size of this change and the potential impact though!

cTn-dev:
> I'm a bit torn on how to handle this change.
> 
> * It will certainly create issues if a ws client with this change connects to a ws server without this change, and vice versa.
> * If added in next major version it will create even more issues as many people (for good reasons) do not upgrade to a major version or do it very slowly.
> * I still don't know if it causes issues with non-ws clients or servers.
> 
> I think the best option is to include this in a minor release.
> 
> @cTn-dev, sorry for pinging you here, but If I remember correctly you were using permessage-deflate in production. Is there any chance you can help us testing this? Thank you.

Sure, i can take it for a spin, is there a specific configuration/scenario you would like me to focus on/emulate?

lpinca:
> (i.e. after we have both close frames we should start half-closing to initiate TCP shutdown)

Yes, that is my understanding as well and the reason why it is currently implemented is this way.

> This change is still required, but only for the case where we send a close frame after an error I think. Helpfully that does reduce the size of this change and the potential impact though!

I still think that the easiest way to follow the spec recommendation of sending a close frame after an error is a best effort approach like the one suggested in https://github.com/websockets/ws/issues/1892#issuecomment-850287052.

@cTn-dev no particular configuration. The ideal scenario would involve some incoming and/or outgoing data buffering on both peers and then see if there is data loss in any of the peer after the connection is closed, for example by monitoring their close code.

pimterry:
> I still think that the easiest way to follow the spec recommendation of sending a close frame after an error is a best effort approach like the one suggested in #1892 (comment).

The problem with using `destroy()` rather than a clean TCP shutdown seems to be that it will send RST packets which may cause loss of pending good data in the remote peer unnecessarily. From the spec:

> on some platforms, if a socket is closed with data in the receive queue, a RST packet is sent, which will then cause recv() to fail for the party that received the RST, even if there was data waiting to be read.

I think that means if we send a RST (due to pending incoming data on these platforms, or any incoming data that's already in transit) then the remote peer will lose data that was successfully sent before the error but which hadn't been read yet, and/or the close frame itself.

It's actually a bit worse in our case I think, because we might have data still being compressed in the sender which would block the close frame. That means the destroy() in your example would be called _before_ we'd sent the close frame or pending data at all, so all of that would always be lost.

Aside from potential data loss, it'll also result in unnecessary socket errors on the remote peer, instead of a clean close frame + TCP shutdown, which is a mildly annoying for everybody.

I agree it's easier to just slam everything shut, but it is cleaner and more correct to `end()` as you added in #1899, and if we can do that I think we should. I'm pretty sure it's possible to do without extra issues, we just need to cover the last edge cases that are left over to make sure we always wait appropriately and make sure everything is tidied up properly.

I think doing so within this PR looks something like:

* Keep the sent/received checks we had before.
* Do `end()` immediately after a error close frame when the websocket connection fails (but not after normal close frames, and not with a `destroy()` immediately).
* Keep all the other new changes here which add support for half-open sockets, so we cleanly shut down in that case too, and so that we handle receiver 'finish' & socket 'end' cleanly.

My understanding is that in that case:

* Normal shutdown behaviour is more or less the same as before, so it's a pretty safe change.
* We fix the current issue on master where a remote half-close means we immediately `end` and so lose all pending outgoing data.
* We fix the edge case issue on master, created by #1899, where simultaneous errors or similar issues mean we don't shutdown the socket at all until timeout. Now we would always end() after errors, so we'd only need the destroy() timeout if we get no socket error but the remote peer doesn't respond at all, which is the case that the timeout is designed for.

Does that make sense and sound sensible to you?

lpinca:
> Does that make sense and sound sensible to you?

Yes.

cTn-dev:
> > (i.e. after we have both close frames we should start half-closing to initiate TCP shutdown)
> 
> Yes, that is my understanding as well and the reason why it is currently implemented is this way.
> 
> > This change is still required, but only for the case where we send a close frame after an error I think. Helpfully that does reduce the size of this change and the potential impact though!
> 
> I still think that the easiest way to follow the spec recommendation of sending a close frame after an error is a best effort approach like the one suggested in [#1892 (comment)](https://github.com/websockets/ws/issues/1892#issuecomment-850287052).
> 
> @cTn-dev no particular configuration. The ideal scenario would involve some incoming and/or outgoing data buffering on both peers and then see if there is data loss in any of the peer after the connection is closed, for example by monitoring their close code.

So, as of right now i don't really have a straight forward way to make sure something is in the buffers on each side.
I tried a custom local build (while comparing the sent vs received close code) which periodically opens and closes client connections.

The close event is returning a pretty frequent 1006 instead of the expected/client sent 1000.
This definitely wasn't the case on 7.4.6 or 7.5.0.

These early results should be taken with a grain of salt since the websocket server could be at "fault" here, i will have to dig deeper to see what is going on here.

lpinca:
@pimterry 

> - We fix the current issue on master where a remote half-close means we immediately end and so lose all pending outgoing data.

There is actually a problem with this:

1. A peer calls `socket.end()` without sending the closing frame.
1. The `'end'` event is emitted by the socket of the other peer.
1. The ready state is changed to `-2` and `receiver.end()` is called.
1. The `receiver` then emits `'finish'`. A close frame was not received so `websocket.close()` is called.
1. A close frame is sent to the peer that originally called `socket.end()` at point 1.
1. The callback of `sender.close()` is called. `socket.end()` is now called only if the `receiver` errored or if a [close frame was receiver](https://github.com/websockets/ws/blob/145480a5b520ee951d848009d51069bfd7ed928c/lib/websocket.js#L241) but neither of these conditions is true.

Does it make sense?

I'm not sure how to fix this if we want to keep the current behavior. If `socket.end()` should be called only after a close frame is sent and received, then receiving an `'end'` event without a close frame means that the other peer is not behaving correctly.

pimterry:
> receiving an 'end' event without a close frame means that the other peer is not behaving correctly.

Ok yes, I think you're right, that's basically true.

In happy behaviour with a clean close, they MUST always send a close frame first, that's quite clear.

If there's some kind of error somewhere, they might not, and they're still technically in the rules. It's encouraged (in [7.1.7](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.7): after errors they SHOULD send a close frame before starting to close the connection, they SHOULD NOT close the connection unexpectedly for other reasons) but not a strict MUST anywhere. If they half-shutdown a socket though they are doing something weird, and it seems very likely that either we or them have made an error at this point which is why the connection is being closed, which probably means they're ignoring anything we send anyway.

It's still important that we finish processing any inputs they sent before closing the socket (so we should accept any pending compressed or close frames) but AFAICT we do do that correctly already. I guess it seems reasonable to just immediately close our end of the connection at that point too, to dump any pending data, and call it an unclear 1006 closure.

> There is actually a problem with this

I think we _could_ work around the case you've got there with something like 'after the close frame is sent, if the socket is half-shutdown already then just end', but given the above I think you're totally right and we shouldn't worry about sending close frames and pending data in this case.

lpinca:
> I think we could work around the case you've got there with something like 'after the close frame is sent, if the socket is half-shutdown already then just end'

Yes, I thought about that. We can do

```js
if (
  this._closeFrameReceived ||
  this._receiver._writableState.finished ||
  this._receiver._writableState.errorEmitted
) {
  this._socket.end();
}
```

in the callback of `sender.close()` but it means breaking the "`socket.end()` on close frame sent and received" contract.

lpinca:
Even with that check in place, a dead lock can be created if the user calls `websocket.close()` in the timeframe that goes from the socket `'end'` event and the receiver `'finish'` event. In  this case, the user initiated `websocket.close()` would not call `socket.end()` because none of the above conditions is true and the `'finish'` initiated `websocket.close()` would not call  `socket.end()` because the ready state is `CLOSING` (2) but no close frame was received.

This is not an issue in the receiver error case because `websocket.close()` is called synchronously when the `'error'` event is emitted.

I think I'm fine dropping buffered outgoing data and sending no close frame is the other peer ends the socket without sending a closing frame.

pimterry:
> I think I'm fine dropping buffered outgoing data and sending no close frame is the other peer ends the socket without sending a closing frame.

I'm convinced, yeah this seems disproportionately complicated :+1:

lpinca:
@pimterry besides the "simultaneous errors" there is another condition that results in a dead lock in master and is not addressed by adding `|| this._receiver._writableState.errorEmitted` to this `if` condition https://github.com/websockets/ws/blob/145480a5b520ee951d848009d51069bfd7ed928c/lib/websocket.js#L241

The problem arises when

- A peer is closing the close frame was sent but not received
- The peer now receives an invalid frame.

To fix that we also need to change this
https://github.com/websockets/ws/blob/145480a5b520ee951d848009d51069bfd7ed928c/lib/websocket.js#L228

to something like `if (this._closeFrameSent && (this._closeFrameReceived || this._receiver._writableState.errorEmitted))`.

I also tested how browsers handle the "socket end with no close frame" and the "error after open" cases.

- In the former case both Safari and Chrome immediately close the connection without sending back a close frame.
- In the latter Safari immediately closes the connection without sending back a close frame. Chrome sends a close frame but only if there is no buffered outgoing data.

```js
const http = require('http');

const data = `<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <script>
      (function () {
        const ws = new WebSocket('ws://localhost:8080');

        ws.onopen = function () {
          console.log('open');
        };

        ws.onclose = function (evt) {
          console.log(evt.code);
        };
      })();
    </script>
  </body>
</html>`;

server.on('request', function (request, response) {
  response.setHeader('Content-Type', 'text/html');
  response.end(data);
});

  const key = crypto
    .createHash('sha1')
    .digest('base64');

  socket.on('data', function() {
    socket.write('ECONNRESET?');
  });

  socket.write(
    [
      'Upgrade: websocket',
      'Connection: Upgrade',
      '\r\n'
    ].join('\r\n')
  );

  process.once('SIGINT', function () {
    console.log('Sending an invalid frame');
    socket.write(Buffer.from([0x85, 0x00]));
  });
});

  console.log('Listening on *:8080');
});
```

The script above triggers an `ECONNRESET` error when using Chrome on macOS as client. However I see no RST with tcpdump.

Finally, `socket.write()` followed by `socket.destroy()` triggers a RST only if the other peer writes to the socket. Yes, I'm still thinking that https://github.com/websockets/ws/issues/1892#issuecomment-850287052 is the best thing to do. We are in an error condition so there is a good change that the other peer is malfunctioning. We should not wait for all buffered data to be written, it might never happen given that the peer is broken, just try to send the close frame which might or might not be received.

lpinca:
> The script above triggers an `ECONNRESET` error when using Chrome on macOS as client. However I see no RST with tcpdump.

I see it on Windows.

rst

Tested with IE 11 19043.1081, Firefox 89.0.2 and Chrome 91.0.4472.114.

pimterry:
> socket.write() followed by socket.destroy() triggers a RST only if the other peer writes to the socket.

That should be true, I agree, but as documented in the spec, some platforms will RST anyway in some cases, e.g. if there is more data waiting in our receive queue. Reports like https://github.com/libuv/libuv/issues/3034 link to various specific cases where Windows will send RSTs in node if you try to shut down too aggressively.

Even without that, it's quite possible that the remote peer has data in flight when we `destroy()`, in which case we'll send a RST on all platforms if that data arrives after `destroy()`.

This will result in unnecessary data loss and connection errors for the remote peer, and if we can we should close things properly to avoid that.

As far as I can tell, #1908 looks great though, and does resolve everything without the larger much riskier changes from this PR. With that PR merged, AFAICT:

* We always try to send a close frame when we get an error.
* We then wait for either a close response, socket end, or error, and then cleanly shut down the socket.
* We now _always_ shut down the socket, so we only need the timeout if the remote peer does nothing at all after we send them an error close frame - ignoring the error close frame entirely and keeping the connection open. Even in that case, the socket still always gets destroyed after 30 seconds.
* We never unnecessarily lose pending data in the remote peer (which could happen if we create a RST), and we don't unnecessarily lose pending outgoing data either.

> We should not wait for all buffered data to be written, it might never happen given that the peer is malfunctioning, just try to send the close frame which might or might not be received.

I don't really mind if we do drop other pending outgoing data, if you want, by sending the close frame immediately instead of adding it to the queue. The spec doesn't define this either way that I can see, and it doesn't seem unreasonable if that makes our lives simpler.

I do think we should not destroy the socket immediately though, since everything explicitly says that we shouldn't because it can cause real problems, and #1908 means everything is now cleaned up properly regardless.

What are the reasons you'd prefer to destroy() immediately instead of doing a clean TCP shutdown? From my POV: the spec explicitly says we should wait, I don't see any substantial downside, and it avoids the risk of data loss of both already sent data and the error close frame.

I'm not sure browsers are good benchmarks for behaviour here by the way - imo server to client error feedback is usually much more important & useful than the opposite, so the tradeoffs for WS are quite different to those for a web browser. As a developer building a websocket API, I really want my API to tell bad clients what they're doing wrong, so they can change their broken implementations! As a user browsing the internet or a developer building an API client, I'm usually not that bothered.

And in general of course, just because Safari is poorly behaved doesn't mean we should be too :smiley:

Issue #1886, opened 2021-05-20
Distribute ESM module in addition to CJS


feljx:
Would it be possible to distribute an ESM module version in addition to the current CommonJS version?
I know it's not a clean solution (gotta use "main", "module" and "exports" fields in package.json to work with all setups), but more and more libraries are distributing both so as to enable users to switch over to ESM.

lpinca:
Yes, probably, but I see no upsides, only downsides.

lpinca:
For example we could use an ES module wrapper as documented here https://nodejs.org/api/packages.html#packages_approach_1_use_an_es_module_wrapper.

```diff
$ git diff --cached
diff --git a/package.json b/package.json
index c2d63af..360ac2d 100644
--- a/package.json
+++ b/package.json
@@ -16,6 +16,10 @@
   "author": "Einar Otto Stangvik <einaros@gmail.com> (http://2x.io)",
   "license": "MIT",
   "main": "index.js",
+  "exports": {
+    "import": "./wrapper.mjs",
+    "require": "./index.js"
+  },
   "browser": "browser.js",
   "engines": {
     "node": ">=8.3.0"
@@ -23,7 +27,8 @@
   "files": [
     "browser.js",
     "index.js",
-    "lib/*.js"
+    "lib/*.js",
+    "wrapper.mjs"
   ],
   "scripts": {
     "test": "nyc --reporter=lcov --reporter=text mocha --throw-deprecation test/*.test.js",
diff --git a/wrapper.mjs b/wrapper.mjs
new file mode 100644
index 0000000..8df8956
--- /dev/null
+++ b/wrapper.mjs
@@ -0,0 +1,9 @@
+import createWebSocketStream from './lib/stream.js';
+import Receiver from './lib/receiver.js';
+import Sender from './lib/sender.js';
+import WebSocket from './lib/websocket.js';
+import WebSocketServer from './lib/websocket-server.js';
+
+export { createWebSocketStream, Receiver, Sender, WebSocket, WebSocketServer };
+
+export default WebSocket;
```

but this might introduce issues with bundles and would break code that uses the library like this

```js
const createWebSocketStream = require('ws/lib/stream');
```

At the very least we should bump the major version and I'm not sure if it makes sense when this

```js
import WebSocket from 'ws';
```

already works without changes.

jimmywarting:
i vote for ditching commonjs.

> but this might introduce issues with bundlers and would break code that uses the library like this

[...]

PR #1904, opened 2021-06-18
Breaking changes

lpinca:
I'm opening this PR to get a feedback on some breaking changes that I have been delaying for a while and planning to release in a new major version soon.


lpinca:
I'm not sure if a02969d makes sense. It is good for performance but adapting existing code to this change might require a little effort. Users have call `buffer.toString()` themselves:

```js
ws.on('message', function (data, isBinary) {
  const message = isBinary ? data : data.toString();
  // Continue as before.
});

ws.on('close', function (code, data) {
  const reason = data.toString();
  // continue as before.
});
```

If anyone is watching this, please share your opinion.

jimmywarting:
> I'm not sure if ab7b6d4 makes sense. It is good for performance but adapting existing code to this change might require a little effort. Users have call buffer.toString() themselves

I say: emit a ArrayBuffer (or Blob) & follow the spec instead do not emit Buffer or strings

lpinca:
@jimmywarting https://github.com/websockets/ws/commit/ab7b6d4b4376dd1a3515b31aad2c58caa851dd92 is about using `Buffer`s instead of strings for text messages and for the close reason, not for binary messages. This is against the HTML specification where a `DOMString` is used and it is done for performance. For binary messages, users can already get an `ArrayBuffer` by setting the `binaryType` attribute to `'arraybuffer'` as per HTML spec.

> If you are planing on making a breaking change anyway, what do you feel about switching to ESM? node "^12.20.0 || ^14.13.0"

See https://github.com/websockets/ws/issues/1886. I'm fine with that as long as it does not cause issues with bundlers.

> While we are at it, go for Event and EventTarget that exist in NodeJS now

This requires Node.js >= 14.5.0 and comes with significant performance regressions. This is currently a no go.

> Fyi, Node ships with experimental Blob support now now (v15.7) also

I know and it might make sense to change the default `binaryType` from `'nodebuffer'` to `'blob'` as per HTML spec but the feature is experimental and I don't want to use polyfills. I prefer to postpone this to the next major release.

lpinca:
I really need feedbacks for 4edae51. If it is too breaking I will remove it.

cc: @cTn-dev @pimterry @netizen-ais @feross

cTn-dev:
This is definitely one hell of a breaking change.

While i am in favor of removing the "implicit" `.toString()` and let the user handle that logic (as already pointed out in some cases, it would be pretty invisible as far `JSON.parse()` is concerned.

Have the removal of the `.toString()` resulted in any significant performance improvement ?
I don't currently have an opinion on the `isBinary` argument (will have to think about it some more)

lpinca:
> Have the removal of the `.toString()` resulted in any significant performance improvement ?

Yes, especially when using the streaming interface. It prevents the data from being copied twice (from `Buffer` to string when a text message is received and back from string to `Buffer` when it is pushed into the `Duplex` readable queue) here https://github.com/websockets/ws/blob/7.5.2/lib/stream.js#L76.

netizen-ais:
My environment isn't big enough to get a hint about performance (2 ws servers with an average of 100 clients that connect to both node processes), but after testing this branch in the develop environment with no code changes on our base, I've tested it in one of the ws servers without any problem.

package.json
```diff
@@ -16,14 +16,19 @@
...
   "license": "MIT",
   "main": "index.js",
+  "exports": {
+    "import": "./wrapper.mjs",
+    "require": "./index.js"
+  },
   "browser": "browser.js",
   "engines": {
-    "node": ">=8.3.0"
+    "node": ">=10.0.0"
```
mcollina, line 25: I would recommend to make the jump to v12.

mcollina (commented):
I'm good with the changes!

package.json
lpinca: v10 is officially no longer supported but only two months passed since its EOL date. I prefer to wait a little longer before dropping support for it, there is no additional maintenance burden compared to v12.

pimterry:
Changes all make sense to me, definitely required quite a few updates on my end so it is a bit disruptive, but all very small and easy to do, and everything's still working nicely afterwards :+1:

cTn-dev:
I did a bit of testing against the 10f27d7a14e5e10ab45b458ac84162c60361c806 with both string and binary messages (haven't tried a mixing them in a single session).
Everything is working fine, the "migration" due to the breaking changes took < 5 minutes, as straight forward as it gets.

jberger:
I've just hit something that is obliquely related. I can open a new bug (and if it gets ignored here I will) but currently it seems to be impossible to write to a websocket stream as text. Any idea if you can make the writeableObjectMode useful while you make the readableObjectMode useful too?

(This affects me because the code that controls the websocket in the browser side assumes that it gets text and I have no control over that code https://github.com/glyptodon/guacamole-client/blob/754e9649f1fa0ba225ee42b56ded64bc283d17df/guacamole-common-js/src/main/webapp/modules/Tunnel.js#L1009)

lpinca:
> I can open a new bug (and if it gets ignored here I will) but currently it seems to be impossible to write to a websocket stream as text.

@jberger this is unrelated to this PR so yes, a new issue would have been better. Anyway you can use the [`decodeStrings`](https://nodejs.org/api/stream.html#stream_implementing_a_writable_stream) option.

```js
import { createWebSocketStream, WebSocket } from 'ws';

const ws = new WebSocket(url);
const duplex = createWebSocketStream(ws, { decodeStrings: false });

duplex.write('foo');
duplex.write('bar');
duplex.write('baz');
```

lpinca:
@pimterry the last two commits are based on our discussion in https://github.com/websockets/ws/pull/1902. Can you please take a look when you have some time? Thank you.

lib/websocket-server.js
```diff
@@ -151,29 +155,41 @@ class WebSocketServer extends EventEmitter {
...
+      if (this._server) {
+        this._removeListeners();
+        this._removeListeners = this._server = null;
+      }
+
+      if (this.clients) {
+        if (!this.clients.size) {
+          process.nextTick(emitClose, this);
+        } else {
+          for (const client of this.clients) client.close(1001);
```
pimterry, line 174: From a networking POV, this all seems like a good improvement to me. It definitely seems clearer for clients if servers send a 1001 close frame and tidily shutdown connections where possible :+1:

That said, it does mean the connections could now unexpectedly stay open whilst waiting for a close response, until the close timeout, if you have any poorly behaving clients. That might be problematic if users are trying to _immediately_ shut everything down, which seems like a reasonable use case, and there's now no easy alternative way to do that when necessary.

Maybe the server should have a `server.terminate()` method too, which does terminate everything immediately?

That would make the server API match the individual websockets API, which already has `websocket.terminate()`, and it would cover that immediate-shutdown use case.

As a bonus, this could also easily support cases like "try to shut down cleanly, but then always close within X milliseconds" with just:

```javascript
server.close();
setTimeout(() => server.terminate(), X);
```

lpinca: Users can terminate clients by themself if they want to close immediately:

```js
wss.close();
for (const ws of wss.clients) client.terminate();
```

pimterry: Ah ok, fair enough. I'm not sure if that trick will be obvious to people, so a convenience `server.terminate()` method to do the same could still be useful, but you're right that that means it's not strictly necessary :+1:

It's probably worth mentioning this in the breaking-changes release notes though imo, as the suggested way to migrate for people who do need the previous behaviour.

lpinca: I have two doubts:

1. If it is better to just prevent new connections from being established and not close/terminate existing clients. Closing clients is user responsibility so they can use `websocket.close()` or `websocket.terminate()` at will. This is also consistent with the behavior of [`server.close()`](https://nodejs.org/api/net.html#net_server_close_callback)
1. Regardless of 1., if/when to emit the `'close'` event when client tracking is disabled. This is currently done in the next tick.

lpinca: I went ahead and changed `WebSocketServer.prototype.close()` behavior in https://github.com/websockets/ws/pull/1904/commits/6a238da218d44ef4e68177dae1d54280bd8eeee8. It no longer closes existing connections as per point 1. of my previous comment.

I've also updated the documentation for point 2. I think emitting `'close'` when connections are still active is a bit misleading but this behavior already exists and was not changed here.

pimterry: Missed this sorry, I've been busy. That sounds very reasonable to me and the consistency with node is great.

I do think documenting this clearly for those who do want to terminate the server will be useful though, and it's not obvious right now. It might be worth a "how to fully shutdown a server" section in the readme, and/or the release notes.

animaonline:
We handle binary and text messages differently, now that there is now way to distinguish the type of the message on the server side we're unable to use version 8.0.0, please add isBinary to the server api.

PR #1906, opened 2021-06-22
Add some extra closing handshake tests

pimterry:
This is an attempt to test some of the behaviour discussed in https://github.com/websockets/ws/pull/1899#discussion_r647348110 and test how that's improved with the changes currently pending in #1902.

The tests are a bit rough in places, feel free to tweak or drop some of them if you think they aren't relevant. When run against master, 2 of these tests currently fail:

* "cleanly shuts down after simultaneous errors"
* "handles socket shutdown while compressing outgoing frames"

Both of these are current bugs on master: simultaneous errors (or anything else that means the remote peer doesn't respond) will break clean connection shutdown so the socket stays fully open until timeout, and also a remote `ws.end()` that starts socket shutdown immediately drops all pending outgoing data, incorrectly & unnecessarily.

Various tests here fail in interesting ways given the partial fixes I suggested in that PR thread, basically as predicted there. When run against the new PR with the full set of changes though, all the tests do happily pass, which is great!

This tries to cover the shutdown for the various half-open combinations too, by directly simulating no-half-open behaviour rather than using allowHalfOpen=false, so that it's still effective even when the new PR resets that. It's not a super rigorous real-world test by any means, and it doesn't test various possible data loss cases for each combination, but from this alone it seems like the basic behaviour is working great :+1:

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -123,7 +123,7 @@ class WebSocket extends EventEmitter {
    * @type {Number}
    */
   get readyState() {
-    return this._readyState;
+    return this._readyState === -2 ? WebSocket.CLOSING : this._readyState;
   }
 
   /**
@@ -159,6 +159,7 @@ class WebSocket extends EventEmitter {
     receiver.on('conclude', receiverOnConclude);
     receiver.on('drain', receiverOnDrain);
     receiver.on('error', receiverOnError);
+    receiver.on('finish', receiverOnFinish);
     receiver.on('message', receiverOnMessage);
     receiver.on('ping', receiverOnPing);
     receiver.on('pong', receiverOnPong);
@@ -201,31 +202,25 @@ class WebSocket extends EventEmitter {
   /**
    * Start a closing handshake.
    *
-   *          +----------+   +-----------+   +----------+
-   *     - - -|ws.close()|-->|close frame|-->|ws.close()|- - -
-   *    |     +----------+   +-----------+   +----------+     |
-   *          +----------+   +-----------+         |
-   * CLOSING  |ws.close()|<--|close frame|<--+-----+       CLOSING
-   *          +----------+   +-----------+   |
-   *    |           |                        |   +---+        |
-   *                +------------------------+-->|fin| - - - -
-   *    |         +---+                      |   +---+
-   *     - - - - -|fin|<---------------------+
-   *              +---+
-   *
    * @param {Number} [code] Status code explaining why the connection is closing
    * @param {String} [data] A string explaining why the connection is closing
    * @public
    */
   close(code, data) {
-    if (this.readyState === WebSocket.CLOSED) return;
     if (this.readyState === WebSocket.CONNECTING) {
       const msg = 'WebSocket was closed before the connection was established';
       return abortHandshake(this, this._req, msg);
     }
 
-    if (this.readyState === WebSocket.CLOSING) {
-      if (this._closeFrameSent && this._closeFrameReceived) this._socket.end();
+    //
+    // `this._readyState` below is not a typo. The
+    // `WebSocket.prototype.readyState` getter returns `WebSocket.CLOSING` even
+    // if `this._readyState` is `-2`.
+    //
+    if (
+      this._readyState === WebSocket.CLOSING ||
+      this.readyState === WebSocket.CLOSED
+    ) {
       return;
     }
 
@@ -238,7 +233,7 @@ class WebSocket extends EventEmitter {
       if (err) return;
 
       this._closeFrameSent = true;
-      if (this._closeFrameReceived) this._socket.end();
+      this._socket.end();
     });
 
     //
@@ -610,6 +605,15 @@ function initAsClient(websocket, address, protocols, options) {
   });
 
   req.on('upgrade', (res, socket, head) => {
+    //
+    // `tls.connect()` does not support the `allowHalfOpen` option in Node.js <
+    // 12.9.0. Also, the socket creation is not always under our control as the
+    // user might use the `http{,s}.request()` `agent` option and by default
+    // `net.connect()` and `tls.connect()` create a socket not allowed to be
+    // half-open.
+    //
+    socket.allowHalfOpen = true;
+
     websocket.emit('upgrade', res);
 
     //
@@ -679,6 +683,7 @@ function initAsClient(websocket, address, protocols, options) {
  * @private
  */
 function netConnect(options) {
+  options.allowHalfOpen = true;
   options.path = options.socketPath;
   return net.connect(options);
 }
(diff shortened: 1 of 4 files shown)


pimterry:
I was testing in Node 14 locally - works great there but it turns out there _are_ test failures in 8, 10 & 12. I'll look into those now.

pimterry:
Ok, I think I've fixed this. This was an actual bug in the PR that the tests found, affecting older node versions only.

In Node older than v14, after a receiver error, the receiver still fires a 'finish' event when the socket itself ends. In Node v14+ it seems that this doesn't happen. No idea why, but this causes the bug.

In all cases: when the receiver hits the error, we close() the websocket with an error code, we `resume()` to ignore all future socket input, and we wait for the other end before shutting down the socket. At this point we're CLOSING.

Unfortunately later when socket actually ends, we set _readyState back to -2 (the new 'PRECLOSING' state where we're still allowed to send a close frame) even though we've already sent a close frame.

That wouldn't normally break anything, and it doesn't with Node v14+. With older node though, we get an extra 'finish' event. When that fires it tries to close the websocket, and it's allowed to close it again, even though we've already closed it, because the readyState is no longer CLOSING and it's not yet CLOSED. This double-close then throws write-after-end on the socket.

This is fixed here by not setting _readyState to -2 unless the connection is currently OPEN, so you can't go from CLOSING back to -2 again. Does that seem reasonable?

lib/websocket.js
```diff
@@ -940,7 +940,9 @@ function socketOnEnd() {
   // the closing handshake but unlike `WebSocket.CLOSING` it does not prevent
   // `WebSocket.prototype.close()` from sending a close frame.
   //
-  websocket._readyState = -2;
+  if (websocket._readyState === WebSocket.OPEN) {
```
lpinca, line 943: Why not removing the `receiverOnFinish()` listener from `receiverOnError()` instead?

lpinca:
Good catch. This confirms your finding

```js

    process.nextTick(function () {
      callback(new Error('Oops'));
    });
  }
});

writable.on('error', function () {
  writable.end();
});

writable.on('finish', function () {
  throw new Error("The 'finish' event was emitted after the 'error' event");
});

writable.write('foo');
```

It seems it was fixed by https://github.com/nodejs/node/commit/ba3be578d8 included in Node.js 13.

lpinca:
I've cherry picked https://github.com/websockets/ws/pull/1906/commits/104c4d3daf0fea43e55860119a99b1755ee23ee3 into the simplify/closing-handshake branch.

pimterry:
I didn't notice before, but the Mac & Windows tests were still running, not passing. They've since timed out, because they were actually failing & stuck in that same parsing error test too.

In their case the failure is different, and I think it's acceptable behaviour, so I've updated the test.

The cause is that on Windows & Mac, unlike Linux, during this error shutdown the event order is:

* Receiver error -> close()
* Error close frame written, client calls socket.end()
* Server receives close frame, server replies
* Response close frame written, server calls socket.end()
* Server socket 'end' event
* Server socket 'close' event, so server emits 'close'
* Client socket 'end' event
* Client socket 'close' event, so client emits 'close'

Meanwhile on Linux, I see:

* Receiver error -> close()
* Error close frame written, client calls socket.end()
* Server receives close frame, server replies
* Response close frame written, server calls socket.end()
* Server socket 'end' event
* Client socket 'end' event
* Client socket 'close' event, so client emits 'close'
* Server socket 'close' event, so server emits 'close'

I.e. the order of server/client 'close' events is different. I suspect on Windows & Mac the 'close' event fires synchronously if the other end has already ended, while on Linux it's async, or there's some other step involved.

The test expected the client to always fire 'close' first because I tested it on Linux, but actually I don't _think_ it really matters. Either way the test checks that both sides have sent close frames and closed their sockets without error, so I think it's fine either way. Does that sound right to you?

For now I've updated the test to allow 'close' to fire in either order, as long as both sides get their close frames with no errors, and it seems to work nicely now.

test/websocket.test.js
```diff
@@ -2763,4 +2765,403 @@ describe('WebSocket', () => {
...
+          perMessageDeflate: { threshold: 0 }
+        },
+        () => {
+          const ws = new WebSocket(`ws://localhost:${wss.address().port}`, {
+            perMessageDeflate: { threshold: 0 }
+          });
+
+          ws.on('open', () => {
+            // Add 200ms delay to compression:
+            makeDeflateSlow(ws, 200);
```
lpinca, line 2877: Can we avoid using timers? We can push the data directly to the socket without relying on the other peer, similarly to how it is done here https://github.com/websockets/ws/blob/145480a5b520ee951d848009d51069bfd7ed928c/test/create-websocket-stream.test.js#L445 and then assert that `sender._deflating === true`.

test/websocket.test.js
```diff
@@ -2763,4 +2765,403 @@ describe('WebSocket', () => {
...
+          });
+        }
+      );
+
+      wss.on('connection', (ws) => {
+        ws.send('hello', { compress: true }); // This will take the client 200ms to decompress
+      });
+    });
+
+    it('handles connection closures over both half-open-allowed sockets', (done) => {
```
lpinca, line 3056: I think this and the next tests are not very useful because no data is buffered/compressed/decompressed. They are not testing edge cases.

lpinca:
> Either way the test checks that both sides have sent close frames and closed their sockets without error, so I think it's fine either way. Does that sound right to you?

Yes, I did the same in https://github.com/websockets/ws/pull/1902.

test/websocket.test.js
```diff
@@ -2763,4 +2765,401 @@ describe('WebSocket', () => {
...
+            makeDeflateSlow(ws, 200);
+
+            ws.send('hello', { compress: true });
+          });
+        }
+      );
+
+      wss.on('connection', (ws) => {
+        // Server closes after 100ms, before the client finishes sending its compressed message:
+        setTimeout(() => ws.close(1000), 100);
```
lpinca, line 2886: Also, this behaves differently depending on where the test is run:

1. On the master branch this just calls `socket.write(closeFrame);`
1. On this branch it calls `socket.write(closeFrame)` and `socket.end()`.

The test should be independent from `websocket.close()` behavior. Should we test case 1, 2, or both?

lpinca:
I cherry picked fad87e333c36ee2f166eb497fb5f115cf5f682f8 and tweaked it in https://github.com/websockets/ws/pull/1902/commits/f713a6efce7f6eedcc80df139e7ebd545c573f26.

pimterry:
> I cherry picked fad87e3 and tweaked it in f713a6e.

Great! Sorry I haven't had time to work through the comments above myself. I think that's fine though, these tests were mostly useful as a way to explore how the old & new behaviour differ and confirm that we've fixed the original issues. If you've pulled in the useful ones then we can close this now.

PR #1910, opened 2021-07-05
minor jsdoc improvements

jimmywarting:
I use this settings on by default in VS code
```
    "javascript.suggest.completeFunctionCalls": true,
    "javascript.suggest.completeJSDocs": true,
    "javascript.suggest.autoImports": true,
    "js/ts.implicitProjectConfig.checkJs": true,
```

It was showing red lines a bit here and there. With this changes it could figure out more what things where with autocompletion

Could improve some bits even further if i may use private fields, But it requires node v12 and you are testing v8...
could we perhaps drop some older versions and align a bit more with current targeted support versions?

### DIFF
--- a/lib/receiver.js
+++ b/lib/receiver.js
@@ -21,8 +21,6 @@ const INFLATING = 5;
 
 /**
  * HyBi Receiver implementation.
- *
- * @extends stream.Writable
  */
 class Receiver extends Writable {
   /**
@@ -586,7 +584,7 @@ module.exports = Receiver;
 /**
  * Builds an error object.
  *
- * @param {(Error|RangeError)} ErrorCtor The error constructor
+ * @param {(typeof Error|typeof RangeError)} ErrorCtor The error constructor
  * @param {String} message The error message
  * @param {Boolean} prefix Specifies whether or not to add a default prefix to
  *     `message`
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -9,6 +9,9 @@ const { mask: applyMask, toBuffer } = require('./buffer-util');
 
 const mask = Buffer.alloc(4);
 
+
+/** @typedef {import('net').Socket} Socket */
+
 /**
  * HyBi Sender implementation.
  */
@@ -16,7 +19,7 @@ class Sender {
   /**
    * Creates a Sender instance.
    *
-   * @param {net.Socket} socket The connection socket
+   * @param {Socket} socket The connection socket
    * @param {Object} [extensions] An object containing the negotiated extensions
    */
   constructor(socket, extensions) {
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -9,13 +9,14 @@ const WebSocket = require('./websocket');
 const { format, parse } = require('./extension');
 const { GUID, kWebSocket } = require('./constants');
 
+
+/** @typedef {import('http').Server} Server */
+/** @typedef {import('http').IncomingMessage} IncomingMessage */
+/** @typedef {import('net').Socket} Socket */
+
 const keyRegex = /^[+/0-9A-Za-z]{22}==$/;
 
-/**
- * Class representing a WebSocket server.
- *
- * @extends EventEmitter
- */
+/** Class representing a WebSocket server. */
 class WebSocketServer extends EventEmitter {
   /**
    * Create a `WebSocketServer` instance.
@@ -34,7 +35,7 @@ class WebSocketServer extends EventEmitter {
    * @param {(Boolean|Object)} [options.perMessageDeflate=false] Enable/disable
    *     permessage-deflate
    * @param {Number} [options.port] The port where to bind the server
-   * @param {http.Server} [options.server] A pre-created HTTP/S server to use
+   * @param {Server} [options.server] A pre-created HTTP/S server to use
    * @param {Function} [options.verifyClient] A hook to reject connections
    * @param {Function} [callback] A listener for the `listening` event
    */
@@ -154,7 +155,7 @@ class WebSocketServer extends EventEmitter {
   /**
    * See if a given request should be handled by this server instance.
    *
-   * @param {http.IncomingMessage} req Request object to inspect
+   * @param {IncomingMessage} req Request object to inspect
    * @return {Boolean} `true` if the request is valid, else `false`
    * @public
    */
@@ -172,8 +173,8 @@ class WebSocketServer extends EventEmitter {
   /**
    * Handle a HTTP Upgrade request.
    *
-   * @param {http.IncomingMessage} req The request object
-   * @param {net.Socket} socket The network socket between the server and client
+   * @param {IncomingMessage} req The request object
+   * @param {Socket} socket The network socket between the server and client
    * @param {Buffer} head The first packet of the upgraded stream
    * @param {Function} cb Callback
    * @public
@@ -251,8 +252,8 @@ class WebSocketServer extends EventEmitter {
    *
    * @param {String} key The value of the `Sec-WebSocket-Key` header
    * @param {Object} extensions The accepted extensions
-   * @param {http.IncomingMessage} req The request object
-   * @param {net.Socket} socket The network socket between the server and client
+   * @param {IncomingMessage} req The request object
+   * @param {Socket} socket The network socket between the server and client
    * @param {Buffer} head The first packet of the upgraded stream
    * @param {Function} cb Callback
    * @throws {Error} If called more than once with the same socket
@@ -375,7 +376,7 @@ function socketOnError() {
 /**
  * Close the connection when preconditions are not fulfilled.
  *
- * @param {net.Socket} socket The socket of the upgrade request
+ * @param {Socket} socket The socket of the upgrade request
  * @param {Number} code The HTTP response status code
  * @param {String} [message] The HTTP response body
  * @param {Object} [headers] Additional HTTP response headers
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -36,7 +36,7 @@ class WebSocket extends EventEmitter {
   /**
    * Create a new `WebSocket`.
    *
-   * @param {(String|url.URL)} address The URL to which to connect
+   * @param {(String|URL)} address The URL to which to connect
    * @param {(String|String[])} [protocols] The subprotocols
    * @param {Object} [options] Connection options
    */
@@ -60,15 +60,14 @@ class WebSocket extends EventEmitter {
       this._bufferedAmount = 0;
       this._isServer = false;
       this._redirects = 0;
+      let proto = Array.isArray(protocols) ? protocols.join(', ') : protocols;
 
-      if (Array.isArray(protocols)) {
-        protocols = protocols.join(', ');
-      } else if (typeof protocols === 'object' && protocols !== null) {
+      if (typeof protocols === 'object' && protocols !== null) {
         options = protocols;
-        protocols = undefined;
+        proto = undefined;
       }
 
-      initAsClient(this, address, protocols, options);
+      initAsClient(this, address, proto, options);
     } else {
       this._isServer = true;
     }
@@ -392,6 +391,11 @@ class WebSocket extends EventEmitter {
   }
 }
 
+WebSocket.CONNECTING = 0
+WebSocket.OPEN = 1
+WebSocket.CLOSING = 2
+WebSocket.CLOSED = 3
+
 readyStates.forEach((readyState, i) => {
   const descriptor = { enumerable: true, value: i };
 
@@ -460,7 +464,7 @@ module.exports = WebSocket;
  * Initialize a WebSocket client.
  *
  * @param {WebSocket} websocket The client to initialize
- * @param {(String|url.URL)} address The URL to which to connect
+ * @param {(String|URL)} address The URL to which to connect
  * @param {String} [protocols] The subprotocols
  * @param {Object} [options] Connection options
  * @param {(Boolean|Object)} [options.perMessageDeflate=true] Enable/disable
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -60,14 +60,15 @@ class WebSocket extends EventEmitter {
       this._bufferedAmount = 0;
       this._isServer = false;
       this._redirects = 0;
-      let proto = Array.isArray(protocols) ? protocols.join(', ') : protocols;
 
-      if (typeof protocols === 'object' && protocols !== null) {
+      if (Array.isArray(protocols)) {
+        protocols = protocols.join(', ');
+      } else if (typeof protocols === 'object' && protocols !== null) {
         options = protocols;
-        proto = undefined;
+        protocols = undefined;
       }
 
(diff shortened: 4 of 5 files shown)


lib/receiver.js
```diff
@@ -586,7 +584,7 @@ module.exports = Receiver;
 /**
  * Builds an error object.
  *
- * @param {(Error|RangeError)} ErrorCtor The error constructor
+ * @param {(typeof Error|typeof RangeError)} ErrorCtor The error constructor
```
lpinca, line 587: Why `typeof`? It should be the `Error` or `RangeError` constructor.

lib/sender.js
```diff
@@ -9,14 +9,17 @@ const { mask: applyMask, toBuffer } = require('./buffer-util');
 
 const mask = Buffer.alloc(4);
 
+
+/** @typedef {import('net').Socket} Socket */
```
lpinca, line 13: This could also be a `tls.Socket`. Also, no TypeScript-specific syntax please.

lib/websocket-server.js
```diff
@@ -9,13 +9,14 @@ const WebSocket = require('./websocket');
 const { format, parse } = require('./extension');
 const { GUID, kWebSocket } = require('./constants');
 
+
+/** @typedef {import('http').Server} Server */
```
lpinca, line 13: Ditto.

lib/websocket.js
```diff
@@ -392,8 +392,13 @@ class WebSocket extends EventEmitter {
   }
 }
 
+WebSocket.CONNECTING = 0
+WebSocket.OPEN = 1
+WebSocket.CLOSING = 2
+WebSocket.CLOSED = 3
+
```
lpinca, line 399: Why this change? It is already done by ` Object.defineProperty(WebSocket, readyState, descriptor);` below.

lib/websocket.js
```diff
@@ -392,8 +392,13 @@ class WebSocket extends EventEmitter {
...
 readyStates.forEach((readyState, i) => {
-  const descriptor = { enumerable: true, value: i };
+  const descriptor = { enumerable: true, value: i, configurable: false, writable: false };
```
lpinca, line 401: Why this change? `false` is the default value for `configurable` and `writable`.

lib/receiver.js
jimmywarting: from VS code (TypeScript) perspective it looks at `@param {Error}` to be a instance of a Error class, when using typeof then it knows that it can be a Class

lib/sender.js
jimmywarting: i tried tls.Socket, didn't work so well... it could not figure out the correct NodeJS namespace
also typedef is a jsdoc thing

lib/websocket.js
jimmywarting: VS inteligens could not figure out what dynamically assigned properties meant

lib/websocket.js
jimmywarting: when i assigned the `WebSocket.CLOSING = 2` it probably changed the configurable to true, so i had to override it again - test was failing if i didn't do this

lpinca: What test? Behavior is tested here https://github.com/websockets/ws/blob/0ad1f9d6a48ed1b30bda09b958cb142c1e09cced/test/websocket.test.js#L102-L111.

lib/sender.js
jimmywarting: https://stackoverflow.com/questions/49836644/how-to-import-a-typedef-from-one-file-to-another-in-jsdoc-using-node-js

lib/websocket.js
lpinca: I prefer to not have changes only required for TypeScript / Visual Studio Code IntelliSense.

lib/websocket.js
jimmywarting: here is my test log if i removed conf + writable

```diff
  Constants
      `CONNECTING`
        1) is enumerable property of class
        ✓ is enumerable property of prototype
      `OPEN`
        2) is enumerable property of class
        ✓ is enumerable property of prototype
      `CLOSING`
        3) is enumerable property of class
        ✓ is enumerable property of prototype
      `CLOSED`
        4) is enumerable property of class
        ✓ is enumerable property of prototype
    Attributes
      `binaryType`
        ✓ is enumerable and configurable
        ✓ defaults to 'nodebuffer'
        ✓ can be changed to 'arraybuffer' or 'fragments'
      `bufferedAmount`
        ✓ is enumerable and configurable
        ✓ defaults to zero
        ✓ defaults to zero upon "open"
        ✓ takes into account the data in the sender queue
        ✓ takes into account the data in the socket queue
      `extensions`
        ✓ is enumerable and configurable
        ✓ exposes the negotiated extensions names (1/2)
        ✓ exposes the negotiated extensions names (2/2)
      `protocol`
        ✓ is enumerable and configurable
        ✓ exposes the subprotocol selected by the server
      `readyState`
        ✓ is enumerable and configurable
        ✓ defaults to `CONNECTING`
        ✓ is set to `OPEN` once connection is established
        ✓ is set to `CLOSED` once connection is closed
        ✓ is set to `CLOSED` once connection is terminated
      `url`
        ✓ is enumerable and configurable
        ✓ exposes the server url
    Events
      ✓ emits an 'error' event if an error occurs
      ✓ does not re-emit `net.Socket` errors
      ✓ emits an 'upgrade' event
      ✓ emits a 'ping' event
      ✓ emits a 'pong' event
    Connection establishing
      ✓ fails if the Sec-WebSocket-Accept header is invalid
      ✓ close event is raised when server closes connection
      ✓ error is emitted if server aborts connection
      ✓ unexpected response can be read when sent by server
      ✓ request can be aborted when unexpected response is sent by server
      ✓ fails if the opening handshake timeout expires (102ms)
      ✓ fails if an unexpected Sec-WebSocket-Extensions header is received
      ✓ fails if the Sec-WebSocket-Extensions header is invalid (1/2)
      ✓ fails if the Sec-WebSocket-Extensions header is invalid (2/2)
      ✓ fails if an unexpected extension is received (1/2)
      ✓ fails if an unexpected extension is received (2/2)
      ✓ fails if server sends a subprotocol when none was requested
      ✓ fails if server sends an invalid subprotocol
      ✓ fails if server sends no subprotocol
      ✓ does not follow redirects by default
      ✓ honors the `followRedirects` option
      ✓ honors the `maxRedirects` option
    Connection with query string
      ✓ connects when pathname is not null
      ✓ connects when pathname is null
    #ping
      ✓ throws an error if `readyState` is `CONNECTING`
      ✓ increases `bufferedAmount` if `readyState` is 2 or 3
      ✓ calls the callback w/ an error if `readyState` is 2 or 3
      ✓ can send a ping with no data
      ✓ can send a ping with data
      ✓ can send numbers as ping payload
      ✓ throws an error if the data size is greater than 125 bytes
    #pong
      ✓ throws an error if `readyState` is `CONNECTING`
      ✓ increases `bufferedAmount` if `readyState` is 2 or 3
      ✓ calls the callback w/ an error if `readyState` is 2 or 3
      ✓ can send a pong with no data
      ✓ can send a pong with data
      ✓ can send numbers as pong payload
      ✓ throws an error if the data size is greater than 125 bytes
    #send
      ✓ throws an error if `readyState` is `CONNECTING`
      ✓ increases `bufferedAmount` if `readyState` is 2 or 3
      ✓ calls the callback w/ an error if `readyState` is 2 or 3
      ✓ can send a big binary message (145ms)
      ✓ can send text data
      ✓ does not override the `fin` option
      ✓ sends numbers as strings
      ✓ can send binary data as an array
      ✓ can send binary data as a buffer
      ✓ can send an `ArrayBuffer`
      ✓ can send a `Buffer`
      ✓ calls the callback when data is written out
      ✓ works when the `data` argument is falsy
      ✓ honors the `mask` option
    #close
      ✓ closes the connection if called while connecting (1/3)
      ✓ closes the connection if called while connecting (2/3) (306ms)
      ✓ closes the connection if called while connecting (3/3)
      ✓ can be called from an error listener while connecting
      ✓ can be called from a listener of the 'upgrade' event
      ✓ throws an error if the first argument is invalid (1/2)
      ✓ throws an error if the first argument is invalid (2/2)
      ✓ throws an error if the message is greater than 123 bytes
      ✓ sends the close status code only when necessary
      ✓ works when close reason is not specified
      ✓ works when close reason is specified
      ✓ permits all buffered data to be delivered
      ✓ allows close code 1013
      ✓ allows close code 1014
      ✓ does nothing if `readyState` is `CLOSED`
      ✓ sets a timer for the closing handshake to complete
    #terminate
      ✓ closes the connection if called while connecting (1/2)
      ✓ closes the connection if called while connecting (2/2) (304ms)
      ✓ can be called from an error listener while connecting
      ✓ can be called from a listener of the 'upgrade' event
      ✓ does nothing if `readyState` is `CLOSED`
    WHATWG API emulation
      ✓ supports the `on{close,error,message,open}` attributes
      ✓ works like the `EventEmitter` interface
      ✓ doesn't return listeners added with `on`
      ✓ doesn't remove listeners added with `on`
      ✓ adds listeners for custom events with `addEventListener`
      ✓ allows to add one time listeners with `addEventListener`
      ✓ supports the `removeEventListener` method
      ✓ wraps text data in a `MessageEvent`
      ✓ receives a `CloseEvent` when server closes (1000)
      ✓ receives a `CloseEvent` when server closes (4000)
      ✓ sets `target` and `type` on events
      ✓ passes binary data as a Node.js `Buffer` by default
      ✓ ignores `binaryType` for text messages
      ✓ allows to update `binaryType` on the fly
    SSL
      ✓ connects to secure websocket server
      ✓ connects to secure websocket server with client side certificate
      ✓ cannot connect to secure websocket server via ws://
      ✓ can send and receive text data
      ✓ can send a big binary message (56ms)
      ✓ allows to disable sending the SNI extension
      ✓ works around a double 'error' event bug in Node.js
    Request headers
      ✓ adds the authorization header if the url has userinfo
      ✓ honors the `auth` option
      ✓ favors the url userinfo over the `auth` option
      ✓ adds custom headers
      ✓ excludes default ports from host header
      ✓ doesn't add the origin header by default
      ✓ honors the `origin` option (1/2)
      ✓ honors the `origin` option (2/2)
    permessage-deflate
      ✓ is enabled by default
      ✓ can be disabled
      ✓ can send extension parameters
      ✓ can send and receive text data
      ✓ can send and receive a `TypedArray`
      ✓ can send and receive an `ArrayBuffer`
      ✓ consumes all received data when connection is closed abnormally
      ✓ handles a close frame received while compressing data
      #close
        ✓ can be used while data is being decompressed
      #send
        ✓ ignores the `compress` option if the extension is disabled
        ✓ calls the callback if the socket is closed prematurely
      #terminate
        ✓ can be used while data is being compressed
        ✓ can be used while data is being decompressed
    Connection close edge cases
      ✓ closes cleanly after simultaneous errors (1/2)
      ✓ closes cleanly after simultaneous errors (2/2)


  325 passing (2s)
  4 failing

  1) WebSocket
       Constants
         `CONNECTING`
           is enumerable property of class:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  {
+   configurable: true,
-   configurable: false,
    enumerable: true,
    value: 0,
+   writable: true
-   writable: false
  }
      + expected - actual

       {
      -  "configurable": true
      +  "configurable": false
         "enumerable": true
         "value": 0
      -  "writable": true
      +  "writable": false
       }
      
      at Context.<anonymous> (test/websocket.test.js:105:18)
      at processImmediate (node:internal/timers:464:21)

  2) WebSocket
       Constants
         `OPEN`
           is enumerable property of class:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  {
+   configurable: true,
-   configurable: false,
    enumerable: true,
    value: 1,
+   writable: true
-   writable: false
  }
      + expected - actual

       {
      -  "configurable": true
      +  "configurable": false
         "enumerable": true
         "value": 1
      -  "writable": true
      +  "writable": false
       }
      
      at Context.<anonymous> (test/websocket.test.js:105:18)
      at processImmediate (node:internal/timers:464:21)

  3) WebSocket
       Constants
         `CLOSING`
           is enumerable property of class:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  {
+   configurable: true,
-   configurable: false,
    enumerable: true,
    value: 2,
+   writable: true
-   writable: false
  }
      + expected - actual

       {
      -  "configurable": true
      +  "configurable": false
         "enumerable": true
         "value": 2
      -  "writable": true
      +  "writable": false
       }
      
      at Context.<anonymous> (test/websocket.test.js:105:18)
      at processImmediate (node:internal/timers:464:21)

  4) WebSocket
       Constants
         `CLOSED`
           is enumerable property of class:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  {
+   configurable: true,
-   configurable: false,
    enumerable: true,
    value: 3,
+   writable: true
-   writable: false
  }
      + expected - actual

       {
      -  "configurable": true
      +  "configurable": false
         "enumerable": true
         "value": 3
      -  "writable": true
      +  "writable": false
       }
      
      at Context.<anonymous> (test/websocket.test.js:105:18)
      at processImmediate (node:internal/timers:464:21)

```

lpinca: Ah I understand now, this happens due to https://github.com/websockets/ws/pull/1910#discussion_r663678860. I guess that is another reason to avoid that change.

lib/sender.js
lpinca: From the comments:

> This is TypeScript-specific syntax. Vote for github.com/jsdoc/jsdoc/issues/1645.

jimmywarting: oh, look at that... 
anyway another way could be to import a unused variable

```js
var http = require('http')

* @param {http.Server}
function foo() {
}
```

lpinca: That's acceptable. However we might need to add an eslint comment for unused variables.

lib/websocket.js
jimmywarting: Do you have any other jsDoc solution up your sleeve that describes WebSocket to have this properties?

lpinca: It's weird but something like this should work:

```
/**
 * Class representing a WebSocket.
 *
 * @property {Number} CONNECTING  description
 */
```

jimmywarting: forgot about that thing

lib/receiver.js
lpinca, line 587:
```suggestion
 * @param {function(new:Error|RangeError)} ErrorCtor The error constructor
```

lpinca:
@jimmywarting are you ok with https://github.com/websockets/ws/pull/1912?

jimmywarting:
closing mine b/c it's superseeded and i'm overwhelmed with other work right now

lib/receiver.js
jimmywarting: 👍

PR #1912, opened 2021-07-06
Improve JSDoc-inferred types

lpinca:
Refs: https://github.com/websockets/ws/pull/1910

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -1,8 +1,13 @@
+/* eslint no-unused-vars: ["error", { "varsIgnorePattern": "^net|tls|https$" }] */
+
 'use strict';
 
 const EventEmitter = require('events');
+const http = require('http');
+const https = require('https');
+const net = require('net');
+const tls = require('tls');
 const { createHash } = require('crypto');
-const { createServer, STATUS_CODES } = require('http');
 
 const PerMessageDeflate = require('./permessage-deflate');
 const WebSocket = require('./websocket');
@@ -34,7 +39,8 @@ class WebSocketServer extends EventEmitter {
    * @param {(Boolean|Object)} [options.perMessageDeflate=false] Enable/disable
    *     permessage-deflate
    * @param {Number} [options.port] The port where to bind the server
-   * @param {http.Server} [options.server] A pre-created HTTP/S server to use
+   * @param {(http.Server|https.Server)} [options.server] A pre-created HTTP/S
+   *     server to use
    * @param {Function} [options.verifyClient] A hook to reject connections
    * @param {Function} [callback] A listener for the `listening` event
    */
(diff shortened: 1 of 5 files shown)


lib/websocket-server.js
```diff
@@ -1,14 +1,23 @@
...
-const { createServer, STATUS_CODES } = require('http');
 
 const PerMessageDeflate = require('./permessage-deflate');
 const WebSocket = require('./websocket');
 const { format, parse } = require('./extension');
 const { GUID, kWebSocket } = require('./constants');
 
+/** @typedef {import('http').Server} Server */
+/** @typedef {import('http').IncomingMessage} IncomingMessage */
+/** @typedef {import('net').Socket} Socket */
```
jimmywarting: if you require('http') then you don't need typedef?

lib/websocket.js
```diff
@@ -30,13 +30,17 @@ const closeTimeout = 30 * 1000;
 /**
  * Class representing a WebSocket.
  *
+ * @property {Number} CONNECTING The connection is not yet open
+ * @property {Number} OPEN The connection is open and ready to communicate
+ * @property {Number} CLOSING The connection is in the process of closing
+ * @property {Number} CLOSED The connection is closed
```
jimmywarting, line 36: tried this also, didn't work for me, maybe it's the right way to do it...

what did work was:
```js
class X {
  static CONNECTING = 1
  CONNECTING = 1
}
```
or like a mention earlier: `x.CONNECTING = 1`
but then it dose not have the right descriptor (unless you add those with defineProperties)

jimmywarting (approved):
LGTM

lib/websocket.js
lpinca: Does it work with

```js
Object.defineProperty(WebSocket.prototype, 'CONNECTING', { enumerable: true, value: 0 });
Object.defineProperty(WebSocket, 'CONNECTING', { enumerable: true, value: 0 });
```

(without the loop)?

lpinca: The same pattern is also used here https://github.com/websockets/ws/blob/7.5.2/lib/websocket.js#L402-L411. I'm not sure I want to change it. It is not a JSDoc issue.

jimmywarting: without the loop it did work fine doing just
```js
Object.defineProperty(WebSocket.prototype, 'CONNECTING', { enumerable: true, value: 0 });
```

tried it if it would work with a object using defineProperties


```js
Object.defineProperties(WebSocket, {
  CONNECTING: {
    enumerable: true,
    value: 0
  }
});
```
The result: `Property 'CONNECTING' does not exist on type 'typeof WebSocket'`

lpinca:
@jimmywarting PTAL.

jimmywarting:
LGTM

lpinca:
Thank you for the review.

Issue #1878, opened 2021-04-29
unnecessary copy with .toString() ? 


bitpup:
Not to sound dumb, but, coming from a Node perspective, isn't this causing an unnecessary copy to occur?

https://github.com/websockets/ws/blob/fc7e27d12ad0af90ce05302afc85c292024000b4/lib/receiver.js#L437

perhaps it's required for browser API?

lpinca:
We can't use the buffer directly because it's a text message so we need to convert it to a string. Is there another way?

bitpup:
it may well be a text payload, leave it's interpretation to the user?

for now i have a revised method with the buffer going straight to JSON.parse

lpinca:
Passing a buffer to `JSON.parse()` will convert it to a string implicitly.

> leave it's interpretation to the user?

We would then need to emit an additional argument in the `'message'` event to let the user know whether the message is a binary message or a text message. This is a breaking change. Currently a `Buffer` is used for binary messages and a `string` for text messages.

bitpup:
Agreed, I have a solution for now.

I'm not too sure if I like the arbitration of payload that WS employs, I'm thinking perhaps, in this respect it does too much for the user, e.g. it just a payload... I am probably wrong ;-)

Well done on your efforts so far

lpinca:
It is consistent with the browser interface where a `DOMString` or a `Blob`/`ArrayBuffer` is used depending on the type of the message.

There are already some deviations and we are not forced to follow it but I'm not very comfortable to always use a `Buffer`/`ArrayBuffer` because it's a big breaking change and there are a few edge cases like `WebSocket.createWebSocketStream()` that must be thought carefully.

I understand that is some cases, for example when passing data around, avoiding the string conversion might lead to better performances.

bitpup:
for completeness, here's my hack

I've also dropped using `isValidUTF8` as I expect/hope `JSON.parse` will implement a similar validity check

```javascript
const
GET_INFO = 0,
Receiver = require('ws/lib/receiver'),
{concat} = require('ws/lib/buffer-util');

function _dataMessage() {
	if (this._fin) {
		if (this._opcode === 2) {
			Receiver.prototype.dataMessage.call(this);

		} else {
			const messageLength = this._messageLength;
			const fragments = this._fragments;
	
			this._messageLength = 0;
			this._fragments = [];

			this.emit('message', concat(fragments, messageLength));
		}
	}

	this._state = GET_INFO;
}
.
.
.
[...]

PR #1928, opened 2021-08-11
Add ability to skip UTF-8 validation

### LINKED ISSUE #1924: Option to skip utf-8 check
      issue.

#### Description

I was wondering if it would be interesting to have an option (disabled by default) to skip the UTF-8 check. This would give a slight performance gain in my opinion. (Only for the client)

lpinca:
Add the `skipUTF8Validation` option to skip UTF-8 validation for text
and close messages.

Refs: https://github.com/websockets/ws/issues/1878
Closes #1924


Mesteery (approved):
Thank you!

Issue #1950, opened 2021-09-23
Why is the default for `perMessageDeflate.threshold` set to 1024?


alexdima:
### 

- [X] I've searched for any related issues and avoided creating a duplicate issue.

### Description

I've found [the original PR](https://github.com/websockets/ws/pull/872) that introduces this functionality, but I don't understand what is the reasoning behind this functionality, nor why the default was chosen to be 1024. I have also tried searching `socket.io` and their repositories, but I could not find the explanation.

From my understanding of how permessage-deflate works, it uses zlib with a window that is stateful across frames (typically 2^15 previous bytes). So each frame is compressed by possibly reusing bytes sent with previous frames. To clarify: this is not like having 1000 small files and creating 1000 independent zip files, this is more like having 1000 small files and creating a single zip file. So it should pay off to compress small files, as it should pay off to compress small frames.

From my observations of all major browsers (Chrome, Safari, Firefox, Edge), all of them, once they negotiate `permessage-deflate`, compress each and every frame they send (by setting the `RSV1` bit). This aligns with my understanding of how zlib works and how it can quickly pay off even for very small frames.

lpinca:
> it uses zlib with a window that is stateful across frames

It depends. Context takeover can be disabled. When not used, a new zlib stream is used for every frame. Compressing small payloads is counterproductive (the compressed payload might be bigger than the original). 1024 was chosen as default because it is a good tradeoff.

alexdima:
@lpinca thank you for your prompt reply and apologies for taking such a long time to follow up, I have been busy with a deadline. I think you are correct that context takeover can be disabled, but I believe that only in the case when context takeover is disabled should the `threshold` setting be configured as `1024` or even have any effect. IMHO, in all other cases, **it is always productive to compress any message, regardless of its size**. 

`ws` is a super popular library, so apologies if I am a bit insisting too much on this topic. If I understand things correctly, I believe the same logic (`1024` threshold) is used when a server writes back to its clients, so changing `ws` to have good defaults here might make a big impact in terms of bandwidth used by all your users.

I have put together the following example to demonstrate my observations: https://github.com/alexdima/ws-compression-repro (I apologize for the quality of the code)

To execute and compare how `ws` is doing relative to the browser WebSocket you can do the following:
1. execute `node server.js`
2. visit in your browser `http://127.0.0.1:8080/` and reload a few times
3. the browser sends each time `200` messages totaling `146865` chars
[...]

PR #1954, opened 2021-09-28
Ignore the `threshold` option if context takeover is enabled

lpinca:
When context takeover is enabled, compress messages even if their size
is below the value of `threshold` option.

Refs: https://github.com/websockets/ws/issues/1950


lpinca:
cc: @alexdima

PR #1956, opened 2021-09-30
Add `WebSocket#pause()` and `WebSocket#resume()`

lpinca:
Add ability to pause and resume a websocket.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -31,15 +31,18 @@
   - [websocket.bufferedAmount](#websocketbufferedamount)
   - [websocket.close([code[, reason]])](#websocketclosecode-reason)
   - [websocket.extensions](#websocketextensions)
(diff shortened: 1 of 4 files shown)


lpinca:
I'm a bit torn about this. I don't like to add more non "standard" methods and properties but they might be useful if someone does not want the additional overhead introduced by the duplex stream returned by `createWebSocketStream()`.

Also, there is an edge case where the wrapper duplex stream might now work as expected if used to echo back messages (`duplex.pipe(duplex)`), due to how data is buffered. If the stream is paused and the remote peer sends a close frame, all buffered data will not be sent when the stream is resumed because the websocket ready state is no longer `OPEN`.

What do you think?

Flecktarnhasi (approved):

Issue #337, opened 2014-07-26
Constantly crashing when sending. Error not opened


thailemon:
I'm suspecting it's because client is trying to send to a dead socket.
Any other way to handle this?

My scenario is mobile App -> browser. App gets disconnect event too late, so it keeps sending packets while the browser is already disconnected.

Here's my test code:

```
var Client = [];
function findClient(url) {
return (Client.indexOf(url));
}


wss.on('connection', function(ws) {
    Client.push(ws);
    console.log('Connected: %s', ws.upgradeReq.url);
    console.log('Clients: %s', Client.length);

ws.on('message', function(message, flags) {

    for( var i = 0; i < Client.length; i++) {
         Client[i].send(message, function(error) {
         if(error != null) {
         console.log('error: %s', error);
         ws.close();
         ws._socket.destroy();
         Client.splice(findClient(ws.upgradeReq.url));
}
});

ws.on('close', function(user) {
    console.log('Disconnected: %s', ws.upgradeReq.url);
    ws.close();
    ws._socket.destroy();
    Client.splice(findClient(ws.upgradeReq.url));
    console.log('Clients: %s', Client.length);



  });
       ws.on('error', function(error) {
        if(error != null) {
         console.log('error: %s', error);
         ws.close();
         ws._socket.destroy();
         Client.splice(findClient(ws.upgradeReq.url));
    }
  });

    ws.on('close', function(user) {
    console.log('Disconnected: %s', ws.upgradeReq.url);
    ws.close();
    ws._socket.destroy();
    Client.splice(findClient(ws.upgradeReq.url));
    console.log('Clients: %s', Client.length);

  });
```

Any help appreciated :)

galambalazs:
I've seen this problem come up for about 99% of the people starting to use this library. 

Interesting to note that the WHATWG spec only talks about throwing an error when the socket is in CONNECTING state (otherwise the send simply has no effect). 

https://html.spec.whatwg.org/multipage/comms.html#dom-websocket-send

> The send(data) method transmits data using the connection. If the readyState attribute is CONNECTING, it must throw an InvalidStateError exception. Otherwise, the user agent must run the appropriate set of steps from the following list: [...]

zag2art:
Imo, throwing an error is a good for browser but it's absolutely bad practice for server.

galambalazs:
My opinion is that send is inherently asynchronous and it should be left for the callback to handle errors (if needed).

piecioshka:
+1

[...]

PR #1966, opened 2021-11-01
Enable on headers handler to reject connection modifying headers param

sdrsdr:
A common problem is authenticating and rejecting websocket clients.  The `headers` handler of the server is in unique position to easily reject the client after expecting all the available information from the request. This minimal patch detects the attempt to close the connection and does so in graceful manner.  See the added test for easy client rejection

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -389,6 +389,13 @@ class WebSocketServer extends EventEmitter {
     socket.write(headers.concat('\r\n').join('\r\n'));
     socket.removeListener('error', socketOnError);
 
+    if (headers.length==0 || !headers[0].startsWith('HTTP/1.1 101 ')) { 
+      //An error status and headers were just send. Terminate the connection
+      socket.end();
+      //as the socket is not writable the code will not be used
+      return abortHandshake(socket, 500); 
+    }
+
     ws.setSocket(socket, head, {
       maxPayload: this.options.maxPayload,
       skipUTF8Validation: this.options.skipUTF8Validation
--- a/test/onheaeder-httperr.test.js
+++ b/test/onheaeder-httperr.test.js
@@ -1,0 +1,53 @@
+/* eslint no-unused-vars: ["error", { "varsIgnorePattern": "^ws$" }] */
+
+'use strict';
+
+const WebSocket = require('..');
+
+
+describe('WebSocketServer', () => { 
+  it('successfully rejects a connection via error heders in on headers handler', (done) => {
+  const wss = new WebSocket.Server(
+    {
+      port: 0
+    },
+    () => {
+      const ws = new WebSocket(`ws://localhost:${wss.address().port}`);
+      wss.on('connection', (ws) => {
+        done(new Error("WSS: on connection?! We should not be here!"));
+        ws.close();
+        wss.close();
+      });
+      wss.on('headers', (headers,req)=>{
+        headers.length=0;
+        headers.push('HTTP/1.1 418 I\'m a teapot');
+        headers.push('Connection: close');
+      });
+
+      ws.on('open', () => {
+       done(new Error("WS: on connection?! We should not be here!"));
+       ws.close();
+       wss.close();
+      });
+      ws.on('error',(err)=>{
+        if (err instanceof Error) {
+          if (err.message.endsWith(' 418')) {
+            //end with no error
+            done();
+            ws.close();
+            wss.close();
+          } else {
+            done(new Error("WS: on err but err is not as expected!"));
+            ws.close();
+            wss.close();
+          }
+        } else {
+          done(new Error("WS: on err but err is not instance of Error!"));
+          ws.close();
+          wss.close();
+        }
+      });
+    }
+  );
+
+})});


lpinca:
The [`'headers'`](https://github.com/websockets/ws/blob/master/doc/ws.md#event-headers) event was not designed to do this. See https://github.com/websockets/ws#client-authentication on how to properly authenticate and reject connections.

sdrsdr:
What I don't like in the recommended  way is that I as a user should be familiar enough with the library so I know weather to call `wss.handleUpgrade` and what the exact callback it should be. Especially  `wss.emit('connection', ws, request, client);` should NOT be in the scope of the user's usage of the library. In my eyes this is HIGHLY unportable and hakish way to use the library. What I currently do in my code is "abuse" the `headers` event and just `req.destroy()` when I have to, but this does not allow me to specify an error code for the HTTP protocol and recently this become important to me, hence the pull request.

lpinca:
> What I don't like in the recommended  way is that I as a user should be familiar enough with the library so he/she knows weather to call `wss.handleUpgrade` and what the exact callback it should be. Especially  `wss.emit('connection', ws, request, client);` should NOT be in the scope of the user's usage of the library. In my eyes this is HIGHLY unportable and hakish way to use the library.

This is [documented](https://github.com/websockets/ws/blob/8.2.3/doc/ws.md#serverhandleupgraderequest-socket-head-callback) and it is the recommended way to reject connections as it gives the user total freedom. Calling `wss.emit('connection', ws, request, client);` is optional, it is only sugar, you could stop at the `server.handleUpgrade()` callback.

Also as written above, the [`verifyClient`](https://github.com/websockets/ws/blob/8.2.3/doc/ws.md#new-websocketserveroptions-callback) hook can also be used to do this if you don't want to use `server.handleUpgrade()`.

Sorry but I'm not ok with this patch. Abusing the `'headers'` event is not the proper way to reject connections.

sdrsdr:
If I'm understanding the code correctly, I can't omit the callback of `server.handleUpgrade()` I MUST call it exactly as specified in the "documentation"  or I'll break the usual flow of the events: `server.on('connection')` handlers will not be called at best, runtime errors will be thrown most probably. 

  I'm not huge fan of my patch also but it looks way better than the proposed way to handle client filtration. As  `verifyClient` is strongly discouraged we're only left with code snippet from a documentation that looks so tightly coupled with inner working of the library that I will expect every other `npm upgrade`  to break it silently.

   Probably you should remove the stigma from `verifyClient` or device a clean method to handle rejection of connections. 

  Most simple way, probably, would be to provide a default callback for  `server.handleUpgrade()`  so the documentation can hide the inner workings of the library and and the uses can be sure that all the workflow will unfold properly no-matter the version of the library used. If you don't like the approach  with optional parameters  you can introduce a `server.handleUpgradePostAuth()`  with the same arguments as `server.handleUpgrade()` sans the `cb` param that internally calls  `server.handleUpgrade()`  with proper callback

lpinca:
You can't omit the callback but you don't need `server.on('connection')`. When `server.handleUpgrade()` invokes the callback, you are done. Your `WebSocket` instance is there. Everything else is sugar.

```js
import WebSocket from 'ws';

const wss = new WebSocketServer({ noServer: true });

  // This function is not defined on purpose. Implement it with your own logic.
  authenticate(request, (err, client) => {
    if (err || !client) {
      socket.write('HTTP/1.1 401 Unauthorized\r\n\r\n');
      socket.destroy();
      return;
    }

    wss.handleUpgrade(request, socket, head, function done(ws) {
      // You are done. `ws` is your `WebSocket` instance, you can do
      // whatever you want with it.
    });
  });
});

server.listen(8080);
```

> Probably you should remove the stigma from `verifyClient` or device a clean method to handle rejection of connections. 

The `verifyClient` hook will not be removed anytime soon.

sdrsdr:
Documenting a way to break the documented workflow of the library is nonsense. Imagine I've build a websocket server according to the documentation using `on('connection')`  handler everything is fine then I decide to add client connection filtering following the same documentation and endup breaking my perfectly working code.
  
I'm pretty sure everybody will prefer to maintain only code related to his/her project and emitting the  `connection` event is strictly library's business and nobody else. 

A robust library should strive to keep a stable interface keeping the user code agnostic to library's inner workings. This will allow confident version upgrades for the users and freedom to expand the library with less worries that the user code could be broken from some needed change. 

`server.handleUpgrade()` in current form is unusable as public API IMHO. If  `verifyClient` is not discouraged please update the documentation.  See Issue #337 ;)

The most important question here: If I create a PR introducing server.handleUpgradePostAuth() that will allow proper, according to me, way to handle client connection filtration, are you willing to consider accepting it?

lpinca:
> I'm pretty sure everybody will prefer to maintain only code related to his/her project and emitting the  `connection` event is strictly library's business and nobody else.

When using `server.handleUpgrade()` directly there is no `'connection'` event. It's up to the user to emit it.

> `server.handleUpgrade()` in current form is unusable as public API IMHO.

I disagree. It is used by a lot projects.

> The most important question here: If I create a PR introducing server.handleUpgradePostAuth() that will allow proper, according to me, way to handle client connection filtration, are you willing to consider accepting it?

No, the current API is ok in my opinion and is used by thousands of projects. We have received no complaints that I remember of on the suggested way of rejecting connections. Perhaps this is the first? If you don't like the API that's ok but I also don't like abusing the `'headers'` event.

Thank you anyway.

sdrsdr:
Take a look here again https://github.com/websockets/ws/issues/377 and evaluate how user frendly and easy to use is your library. Allowing small, backward compatible  and consistent changes that will benefit your users will only help your project.

At least allowing passing no callback to `server.handleUpgrade()` and doing the proper thing that will emit `connection` event is something you should strongly consider.

lpinca:
It seems that overall users are happy with the [suggested solution](https://github.com/websockets/ws/issues/377#issuecomment-462152231). Some advantages over the `verifyClient` hook are:

1. Total control on the socket. Users can write whatever they want on it. 
1. Ability to add custom arguments to the `'connection'` event (if wanted) without hacks, which is the reason why #377 was opened in the first place.

> At least allowing passing no callback to `server.handleUpgrade()` and doing the proper thing that will emit `connection` event is something you should strongly consider.

As written in https://github.com/websockets/ws/issues/377#issuecomment-46215223,1 `server.handleUpgrade()` takes a callback because the `verifyClient` hook can be asynchronous.

Again, you might not like the API and that's perfectly fine. We can't make everyone happy. Thank you for sharing your opinion.

sdrsdr:
I Don't see how default callback will break anybody's code but mandating copy-paste a snippet from documentation to maintain default behaviour is over the red line for me

lpinca:
There is no default behavior, just two different modes. If the users use `server.handleUpgrade()`, then they should rely on its callback. If users do not use `server.handleUpgrade()`, then they should rely on the `'connection'` event. The fact that `server.handleUpgrade()` is also used internally is an implementation detail.

sdrsdr:
I personally find this   https://en.wikipedia.org/wiki/Principle_of_least_astonishment very important. This leads to easy code refactoring and ease of use. If there is a event called when a connection is established it should be called when connection is established no matter how the connection is established. And this should happen no matter what my code does as long as the library does not crash. But this is just my expectations and all my code tries to be as predictable as possible but this is not my library and you're free to code it as you like

PR #1967, opened 2021-11-02
Allow cb to be optional for WebSocketServer.handleUpgrade

sdrsdr:
this allows WebSocketServer.handleUpgrade to be called without a callback, in which case the default behaviour will be triggered  and the `connection` event will be fired

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -121,6 +121,7 @@ class WebSocketServer extends EventEmitter {
 
     this.options = options;
     this._state = RUNNING;
+    //if (options.noServer && typeof(callback)=='function') setImmediate(callback);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lpinca:
See https://github.com/websockets/ws/pull/1968 for a possibly cleaner implementation.

lib/websocket-server.js
```diff
@@ -121,6 +121,7 @@ class WebSocketServer extends EventEmitter {
 
     this.options = options;
     this._state = RUNNING;
+    //if (options.noServer && typeof(callback)=='function') setImmediate(callback);
```
sdrsdr, line 124: What about this fix? You see when `noServer` option is set the callback never get called. This will lead to confusion when somebody is porting old code with no client filtering to the recommended way for filtering with "stand alone" http/https server and  noServer websocket server.

lpinca: That is expected. Calling the callback in `noServer` mode would create scenarios where the callback is called and the external server is not listening or already listening. The user should rely on the `'listening'` event of the external server instead.

sdrsdr: Ok, it's fare enough, although the introduced new variable to hold the reference to the new server can not be already present in the code under refactoring, but granted some use-before-set scenario is possible, but this is also possible when moving the old code outside of the now-not-woking callback. To be on the safe side throw an error when callback and `noServer` options are specified together? This will make the need to migrate the code from the callback VERY visible

lpinca: That is an option but it is not strictly needed. Anyone using the callback will notice if it is no longer called, plus it is a breaking change.

PR #1968, opened 2021-11-02
Make `WebSocketServer#handleUpgrade()` callback optional

lpinca:
If the upgrade is successful and the `callback` argument is not
provided, then emit the `'connection'` event.

Refs: https://github.com/websockets/ws/pull/1966

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -219,7 +219,7 @@ tracking is disabled. In this case the `'close'` event is emitted in the next
 tick. The optional callback is called when the `'close'` event occurs and
 receives an `Error` if the server is already closed.
 
-### server.handleUpgrade(request, socket, head, callback)
+### server.handleUpgrade(request, socket, head[, callback])
 
 - `request` {http.IncomingMessage} The client HTTP GET request.
 - `socket` {net.Socket} The network socket between the server and client.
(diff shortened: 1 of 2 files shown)


lpinca:
I don't like this because the method behavior depends on the `callback` argument. Instead it should always or never emit the `'connection'` event, regardless of whether the `callback` argument is given or not.

sdrsdr:
Works for me! 😃

sdrsdr:
https://github.com/websockets/ws/blob/5b87c5191ddbdae974f553fa841270b7919ae10c/test/handleupgrade-default-cb.test.js this is a test case I've forgotten in my firts commit in my PR. You can use it to test your implementation also

lpinca:
> https://github.com/websockets/ws/blob/5b87c5191ddbdae974f553fa841270b7919ae10c/test/handleupgrade-default-cb.test.js this is a test case I've forgotten in my firts commit in my PR. You can use it to test your implementation also 

It's already covered by this change https://github.com/websockets/ws/blob/bc73223c3e0ec7ac503ef59ccb4d44479f5b4b8a/lib/websocket-server.js#L108. A lot of existing test would fail if the `'connection'` event is not emitted.

lpinca:
I will keep this open for a while to collect feedback. If no one will comment I will close it.

sdrsdr:
If you publish new package version I can try update the TypeScript definitions at DefinitelyTyped

sdrsdr:
What is the timeframe for this PR?

lpinca:
See https://github.com/websockets/ws/pull/1968#issuecomment-957777758.

KVROHITH098 (changes requested):
I approve But I want this to be validated.

lpinca:
I'm closing this as per https://github.com/websockets/ws/pull/1968#issuecomment-957777758.

sdrsdr:
Closed but not merged? I'm not sure what's going on here!

lpinca:
Yes, aside from yours, there seems to be no interest in getting this merged.

PR #1972, opened 2021-11-13
docs: simple example that work locally out of the box

yveslange:
Was trying to quickly set up a websocket using `ws` and to test it using the browser. 

With the new policies of Chrome, it is not possible to send messages to the localhost if not specified as a parameter `host` of the websocket server when not using WSS.

When testing a library, we want to quickly see the result and if it's working. We don't want to set a TLS certificat and it should directly work locally since it should not be tested on a remote server (that should have opened port).

I am suggesting to clear the README.md so that the example will work and display console message out of the box by copying the first example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,26 +100,18 @@ See [the docs][ws-server-options] for more options.
 import WebSocket, { WebSocketServer } from 'ws';
 
 const wss = new WebSocketServer({
+  host: "127.0.0.1",
   port: 8080,
-  perMessageDeflate: {
-    zlibDeflateOptions: {
-      // See zlib defaults.
-      chunkSize: 1024,
-      memLevel: 7,
-      level: 3
-    },
-    zlibInflateOptions: {
-      chunkSize: 10 * 1024
-    },
-    // Other options settable:
-    clientNoContextTakeover: true, // Defaults to negotiated value.
-    serverNoContextTakeover: true, // Defaults to negotiated value.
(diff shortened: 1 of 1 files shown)


lpinca:
> With the new policies of Chrome, it is not possible to send messages to the localhost if not specified as a parameter `host` of the websocket server when not using WSS.

It does not make much sense. The `host` option specifies the address where to bind the server. If omitted it will accept connections on all interfaces. See https://nodejs.org/api/net.html#serverlistenoptions-callback.

I'm running the latest Chrome and I can't reproduce the issue:

```js
import http from 'http';

const data = `<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <script>
      (function () {
        var ws = new WebSocket('ws://localhost:8080');

        ws.onopen = function () {
          console.log('open');
        };

        ws.onmessage = function (evt) {
          console.log(evt.data);
        }
      })();
    </script>
  </body>
</html>`;

server.on('request', (req, res) => {
  res.setHeader('Content-Type', 'text/html');
  res.end(data);
});

const wss = new WebSocketServer({ server });

wss.on('connection', function (ws) {
  ws.send('hello');
});

  console.log('server listening on port 8080');
});
```

Also, this PR changes the "WebSocket compression" example and removes all compression options from it. The compression options are now in the "Simple server" example. Sorry, but I'm not ok with this.

lpinca:
I'm closing this. Discussion can continue if needed.

PR #1973, opened 2021-11-16
extending ws to support 48bit values

prettydiff:
The ws library is using 32 bit values on the read and write for payload length.  This is because web sockets supports 63 bit payload lengths if the payload flag is 127 but JavaScript's max number is 53 bits - 1.

Node.js supports a 48 bit read and write operation.  The maximum payload length supported by 32 bits is 2.1gb.  The maximum payload length supported by 48 bits is 281tb.

Is there test automation or CI I can run this against?

### DIFF
--- a/lib/receiver.js
+++ b/lib/receiver.js
@@ -356,7 +356,7 @@ class Receiver extends Writable {
       );
     }
 
-    this._payloadLength = num * Math.pow(2, 32) + buf.readUInt32BE(4);
+    this._payloadLength = buf.readUIntBE(2, 6);
     return this.haveLength();
   }
 
--- a/lib/receiver.js
+++ b/lib/receiver.js
@@ -356,7 +356,7 @@ class Receiver extends Writable {
       );
     }
 
-    this._payloadLength = buf.readUIntBE(2, 6);
+    this._payloadLength = num * Math.pow(2, 32) + buf.readUInt32BE(4);
     return this.haveLength();
   }
 
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -75,8 +75,7 @@ class Sender {
     if (payloadLength === 126) {
       target.writeUInt16BE(data.length, 2);
     } else if (payloadLength === 127) {
-      target.writeUInt32BE(0, 2);
-      target.writeUInt32BE(data.length, 6);
+      target.writeUIntBE(data.length, 4, 6);
     }
 
     if (!options.mask) return [target, data];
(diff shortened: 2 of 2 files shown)


lib/receiver.js
```diff
@@ -356,7 +356,7 @@ class Receiver extends Writable {
       );
     }
 
-    this._payloadLength = num * Math.pow(2, 32) + buf.readUInt32BE(4);
+    this._payloadLength = buf.readUIntBE(2, 6);
```
lpinca, line 359: This actually reduces accuracy from 53 bits to 48 bits.

lib/sender.js
```diff
@@ -75,8 +75,7 @@ class Sender {
     if (payloadLength === 126) {
       target.writeUInt16BE(data.length, 2);
     } else if (payloadLength === 127) {
-      target.writeUInt32BE(0, 2);
-      target.writeUInt32BE(data.length, 6);
+      target.writeUIntBE(data.length, 4, 6);
```
lpinca, line 78: This looks ok but we need to keep `target.writeUInt32BE(0, 2);` or use `target.writeUInt16BE(0, 2);` to ensure that the first two bytes of the 8 bytes payload length are zeroed. Note the use of `Buffer.allocUnsafe()` above that returns uninitialized memory.

That said, it is extremely unlikely that anyone sends more that 4 GiB of data as a single message. No issues have been raised, that I remember of, in this regard since I am a maintainer of this library.

lib/receiver.js
prettydiff: So, the prior existing math is correct in that it accounts for 64 bits of addressable space after erroring if the payload size is greater than the (2**53) - 1.  Maybe for a future pull request I could submit something like `this._payloadLength = [num, buf.readUInt32BE(4)];` similar to the orientation Node uses for the now deprecated process.hrtime.  That would also be cleaner than accounting for the 53bit limitation, but its a larger refactor than the subject of this pull request.

I have found in my own web socket implementation the payload length of the receiving payload is more important than I originally thought because sometimes include trailing noise on the fit packet that needs to be trimmed.

lib/sender.js
prettydiff: I have not used the `allocUnsafe` method before.  When you create a buffer using `Buffer.alloc(10)` all indexes are populated with 0 values unless overwritten.

lpinca (approved):

lpinca:
Thank you.

lpinca:
FWIW the largest allowed size for a `Buffer` is [2^32](https://nodejs.org/docs/latest-v16.x/api/buffer.html#bufferconstantsmax_length), so right now, even with this change, it is not possible to write frames with more than 2^32 bytes of data.

PR #1975, opened 2021-11-19
[bug] initAsClient throws an uncaughtException

KevinYeramian:
`initAsClient` is used in the constructor and in the listener 'response' of the request [here](https://github.com/websockets/ws/blob/master/lib/websocket.js#L751) during redirection. `initAsClient` can throw an error in the listener which lead to `uncaughtException`.  The only solution I found to catch the error is to add a listener to `process` on the event `uncaughtException`.

You can also reproduce the bug with this code:
```

import WebSocket from 'ws'

async function main () {
    const url = 'ws://data.com'
    const wsOptions = {
        followRedirects: true,
    }
        
    const ws = new WebSocket(url, wsOptions)
}

main()
```

The problem is a redirection from `ws` to `http` for example.  

Should `initAsClient` emit an error instead of throw an error? 

System:
 - OS: macOS 11.6
 - CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
 - Memory: 477.56 MB / 16.00 GB
 -  Shell: 5.8 - /bin/zsh

ws version: `8.2.3` 
node version: `v12.22.4`

PS: I was not able to create an issue (400 on the API).

### DIFF
--- a/test/websocket.test.js
+++ b/test/websocket.test.js
@@ -1028,6 +1028,35 @@ describe('WebSocket', () => {
         ws.on('close', () => done());
       });
     });
+
+    it('emits an error on incorrect protocol redirection', (done) => {
+      const wss = new WebSocket.Server({ noServer: true, path: '/foo' });
+      const port = server.address().port;
+
+      server.once('upgrade', (req, socket) => {
+        socket.end(`HTTP/1.1 302 Found\r\nLocation: https://localhost:${port}/foo\r\n\r\n`);
+        server.once('upgrade', (req, socket, head) => {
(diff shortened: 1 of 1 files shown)


lpinca:
I'm not sure if this should be fixed. We are following the [HTML specification](https://html.spec.whatwg.org/#the-websocket-interface) and a thrown error is expected in this case. The same happens if the redirect URL contains a fragment identifier.

We could emit the error instead of throwing it if we are in a redirect but I'm not very happy with it because this changes the behavior depending on whether the connection is the initial one or not.

PR #1980, opened 2021-11-21
Do not throw if the redirect URL is invalid

lpinca:
If the redirect URL is invalid, then emit the error instead of throwing
it, otherwise there is no way to handle it.

Closes #1975


lpinca:
cc: @KevinYeramian

PR #1986, opened 2021-12-17
perf: use a random pool

ronag:
In production `randomFillSync` is taking 8-9% of our cpu time. This tries to reduce the overhead by pooling random data.


lib/random-util.js
```diff
@@ -0,0 +1,36 @@
...
+    randomPool = buf;
+    randomPoolIdx = 0;
+  }
+}
+
+let randomPending = true;
+crypto.randomBytes(RANDOM_POOL_SIZE, onRandomBytes);
+
+function randomFillSync (buffer, offset, size) {
+  if (RANDOM_POOL_SIZE - randomPoolIdx < size) {
```
sdrsdr, line 21: What happens if randomFillSync is constantly called with size > RANDOM_POOL_SIZE we end up always calling crypto.randomFillSync that we like to avoid
 A better strategy IMHO will be to use the random pool as a ring buffer and try to refresh it with real randomness but if there is a need (we wrapped around the ring buffer) we do a quick XOR over the whole buffer with some math.rand  based number and fulfil the request and then try to refresh the random poll

lib/sender.js
```diff
@@ -4,7 +4,7 @@
 
 const net = require('net');
 const tls = require('tls');
-const { randomFillSync } = require('crypto');
+const { randomFillSync } = require('./random-util');
```
sdrsdr, line 7: I'm not a huge fan of this kind of api replace. I prefer when reading the code to be noted that this is altered version of the api. I know this will require more changes but states the intent more clearly.

sdrsdr (commented):
I like the idea in general, but I'm not a maintainer of the project. I hope your PR get merged :)

lib/random-util.js
ronag: I update PR. It's always called with size === 4

lib/sender.js
ronag: randomFillMask

lib/sender.js
```diff
@@ -81,7 +81,7 @@ class Sender {
 
     if (!options.mask) return [target, data];
 
-    randomFillSync(mask, 0, 4);
```
sdrsdr, line 84: Ok this explains a lot so no need to overdo! Forget about my previous ideas with  ring buffers :)

lib/random-util.js
```diff
@@ -0,0 +1,38 @@
...
+    randomPoolIdx = 0;
+  }
+}
+
+let randomPending = true;
+crypto.randomBytes(RANDOM_POOL_SIZE, onRandomBytes);
+
+function randomFillMask (buffer) {
+  if (RANDOM_POOL_SIZE - randomPoolIdx < 4) {
+    return crypto.randomFillSync(buffer, offset, size);
```
sdrsdr, line 22: It's still best to try to refresh the whole poll at this point also

sdrsdr (commented):
Much simple problem than I thought initially! I really do hope this gets merged!

lib/random-util.js
sdrsdr: BTW once this if triggers it will trigger forever. Granted it won't trigger in most of the cases .. you have plenty of room but yet it's dangerous as it is

ronag: Why do you thing it will trigger forever? When we are here then we are already refreshing the pool. So this will only trigger until the randomBytes callback has been invoked.

sdrsdr: Uh yes .. you're right this is just the  specifics of the side effects of this particular case. Just put a comment here to help the next one reading the code. simply returning like this without any state change is huge red light in most heads. BTW randomPending might  go as well?

lib/random-util.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const RANDOM_POOL_SIZE = 8192;
+const RANDOM_POOL_REFRESH = 1024;
+
+let randomPool = Buffer.alloc(RANDOM_POOL_SIZE);
```
sdrsdr, line 6: We don't need to .alloc this - as we use the buffer from the callback. Just set randomPoolIdx to RANDOM_POOL_SIZE to prevent "use-before-set"

lib/sender.js
```diff
@@ -4,15 +4,31 @@
...
 const { isValidStatusCode } = require('./validation');
 const { mask: applyMask, toBuffer } = require('./buffer-util');
+const { randomBytes } = require('crypto');
 
 const mask = Buffer.alloc(4);
 
+const RANDOM_POOL_SIZE = 8192;
+const RANDOM_POOL_REFRESH = 1024;
+
+let randomPool = Buffer.alloc(RANDOM_POOL_SIZE);
```
sdrsdr, line 19: Maybe Buffer.alloc(0) here to allow the check on line 100 to trigger and not waste some heap?

sdrsdr:
Fingers crossed this gets trough :)

lib/sender.js
```diff
@@ -81,13 +97,24 @@ class Sender {
 
     if (!options.mask) return [target, data];
 
-    randomFillSync(mask, 0, 4);
+    if (randomPool.length - randomPoolIdx < 4) {
+      randomPool = randomBytes(32);
```
lpinca, line 101: I'm not convinced this is much better than the status quo. It would only take 8 `WebSocket` each sending a single message to empty the pool and there is a new `Buffer` allocation every time.

lpinca:
I think that if there are a lot of sockets all of them sending data at the same time synchronously, then this does not change much. I would like to see a "before" and "after" comparison. Do you have a flamegraph?

lpinca:
Also from my experience generating random data is a nothing compared to the actual masking of the data in terms of CPU time.

sdrsdr:
It will be best to skip masking for TLS websockets..

lpinca:
> It will be best to skip masking for TLS websockets

No sure if it is spec compliant. What about proxies?

ronag:
> Also from my experience generating random data is a nothing compared to the actual masking of the data in terms of CPU time.

From my profiles this is not the case.

lib/sender.js
ronag: This is just the fallback path. We shouldn't hit this often.

ronag:
> I think that if there are a lot of sockets all of them sending data at the same time synchronously, then this does not change much.

Why? We generate the random data asynchronously off thread so most of the overhead is off the main thread.

ronag:
Do you have any existing good benchmark we can use for before/after?

lib/sender.js
ronag: I think you are misunderstanding something here...?

lpinca:
> From my profiles this is not the case.

Is it possible to share the code/data. My tests were based on `crypto.randomBytes()` (sync) but I would be surprised if `crypto.randomFillSync()` is actually worse.

> Why? We generate the random data asynchronously/parallel off thread so most of the overhead is off the main thread.

This is not the case if we are near the end or already emptied the pool and we need data now.

> Do you have any existing good benchmark we can use for before/after?

No, but a simple echo server with a bunch of clients sending data should work.

ronag:
> This is not the case if we are near the end or already emptied the pool and we need data now.

We dispatch the next "parallel" task before the pool has emptied.

lpinca:
> We dispatch the next parallel task before the pool has emptied. `RANDOM_POOL_REFRESH`

Yes but the refill happens asynchornously.

ronag:
> > We dispatch the next parallel task before the pool has emptied. `RANDOM_POOL_REFRESH`
> 
> Yes but the refill happens asynchornously.

Yes, the idea is that there is sufficient data in the pool until the refill has completed.

lpinca:
> 256 messages

That's not much considering that that number is shared between the number of sockets, no?

sdrsdr:
Probably for really stressed servers it might be beneficial to keep two buffers filled. One that is used and one on "standby". And once we use up one start with the standby and request a new standby. For some reeeealy stressed systems a adaptive growth to the buffer size might be advised but I think this is overcomplicating things with no real world scenarios

lpinca:
Servers are not affected. Server do no mask data.

ronag:
Screenshot 2021-12-17 at 13 06 41

Here is my profile. Notice `mask` and `randomFillSync`.

ronag:
> That's not much considering that that number is shared between the number of sockets, no?

When a server connects as a client then there is not that many sockets...

ronag:
Screenshot 2021-12-17 at 13 09 04

140 ms

sdrsdr:
I can imagine a server system initiating a websocket conections to other servers for example forming a mesh but how big a mesh can be? Not really a problem in need of solving ATM

ronag:
@lpinca Does the mask need to be cryptographically random or would pseudo random numbers also be ok here?

lpinca:
> cryptographically random

This.

> Clients MUST choose a new masking key for each frame, using an
   algorithm that cannot be predicted by end applications that provide
   data.  For example, each masking could be drawn from a
   cryptographically strong random number generator.

sdrsdr:
https://stackoverflow.com/questions/14174184/what-is-the-mask-in-a-websocket-frame
```
Why is there masking at all? Because apparently there is enough broken infrastructure out there that lets the upgrade header go through and then handles the rest of the connection as a second HTTP request which it then stuffs into the cache. I have no words for this. In any case, the defense against that is basically a strong 32bit random number as masking key. Or you know… use TLS and don't use shitty proxies.
```

lpinca:
>I have `randomFillSync` calls that take up to 300 ms.

That's really weird. Is it the same with `crypto.randomBytes()` sync?

ronag:
> Is it the same with `crypto.randomBytes()` sync? I guess the underlying code is the same.

Yea, it's the same.

ronag:
I solved this in production buy doing `send(data, { mask: false })`

ronag:
> I solved this in production buy doing `send(data, { mask: false })`

Actually... that doesn't work at all.

lpinca:
Yes the server must close the connection if an unmasked frame is received.

sdrsdr:
Haha! I see a spot here for "ignore-the-standards-and-be-fast" setup if both client and server are willing to disobey the rules

ronag:
Another workaround:


```
'use strict'

const { Worker } = require('worker_threads')
const path = require('path')

class Socket {
  constructor(url) {
    this._worker = new Worker(path.join(__dirname, 'socket-worker.js'), {
      workerData: {
        url: url.href,
      },
    })
      .on('error', (err) => {
        this.onerror?.(err)
      })
      .on('message', ({ event, data }) => {
        if (event === 'open') {
          this.readyState = this.OPEN
          this.onopen?.()
        } else if (event === 'error') {
          this.readyState = null
          this.onerror?.(data)
        } else if (event === 'close') {
          this.readyState = null
          this.onclose?.()
        } else if (event === 'data') {
          this.onmessage?.({ data: Buffer.from(data) })
        }
      })

    this.OPEN = 'OPEN'

    this.onopen = null
    this.onerror = null
    this.onclose = null
    this.onmessage = null
    this.readyState = null
  }

  send(data) {
    // TODO (perf): Transfer Buffer?
    this._worker.postMessage(data)
  }

  close() {
    this._worker.terminate()
  }
}

module.exports = Socket

```

```
'use strict'

const { Worker } = require('worker_threads')
const path = require('path')

class Socket {
  constructor(url) {
    this._worker = new Worker(path.join(__dirname, 'socket-worker.js'), {
      workerData: {
        url: url.href,
      },
    })
      .on('error', (err) => {
        this.onerror?.(err)
      })
      .on('message', ({ event, data }) => {
        if (event === 'open') {
          this.readyState = this.OPEN
          this.onopen?.()
        } else if (event === 'error') {
          this.readyState = null
          this.onerror?.(data)
        } else if (event === 'close') {
          this.readyState = null
          this.onclose?.()
        } else if (event === 'data') {
          this.onmessage?.({ data: Buffer.from(data) })
        }
      })

    this.OPEN = 'OPEN'

    this.onopen = null
    this.onerror = null
    this.onclose = null
    this.onmessage = null
    this.readyState = null
  }

  send(data) {
    // TODO (perf): Transfer Buffer?
    this._worker.postMessage(data)
  }

  close() {
    this._worker.terminate()
  }
}

module.exports = Socket
```

ronag:
We're good with disabling masking and/or worker thread. Solves our perf issues.

lpinca:
@ronag are you running it on a resource constrained system? This is what I get on my machine:

```js

const buffer = Buffer.alloc(4);

let max = 0n;

for (let i = 0; i < 1e6; i++) {
  const start = process.hrtime.bigint();
  crypto.randomFillSync(buffer, 0, 4);
  const end = process.hrtime.bigint();

  const elapsed = end - start;

  if (elapsed > max) {
    max = elapsed;
  }
}

console.log(max);
```

```
$ node test.js 
2040128n
```

ronag:
Not really. It’s an EPYC server. Node 16.13

lpinca:
For correctness I have to take this https://github.com/websockets/ws/pull/1986#issuecomment-996633329 back. Recent benchmarks show that masking ~10 KiB of data in plain JavaScript is now comparable to acquiring 4 bytes of random data :/

lpinca:
If there is evidence (I'm still not convinced) that this improves things in production, then I'm fine with reopening and merging.

ronag:
I think it gets slow once the entropy pool is emptied.

PR #1988, opened 2021-12-18
add option 'clear' mask

e3dio:
There was discussion here https://github.com/websockets/ws/pull/1986 @ronag about disabling mask to improve client performance for cases when mask is not necessary, to stay spec compliant can add option to use a clear mask to skip random number generation and skip XOR on all bits, and a server will not close the connection from a missing mask

### DIFF
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -53,8 +53,8 @@ class Sender {
    * @public
    */
   static frame(data, options) {
-    const merge = options.mask && options.readOnly;
     const clear = options.mask == 'clear';
+    const merge = options.mask && options.readOnly && !clear;
     let offset = options.mask ? 6 : 2;
     let payloadLength = data.length;
 
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -42,8 +42,8 @@ class Sender {
    * @param {Object} options Options object
    * @param {Boolean} [options.fin=false] Specifies whether or not to set the
    *     FIN bit
-   * @param {Boolean} [options.mask=false] Specifies whether or not to mask
-   *     `data`
+   * @param {Boolean|String} [options.mask=false] Specifies whether or not to mask
+   *     `data`. A 'clear' value applies an empty mask to improve performance
    * @param {Number} options.opcode The opcode
    * @param {Boolean} [options.readOnly=false] Specifies whether `data` can be
    *     modified
@@ -54,6 +54,7 @@ class Sender {
    */
   static frame(data, options) {
     const merge = options.mask && options.readOnly;
+    const clear = options.mask == 'clear';
     let offset = options.mask ? 6 : 2;
     let payloadLength = data.length;
 
@@ -81,13 +82,15 @@ class Sender {
 
     if (!options.mask) return [target, data];
 
-    randomFillSync(mask, 0, 4);
+    clear ? mask.fill(0) : randomFillSync(mask, 0, 4);
 
     target[1] |= 0x80;
     target[offset - 4] = mask[0];
     target[offset - 3] = mask[1];
     target[offset - 2] = mask[2];
     target[offset - 1] = mask[3];
+
+    if (clear) return [target, data];
 
     if (merge) {
       applyMask(data, mask, target, offset, data.length);


ronag:
XORing against 0 still mutates? No?

e3dio:
A | B | A XOR B
-- | -- | --
0 | 0 | 0
0 | 1 | 1
1 | 0 | 1
1 | 1 | 0

ronag:
Sorry. Momentary confusion. 😄

lpinca:
Sorry, but I'm comfortable with this. Using a fake mask is not spec compliant. It completely defats the point of masking.

e3dio:
> Sorry, but I'm comfortable with this

Assuming not comfortable. It's easy enough to fork for special needs @ronag

ronag:
> > Sorry, but I'm comfortable with this
> 
> Assuming not comfortable. It's easy enough to fork for special needs @ronag

Already done. Thanks!

ronag:
Too bad uws and ws are on total opposite ends on this. 😞

lpinca:
We recently added the ability to skip UTF8 validation on the server (https://github.com/websockets/ws/pull/1928). This is similar. I understand that they might be desirable in a trusted environment but I don't like to keep adding them. It would also open the door for an option to accept unmasked frames.

e3dio:
`ws` already has no mask option, best thing is use no mask send on client, and edit the uWS source to accept no mask option, he might accept PR if submit it, probably just as easy to add looking now https://github.com/uNetworking/uWebSockets/blob/master/src/WebSocketProtocol.h

lpinca:
For this specific use case a possible tradeoff would be to accept a user provided mask. The downside is that we can't skip masking unless we check that all bytes are zero.

lpinca:
I mean something like this:

```js
const mask = Buffer.alloc(4);

ws.send(message, { mask });
ws.send(message, { mask });
ws.send(message, { mask });
```

e3dio:
A user provided mask is probably a good option, you can choose the level of masking you need. If you control all servers and network just use a 0 'clear' mask. If you want mask but don't need high-grade random numbers, start at random int and increase it each message. `ws` library can check for 0 'clear' mask on both send and receive to skip masking all bits when needed

alexhultman:
> Too bad uws and ws are on total opposite ends on this. 😞

Uws follows spec and does not have any options to break spec because of performance reasons. Like I said in your thread - if websockets/ws has performance problems when masking then websockets/ws has to solve their performance problems. Uws does not have any such problems.

lpinca:
FWIW `ws` switched from `Math.random()` to `crypto.randomBytes()` in https://github.com/websockets/ws/commit/7253f06f5432c76f3e82e2c055fcea08b612d8b2 as per https://github.com/websockets/ws/pull/832.

I think the overhead is not a big problem on the client. I think I've never experienced [entropy exhaustion](https://github.com/websockets/ws/pull/1986#issuecomment-997023433).

alexhultman:
There is literally no reason to use random masks, even less so cryptographically secure masks. That only applies to web browsers running untrusted scripts. Any Node.js client should use the most simple mask, as it literally improves nothing whatsoever in terms of security to use cryptographically secure masks.

e3dio:
A Node.js server can assume to be secure, but it can send user generated input across the network, if the user knows the mask they can choose the bytes to send. The use case this PR is referring to is for sending from Server to Server across your own network, where there is zero reason for a mask at all, no reason to waste any CPU cycles on your machines generating crypto numbers and masking bits

e3dio:
Also using SSL is redundant so you can disable masking, might as well make it an option for those who want it

alexhultman:
> Also using SSL is redundant so you can disable masking, might as well make it an option for those who want it

No. You can't just break spec. because you feel like it. Any message sent from a client without mask is invalid and will be refused.

What you can do, and what is still spec. compliant is to use a simple mask like a number that increments by 1 every time.

e3dio:
This PR uses a clear mask = 0 which allows skipping XOR all bits, or even better allow user supplied mask. Server will still be happy, but you could add a server check to skip unmask on mask = 0, save some CPU cycles

alexhultman:
Yes but the spec says the mask must change with every message. So to hit all the criteria of the spec you can use an incrementing 4 byte integer. That's certainly spec compliant in every single way (for non-web browsers).

e3dio:
The spec says suffeciently random numbers which means Crypto API, but the Spec is meant for web browsers, this is an option for advanced usage on non-browser, no one is forcing you to follow web browser spec exactly

alexhultman:
> The spec says sufficiently random numbers which means Crypto API, but the Spec is meant for web browsers, this is an option for advanced usage on non-browser, no one is forcing you to follow web browser spec exactly

You are literally **making things up** as you go. This is what the spec says, verbatim:

> The **server MUST close the connection upon receiving a
   frame that is not masked**.

And 

>    Clients **MUST choose a new masking key** for each frame, using an
   algorithm that **cannot be predicted by end applications that provide
   data**.  For **example**, each masking **could** be drawn from a
   cryptographically strong random number generator.  If the same key is
   used or a decipherable pattern exists for how the next key is chosen,
   the attacker can send a message that, when masked, could appear to be



Fette & Melnikov             Standards Track                   [Page 51]

RFC 6455                 The WebSocket Protocol            December 2011


   an HTTP request (by taking the message the attacker wishes to see on
   the wire and masking it with the next masking key to be used, the
   masking key will effectively unmask the data when the client applies
   it).
   
   There is nothing at all in the spec that says you MUST use cryptographically random masks that is only mentioned as **an example**. The main takeaway is that the **data provider** must not be able to guess what the mask will be, which is possible to uphold in a web browser since web browsers have a separation between data provider (the untrusted script) and the WebSocket mask generation (Google Chrome). For Node.js, the data provider is in the same sandbox as the WebSocket mask generation so that whole sentence becomes "impossible to uphold" since any script in Node.js can do literally whatever it wants.

So you're not really reading the spec. just making things up as you go.

alexhultman:
It even continues with requirements that are entirely impossible to implement in a non-web browser:

> It is also necessary that once the transmission of a frame from a
   client has begun, the payload (application-supplied data) of that
   frame **must not be capable of being modified by the application**.
   Otherwise, an attacker could send a long frame where the initial data
   was a known value (such as all zeros), compute the masking key being
   used upon receipt of the first part of the data, and then modify the
   data that is yet to be sent in the frame to appear as an HTTP request
   when masked.  (This is essentially the same problem described in the
   previous paragraph with using a known or predictable masking key.)

This requirement is **impossible** to implement in a programming language or platform that has no sandboxing between the business logic and the WebSocket implementation. For instance, you can easily get the net.Socket from a WebSocket and just straight up inject anything you want there, completely bypassing all the cryptographically random masking. So all of this "secure" randomness is completely pointless for all implementations that are not **web browsers**. Only web browsers have this security model implemented.

e3dio:
> There is nothing at all in the spec that says you MUST use cryptographically random masks. You're not really reading the spec. just making things up as you go

Have you looked at the spec at all ?

RFC 6455, page 33:

> the masking key MUST be derived from a strong
   source of entropy..  RFC 4086 [RFC4086] discusses what
   entails a suitable source of entropy

RFC 4086 describes entropy sources as used in Crypto API. 

The spec is for web browsers to safely implement websockets over cleartext public networks. We are talking about advanced usage for non-browsers over secure or private networks. Your web server will not be affected, it won't know any difference. However if you want to optimize you are welcome to check for mask = 0 to skip unmasking every byte in every message

alexhultman:
>    The masking key is a 32-bit value chosen at random by the client.
   When preparing a masked frame, the client MUST pick a fresh masking
   key from the set of allowed 32-bit values.  The masking key needs to
   be unpredictable; thus, the masking key MUST be derived from a strong
   source of entropy, and the masking key for a given frame MUST NOT
   make it simple for a server/proxy to predict the masking key for a
   subsequent frame.  The unpredictability of the masking key is
   essential to prevent authors of malicious applications from selecting
   the bytes that appear on the wire.  RFC 4086 [RFC4086] discusses what
   entails a suitable source of entropy for security-sensitive
   applications.

Well in that case, the spec. is impossible to implement for non-browsers without setting up some kind of sandboxed mess where some root process runs the websocket implementation and the application runs as regular user, only being able to send payload to the root process or something crazy like that, which is not how things are actually made in the real world so might as well just ignore the whole section on masking and just put a simple, static or easily changing mask. None of this makes any sense for non-web browsers and it doesn't improve security whatsoever to use a high grade cryptographically random mask generator.

e3dio:
I think we already decided on user-supplied mask, which personally I will be setting as 0 'clear' mask for server-to-server communication on private network. The only question is are you going to add a server check for 0 mask to skip unmasking every bit sent with every message

alexhultman:
Btw, I found this in spec.

>    frame-masking-key       = 4( %x00-FF )
>                              ; present only if frame-masked is 1
>                              ; 32 bits in length

So a mask of 0 is indeed allowed.

And again,

> The unpredictability of the masking key is
   essential to prevent authors of malicious applications from **selecting
   the bytes that appear on the wire**.

This goes out the window when net.Socket.write can be accessed to emit literally anything, anyways.

e3dio:
> So a mask of 0 is indeed allowed

Might as well add a check on the server to skip unmask for that case, I don't know how much CPU it saves but it is some amount, especially if the client happens to send many of these

> This goes out the window when net.Socket.write can be accessed

I don't think an attacker needs top level script access, they just need to supply the data that gets sent across the network, which a mask can prevent. But literally the only reason there is a mask is to protect broken http proxies that don't understand websocket protocol from a cache poison attack. If you use SSL there is no reason for mask, if you are on local network there is no reason for mask, if you are sending large amounts of messages I think you should have the option of sending a mask = 0, or a user supplied math.random etc

alexhultman:
> I don't think an attacker needs top level script access

websockets/ws runs in Node.js (primarily?) so it doesn't matter if it so uses an Uranium based dedicated randomness module from the Greek gods - the fact you have net.Socket.write access entirely circumvents any security whatsoever from randomness.

e3dio:
Looks like @lpinca added a nice PR for user supplied mask https://github.com/websockets/ws/pull/1989/files

ronag:
I guess running ws in a worker could improve that security part?

PR #1989, opened 2021-12-19
Add ability to specify the masking key

lpinca:
Refs: https://github.com/websockets/ws/pull/1986
Refs: https://github.com/websockets/ws/pull/1988

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -462,8 +462,9 @@ is a noop if the ready state is `CONNECTING` or `CLOSED`.
 
 - `data` {Array|Number|Object|String|ArrayBuffer|Buffer|DataView|TypedArray} The
   data to send in the ping frame.
-- `mask` {Boolean} Specifies whether `data` should be masked or not. Defaults to
-  `true` when `websocket` is not a server client.
+- `mask` {Boolean|Buffer} Specifies whether `data` should be masked or not. If a
+  `Buffer` is provided, then its contents is used as the masking key. Defaults
+  to `true` when `websocket` is not a server client.
 - `callback` {Function} An optional callback which is invoked when the ping
   frame is written out. If an error occurs, the callback is called with the
   error as its first argument.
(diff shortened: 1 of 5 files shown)


lpinca:
cc: @ronag @e3dio

ronag (approved):

lpinca:
There are two issues with this

1. There is no way to provide a masking key for `ws.close()`.
2. The user provided masking key is not copied. If data is buffered and the user mutates the `Buffer`'s contents, then the masking key might not be the expected one. For example:

    ```js
    const mask = Buffer.alloc(4);

    mask.fill('a');

    ws.send(message, { mask });

    mask.fill('b');

    ws.send(message, { mask });

    mask.fill('b');
    ```

    In this case, all buffered messages will be masked with the same masking key. I'm not sure if it is better to document this or  copy the user provided key. What do you think?

lpinca:
Another possible alternative, which might be simpler, could be a `randomMask` option on `WebSocket` constructor. The option specifies a user provided function to fill the mask buffer.

ronag:
I think copying should also be fine. It’s 4 bytes.

alexhultman:
> I think copying should also be fine. It’s 4 bytes.

Hehe, yes. A reference is **at least** 8 bytes so a copy is definitely fine.

e3dio:
I would not want to require adding mask details to all my ws.send() calls individually, a `new WebSocket(url, { options })` mask option would be better, a function that takes the mask buffer as parameter would be good:
```javascript
const ws = new WebSocket(url, {
  maskBuffer(buf) {
    require('crypto').randomFillSync(buf, 0, 4) // default function
  }
})

const ws = new WebSocket(url, {
  maskBuffer(buf) {
    buf.writeUInt32BE(Math.random()*2**32) // Math.random example
  }
})

const ws = new WebSocket(url, {
  maskBuffer(buf) {
    buf.fill(0) // clear mask example
  }
})
```
Any ws.send() calls with a mask specified could override the maskBuffer function

ronag:
I’m good with either option.

e3dio:
FYI @alexhultman added clear mask checking on already optimized uWS server https://github.com/uNetworking/uWebSockets/commit/0c8be4d20ccce75be632dd8aae109b68485ccb40 and is reporting 10% performance improvement for large messages which is pretty good so it's ready and compatible with this PR, and you will probably see better improvement for this Node client which currently does crypto and bit xor

lpinca:
See https://github.com/websockets/ws/pull/1990.

lpinca:
Closing in favor of #1990.

PR #1990, opened 2021-12-19
 Introduce the `generateMask` option

lpinca:
The `generateMask` option specifies a function that can be used to
generate custom masking keys.
    
Refs: https://github.com/websockets/ws/pull/1986
Refs: https://github.com/websockets/ws/pull/1988
Refs: https://github.com/websockets/ws/pull/1989

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -270,6 +270,10 @@ This class represents a WebSocket. It extends the `EventEmitter`.
 - `options` {Object}
   - `followRedirects` {Boolean} Whether or not to follow redirects. Defaults to
     `false`.
+  - `generateMask` {Function} The function used to generate the masking key. It
+    takes a `Buffer` that must be filled synchronously and is called before a
+    message is sent, for each message. By default the buffer is filled with
+    cryptographically strong random bytes.
   - `handshakeTimeout` {Number} Timeout in milliseconds for the handshake
     request. This is reset after every redirection.
   - `maxPayload` {Number} The maximum allowed message size in bytes.
--- a/lib/receiver.js
+++ b/lib/receiver.js
@@ -417,7 +417,13 @@ class Receiver extends Writable {
       }
 
       data = this.consume(this._payloadLength);
-      if (this._masked) unmask(data, this._mask);
+
+      if (
+        this._masked &&
+        (this._mask[0] | this._mask[1] | this._mask[2] | this._mask[3]) !== 0
+      ) {
+        unmask(data, this._mask);
+      }
     }
 
     if (this._opcode > 0x07) return this.controlMessage(data);
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -65,8 +65,26 @@ class Sender {
    * @public
    */
   static frame(data, options) {
-    const merge = options.mask && options.readOnly;
-    let offset = options.mask ? 6 : 2;
+    let mask;
+    let merge = false;
+    let offset = 2;
+    let skipMasking = false;
+
+    if (options.mask) {
+      mask = options.maskBuffer || maskBuffer;
+
+      if (options.generateMask) {
+        options.generateMask(mask);
+      } else {
+        randomFillSync(mask, 0, 4);
+      }
+
+      skipMasking = (mask[0] | mask[1] | mask[2] | mask[3]) === 0;
+      if (options.readOnly && !skipMasking) merge = true;
+
+      offset = 6;
+    }
+
     let payloadLength = data.length;
 
     if (data.length >= 65536) {
@@ -93,19 +111,13 @@ class Sender {
 
     if (!options.mask) return [target, data];
 
-    const mask = options.maskBuffer ? options.maskBuffer : maskBuffer;
-
-    if (options.generateMask) {
-      options.generateMask(mask);
-    } else {
-      randomFillSync(mask, 0, 4);
-    }
-
     target[1] |= 0x80;
     target[offset - 4] = mask[0];
     target[offset - 3] = mask[1];
     target[offset - 2] = mask[2];
     target[offset - 1] = mask[3];
+
+    if (skipMasking) return [target, data];
 
     if (merge) {
       applyMask(data, mask, target, offset, data.length);
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -11,7 +11,7 @@ const { EMPTY_BUFFER } = require('./constants');
 const { isValidStatusCode } = require('./validation');
 const { mask: applyMask, toBuffer } = require('./buffer-util');
 
-const mask = Buffer.alloc(4);
+const maskBuffer = Buffer.alloc(4);
 
 /**
  * HyBi Sender implementation.
@@ -22,9 +22,17 @@ class Sender {
    *
    * @param {(net.Socket|tls.Socket)} socket The connection socket
    * @param {Object} [extensions] An object containing the negotiated extensions
+   * @param {Function} [generateMask] The function used to generate the masking
+   *     key
    */
-  constructor(socket, extensions) {
+  constructor(socket, extensions, generateMask) {
     this._extensions = extensions || {};
+
+    if (generateMask) {
+      this._generateMask = generateMask;
+      this._maskBuffer = Buffer.alloc(4);
+    }
+
     this._socket = socket;
 
     this._firstFragment = true;
@@ -42,8 +50,12 @@ class Sender {
    * @param {Object} options Options object
    * @param {Boolean} [options.fin=false] Specifies whether or not to set the
    *     FIN bit
+   * @param {Function} [options.generateMask] The function used to generate the
+   *     masking key
    * @param {Boolean} [options.mask=false] Specifies whether or not to mask
    *     `data`
+   * @param {Buffer} [options.maskBuffer] The buffer used to store the masking
+   *     key
    * @param {Number} options.opcode The opcode
    * @param {Boolean} [options.readOnly=false] Specifies whether `data` can be
    *     modified
@@ -81,7 +93,13 @@ class Sender {
 
     if (!options.mask) return [target, data];
 
-    randomFillSync(mask, 0, 4);
+    const mask = options.maskBuffer ? options.maskBuffer : maskBuffer;
+
+    if (options.generateMask) {
+      options.generateMask(mask);
+    } else {
+      randomFillSync(mask, 0, 4);
+    }
 
     target[1] |= 0x80;
     target[offset - 4] = mask[0];
@@ -156,6 +174,8 @@ class Sender {
         rsv1: false,
         opcode: 0x08,
         mask,
+        maskBuffer: this._maskBuffer,
+        generateMask: this._generateMask,
         readOnly: false
       }),
       cb
@@ -200,6 +220,8 @@ class Sender {
         rsv1: false,
         opcode: 0x09,
         mask,
+        maskBuffer: this._maskBuffer,
+        generateMask: this._generateMask,
         readOnly
       }),
       cb
@@ -244,6 +266,8 @@ class Sender {
         rsv1: false,
         opcode: 0x0a,
         mask,
+        maskBuffer: this._maskBuffer,
+        generateMask: this._generateMask,
         readOnly
       }),
       cb
@@ -299,6 +323,8 @@ class Sender {
         rsv1,
         opcode,
         mask: options.mask,
+        maskBuffer: this._maskBuffer,
+        generateMask: this._generateMask,
         readOnly: toBuffer.readOnly
       };
 
@@ -314,6 +340,8 @@ class Sender {
           rsv1: false,
           opcode,
           mask: options.mask,
+          maskBuffer: this._maskBuffer,
+          generateMask: this._generateMask,
           readOnly: toBuffer.readOnly
         }),
         cb
@@ -331,8 +359,12 @@ class Sender {
    * @param {Number} options.opcode The opcode
    * @param {Boolean} [options.fin=false] Specifies whether or not to set the
    *     FIN bit
+   * @param {Function} [options.generateMask] The function used to generate the
+   *     masking key
    * @param {Boolean} [options.mask=false] Specifies whether or not to mask
    *     `data`
+   * @param {Buffer} [options.maskBuffer] The buffer used to store the masking
+   *     key
    * @param {Boolean} [options.readOnly=false] Specifies whether `data` can be
    *     modified
    * @param {Boolean} [options.rsv1=false] Specifies whether or not to set the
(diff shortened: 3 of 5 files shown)


lib/receiver.js
```diff
@@ -417,7 +417,13 @@ class Receiver extends Writable {
       }
 
       data = this.consume(this._payloadLength);
-      if (this._masked) unmask(data, this._mask);
+
+      if (
+        this._masked &&
+        (this._mask[0] | this._mask[1] | this._mask[2] | this._mask[3]) !== 0
```
lpinca, line 423: I think this optimization does not make sense. It is actually counterproductive in the common case. If a random mask is used, then the probability of this event is 1 / 2**32. The same applies for the same optimization on `lib/sender`.

lib/sender.js
```diff
@@ -81,14 +86,24 @@ class Sender {
...
+      randomFillSync(_mask, 0, 4);
+    }
 
     target[1] |= 0x80;
     target[offset - 4] = mask[0];
     target[offset - 3] = mask[1];
     target[offset - 2] = mask[2];
     target[offset - 1] = mask[3];
 
+    if ((mask[0] | mask[1] | mask[2] | mask[3]) === 0) {
```
lpinca, line 103: I think this optimization does not make sense. It is actually counterproductive in the common case. If a random mask is used, then the probability of this event is 1 / 2**32. The same applies for the same optimization on `lib/receiver.js`.

ronag: I don’t think it hurts and allows for hot path optimization for clients that know the hotel path “trick”.

ronag (approved):

lib/sender.js
e3dio: I benchmarked this line `(mask[0] | mask[1] | mask[2] | mask[3]) === 0` and it came out to less than 1 nanosecond, which is basically zero length of time. I benchmarked the time to mask a 10kb string with `applyMask(data, mask, target, offset, data.length)` and it came to 1.4 milliseconds, 3 million times larger. Masking 100 10kb strings would take 140ms, checking for mask 100 times would 50 nano seconds. The larger the string, the larger the mask time, while checking mask stays constant <1 nanosecond. That being said, if you want the best performance you don't need to check the mask every ws.send, you only need to check it on initial websocket setting, but currently the PR has it a function that must be generated every ws.send. Though 1 nanosecond may be worth the effort if you want the simplified generateMask function option

e3dio: If you want best performance you should accept a Buffer or Function for `generateMask` option

e3dio: Probably the biggest performance issue is repeatedly creating new Buffer objects in `generateMask`, this should be re-coded to allow accepting a Buffer option, and the Function method should modify the existing Buffer instead of generate new one as suggested here https://github.com/websockets/ws/pull/1989#issuecomment-997392448

lib/sender.js
```diff
@@ -81,14 +86,24 @@ class Sender {
 
     if (!options.mask) return [target, data];
 
-    randomFillSync(mask, 0, 4);
+    let mask = _mask;
+
+    if (options.generateMask) {
+      mask = options.generateMask();
```
e3dio, line 92: This generates a new Buffer object on every send, instead it should modify existing buffer

e3dio:
This stuff is not really required for most people, anyone who knows what they need can modify the couple lines of code, so I can see an argument for not even including this, feel free to close if wanted, or keep anything you want there

lib/sender.js
alexhultman: Yeah it's a totally invalid argument to make that checking if an uint32_t is 0 (one jz instruction in constant time) would be too costly, while having an API that does garbage collected memory allocations in the hot path (literally gazillions of instructions in nondeterministic time).

alexhultman: That's like arguing about the weight of a toothpick, when moving a house.

e3dio:
I checked npm and this library is downloaded 48 million times every week which is crazy. You have to consider all the people using this, 99% probably have no idea what do with a config option that says Modify this buffer or supply a new Buffer to update the mask. If you want to make it easy for people you could just pick the 3 most common mask options: `{ mask: 'crypto' }` or `{ mask: 'random' }` or `{ mask: 'clear' }` plus `mask: true` and `mask: false`, true defaults to crypto, that would reduce the complexity to about zero and I'd probably prefer it myself

lib/sender.js
lpinca: I initially called the option `fillMaskBuffer`, and made it take the mask buffer like `crypto.randomFillSync()`. Then I changed my mind and went for `generateMask`. The reason is that it is simpler to use. The user does not have to remember to fill the buffer. They are not forced to do that but we should then document that the last one is used even if the `WebSocket` instance is different. Not returning a buffer results in a thrown error.

Finally, `generateMask` does not force the user to return a new `Buffer` instance every time. The user can return the same instance every time:

```js
const mask = Buffer.alloc(4);

const ws = new Websocket(url, {
  generateMask() {
    mask.fill(0);
    return mask;
  }
});
```

lib/sender.js
lpinca: Not necessarily.

ronag:
LGTM! Thanks!

lib/sender.js
ronag: Isn't this sufficient? I don't think `ws` will mutate the buffer, right?

```js
const mask = Buffer.alloc(4);

const ws = new Websocket(url, {
  generateMask() {
    return mask;
  }
});
```

lpinca: Correct.

lib/sender.js
e3dio: It makes the config setting awkward to require maintaining the buffer outside the setting, it would be more convenient and succinct to pass the buffer to the function: 
```javascript
new Websocket(url, { generateMask: buf => require('crypto').randomFillSync(buf) } )
```
and then allow passing a Buffer instead of a function if you don't plan on updating it, if you want to optimize:
```javascript
new Websocket(url, { generateMask: Buffer.alloc(4) } )
```

lpinca: > it would be more convenient and succinct to pass the buffer to the function.

The only reason why I did not do that is that the buffer is shared between `WebSocket` instances and this might generate confusion even it is documented because people do not read.

ronag: Does it need to be shared?

lpinca: No, it only saves some bytes.

e3dio: If you want to use the clear mask you could pass an empty function, which will keep the Buffer at its default 0 value:
```javascript
new Websocket(url, { generateMask: buf => { } } )
```

lpinca: > If you want to use the clear mask you could pass an empty function, which will keep the Buffer at its default 0 value

This is actually what made me use a function that returns a `Buffer`, because I wrote the test include here that way, and it was failing due to https://github.com/websockets/ws/pull/1990#discussion_r772217927.

e3dio: I think it is ok to have a mask Buffer per WebSocket connection, vs sharing 1 Buffer between all WebSocket connections

lpinca: It requires more changes because I don't want to add a mask buffer also for server clients and currently the `Sender` is server/client agnostic.

lpinca: There is also the static `Sender.frame()` method. It gets kinda messy.

e3dio: You only need to create an extra Buffer if the Sender has the config option `generateMask` set, otherwise you use the shared buffer

lpinca: See https://github.com/websockets/ws/pull/1990/commits/8ab0db4fa664ac8049db2ccef80e94da73b67944

ronag (approved):

lib/sender.js
e3dio: Looks like exactly what is needed for that, this would be a great feature nice work

doc/ws.md
```diff
@@ -270,6 +270,9 @@ This class represents a WebSocket. It extends the `EventEmitter`.
 - `options` {Object}
   - `followRedirects` {Boolean} Whether or not to follow redirects. Defaults to
     `false`.
+  - `generateMask` {Function} The function used to generate the masking key. It
```
e3dio, line 273: May want to clarify this function gets called before every send, the current definition could mean it only calls once to initialize

ronag (commented):
I think allocating target is not necessary on the === 0 fast path?

lpinca:
> I think allocating target is not necessary on the === 0 fast path?

It is but `target` length should be `offset`. This is not the case if the `merge` is true. It must be fixed. `merge` should be false if the masking key is zero.

e3dio:
I had to come back and fix that on my 1st PR https://github.com/websockets/ws/pull/1988/commits/9a226892498dc87639f4a53de3bdfb04f53189c4

lpinca:
Will fix it later, when I get a chance.

lib/sender.js
lpinca: @e3dio 

> I benchmarked this line `(mask[0] | mask[1] | mask[2] | mask[3]) === 0` and it came out to less than 1 nanosecond

How can you get that precision? Two sequential `process.hrtime.bigint()` with nothing between take ~0.1 milliseconds / ~100000 nanoseconds on my machine.

lpinca:
Thank you for the reviews.

lib/sender.js
e3dio: When I do 1 iteration I get 2000 nanosec, doing a loop benchmark gets you to .4 nanosec, I assume JS engine is going to optimize that code for a hot path (hot vs slow execution), not sure how you are getting 100,000 nano sec for 1 iteration though that seems off, are you sure you have the numbers right? 0.1 milliseconds = 100,000,000 nanoseconds so I think you have wrong number
```javascript
const mask = Buffer.from([1,2,3,4]);
const start = process.hrtime.bigint();
(mask[0] | mask[1] | mask[2] | mask[3]) === 0;
console.log(process.hrtime.bigint() - start)
```

e3dio: ok you got 100,000 nanosec, did you try the above code?

e3dio: I tested the `applyMask` function again, looks like in my 1st post I gave the 1 iteration benchmark, so to compare correctly against mask check I should have done the loop benchmark also, I am getting 14,000 nanosec for `applyMask`, so that function is not that bad, only 35,000 times worse than mask check, not 2 million times worse, I should update that post

e3dio: But probably the most impactful thing is skipping the crypto generation which I haven't even tried benchmark yet

alexhultman: Why don't you make an uint32array from the memory and just check if (a[0])

e3dio: I tried `buf.readInt32BE()===0` it gives 1.81 nanosec vs the existing is .45 nanosec. .45 is hard to beat, that is including the integer iteration which probably takes up a chunk. 8700k CPU is 4.70GHz turbo so that's almost 5 cpu cycles per nanosecond, .45ns takes up 2 cpu cycles, hard to beat that

alexhultman: Buffer.readUInt32BE is implemented as a _function_ by Node.js guys. Uint32Array is an _intrinsic_ to the very V8 engine. It should be just a jz. But, in any case, this is just nerd talk as we are literally discussing an if-statement involving 4 bytes. None of this matters in the given context.

e3dio: I was going to say this is the most insignificant thing possible to discuss improving on .45ns but you were the one that brought it up so I gave it a try

alexhultman: Ah, my idea was more about _simplifying the code_. You don't need to OR together 4 individual bytes if you simply have an uint32array. Passing an uint32array around is also simpler to fill, since you can just pick a random integer between 0 and uint32_max (which is exactly what rand() returns). Point being, an uint32 is simpler dealing with IMO. Never mind.

alexhultman: uint32 is also simpler dealing with if you want to have a mask that just increments by 1 every time. IMO uint32 is the perfect data type for the mask.

e3dio: Ok sorry I see what you are saying, pass around Uint32Array's instead of Buffers for the mask. You were throwing me off randomly commenting in this thread, he was asking me about benchmarking what amounts to an equals statement which is non-productive, but using Uint32Array is a valid point
```javascript
uint32 = new Uint32Array(1) // new mask default value = 0

{ generateMask: uint32 => require('crypto').randomFillSync(uint32) }
{ generateMask: uint32 => uint32[0] = Math.random()*2**32 }
{ generateMask: uint32 => uint32[0]++ }

const skipMasking = uint32[0] === 0;
```
They are easier to work with than Buffer because the mask is 32 bit unsigned int. Maybe only reason to use Buffer is more people are familiar with Buffer? You could do either one, Uint32Array makes more sense, you just have to know how it works. Buffer is easier to say than Uint32Array. Ideally Uint32Array but Buffer is ok

lpinca: > Maybe only reason to use Buffer is more people are familiar with Buffer?

Yes, and because `bufferUtil.mask()` was written to take a buffer at the time https://github.com/websockets/bufferutil/blob/cd48704aa8e87c19d7497812d466fd6fb6a28c09/src/bufferutil.cc#L89-L90, when typed array were not supported, and that did not change over the time. We could make it take a `Uint32Array` but I do not think it is worth the effort.

e3dio: > How can you get that precision?

@lpinca this is the code I use for loop benchmark:
```javascript
const iterations = 1e9
const cycleNS = null

const fn=()=>{
   // code to benchmark
}

let i=iterations,start=process.hrtime.bigint()
while(i--)fn()
const ns=Number(process.hrtime.bigint()-start)/iterations
console.log(ns.toFixed(2),'ns')
cycleNS&&console.log((ns/cycleNS).toFixed(2),'cycles')
```
You want to adjust the iteration number `1e9` so it runs for at least a second so the JS engine optimizes the function. You can count how many cpu cycles a function takes, the nanoseconds fall on clear intervals of the cpu cycle time. If you bench an empty function that should take 2 cycles, divide the result by 2 for your cycle time, for me 1 cycle is .228 nanoseconds, which happens to match my CPU speed 4.7GHz = 4.7 cycles per nanosecond. Change `cycleNS` to your cycle time to add # cycles to the result log

Using this you can see that Uint32Array can be faster to work with than Buffer, here are some cpu cycles per expression:
```javascript
uint32[0] = 123 // 2 cycles
buf.writeUInt32BE(123) // 11 cycles
buf[0] = 1; buf[1] = 2; buf[2] = 3; buf[3] = 4 // 5 cycles

uint32[0] // 2 cycles
buf.readUInt32BE() // 8 cycles
buf[0]; buf[1]; buf[2]; buf[3] // 2 cycles

Math.random() // 30 cycles
crypto.randomFillSync(buf) // 6200 cycles
```
These numbers are so small they don't really matter but it is interesting to see how it works

PR #1991, opened 2021-12-20
fix target

e3dio:
move a couple things above merge


lpinca:
I did pretty much the same in the latest push. Thank you.

PR #1994, opened 2021-12-30
[doc] Clarify interpretation of `verifyClient`

jugglinmike:
Prior to this commit, the documentation for the `verifyClient` option
was somewhat confusing due to its use of the passive voice and an
uncommon interpretation of the term "arguments."

Explain the way `ws` interprets the value of `verifyClient` using an
active voice and more traditional meanings of "parameters" and
"arguments."

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -110,8 +110,8 @@ has a single parameter then `ws` will invoke it with the following argument:
 The return value (`Boolean`) of the function determines whether or not to accept
 the handshake.
 
-If `verifyClient` has two parameters then `ws` will invoke it with the
-following arguments:
+If `verifyClient` has two parameters then `ws` will invoke it with the following
+arguments:
 
 - `info` {Object} Same as above.
 - `cb` {Function} A callback that must be called by the user upon inspection of
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -110,7 +110,7 @@ has a single parameter then `ws` will invoke it with the following argument:
 The return value (`Boolean`) of the function determines whether or not to accept
 the handshake.
 
-if `verifyClient` has two parameters then `ws` will invoke it with the
+If `verifyClient` has two parameters then `ws` will invoke it with the
 following arguments:
 
 - `info` {Object} Same as above.
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -99,7 +99,7 @@ to share a single HTTP/S server between multiple WebSocket servers.
 > more details.
 
 If `verifyClient` is not set then the handshake is automatically accepted. If it
-is provided with a single argument then that is:
+has a single parameter then `ws` will invoke it with the following argument:
 
 - `info` {Object}
   - `origin` {String} The value in the Origin header indicated by the client.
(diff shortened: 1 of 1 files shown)


doc/ws.md
```diff
@@ -110,7 +110,8 @@ is provided with a single argument then that is:
 The return value (`Boolean`) of the function determines whether or not to accept
 the handshake.
 
-if `verifyClient` is provided with two arguments then those are:
+if `verifyClient` has two parameters then `ws` will invoke it with the
```
lpinca, line 113:
```suggestion
If `verifyClient` has two parameters then `ws` will invoke it with the
```

lpinca (approved):

lpinca:
Can you please fix the lint issues? Thanks.

jugglinmike:
Done. Odd that the CI system didn't attempt to validate the original commit.

By the way, there's no documentation about how to do that or that it's necessary at all. This is typically the job of a contributing guide, though even just a dedicated npm script would give more savvy contributors a hint about how they can satisfy the project's requirements.

lpinca:
> Done. Odd that the CI system didn't attempt to validate the original commit.

CI does not run for first-time contributors. It must be approved.

> By the way, there's no documentation about how to do that or that it's necessary at all. This is typically the job of a contributing guide, though even just a dedicated npm script would give more savvy contributors a hint about how they can satisfy the project's requirements.

True, there is actually a dedicated [npm script](https://github.com/websockets/ws/blob/8.4.0/package.json#L36). We could use `husky` + `lint-staged` to improve the contributor experience.

lpinca:
Thank you.

jugglinmike:
> True, there is actually a dedicated [npm script](https://github.com/websockets/ws/blob/8.4.0/package.json#L36). We could use `husky` + `lint-staged` to improve the contributor experience.

Yeah! That script is what the CI environment uses for validation. It doesn't correct violations, and it doesn't provide any detail about the linting violations--only that linting violations are present:

>     Run npm run lint
>     
>     > ws@8.4.0 lint
>     > eslint --ignore-path .gitignore . && prettier --check --ignore-path .gitignore "**/*.{json,md,yaml,yml}"
>     
>     Checking formatting...
>     [warn] doc/ws.md
>     [warn] Code style issues found in the above file(s). Forgot to run Prettier?
>     Error: Process completed with exit code 1.

Here's the command I needed to run to have the necessary changes made:

    $ ./node_modules/.bin/prettier "**/*.{json,md,yaml,yml}" --write

> Thank you.

Thanks for maintaining the project!

lpinca:
> It doesn't correct violations, and it doesn't provide any detail about the linting violations

There are actually some details.

```
[warn] doc/ws.md
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

There were issues only on `doc/ws.md` so running only `npx prettier --write doc/ws.md` would have been faster. This is why I did not add a script that runs eslint and prettier on all files and fixes everything. It might be convenient but it does not make much sense when there are only a couple of files with lint issues. The `husky` + `lint-staged` solution is a little better because only staged files are linted/fixed but it also has some limitations. It does not work if the user does not run `npm install` and sometimes it is not necessary, for example for small doc changes like these.

Maybe having both `husky` + `lint-staged` and a "fix everything" script is the best thing to do, I do not know, I'm not convinced :).

lpinca:
A PR to add contributing guidelines is welcome!

PR #1998, opened 2022-01-04
perf: try to avoid buffer allocations

ronag:


lib/sender.js
```diff
@@ -85,7 +85,8 @@ class Sender {
       offset = 6;
     }
 
-    let payloadLength = data.length;
+    const dataLength = Buffer.byteLength(data);
```
lpinca, line 88: I'm not sure I understand, everything is already a buffer here.

ronag: No it's not, that's the point. See https://github.com/websockets/ws/pull/1998/files#diff-ee0f1feef959a4232309df5db73baad8cd2b6d2d5be97f3ef6565a29093c61a8R353

lpinca: Correct I did not read the `send()` changes below. Anyway I think there is an implicit conversion to `Buffer` when writing to the socket?

ronag: Yes, however that happens in native code and multiple chunks are converted at the same time. Less allocations and better perf.

lpinca: Hmm, unconvinced, `writevGeneric()` was  optmized for the "all buffers" case. Do you have benchmarks?

lpinca: Also I think this crashes if the user tries to write an `Array` or `Object`.

lib/sender.js
```diff
@@ -347,7 +350,7 @@ class Sender {
       }
     } else {
       this.sendFrame(
-        Sender.frame(buf, {
+        Sender.frame(data, {
```
lpinca, line 353: If `toBuffer()` is skipped here, then `toBuffer.readOnly` might not be the expected one. For example, it might be `true` when writing a string leading to an unnecessary copy of the data when masking the data.

```js
ws.send(Buffer.from('foo'));
ws.send('bar');
```

ronag:
```js
const net = require('net')

const server = net.createServer((socket) => {
  socket.resume()
  socket.on('error', () => {})
}).listen(0)

const socket = net.connect(server.address().port, '0.0.0.0', () => {
  let str = '1'.repeat(4096)
  for (let n = 0; n < 1024*128; ++n) {
    socket.write(str)
    // socket.write(Buffer.from(str))
  }
  socket.end()
  socket.on('error', () => {})
  socket.on('finish', () => {
    server.close()
  })
})
```

```
ws$ time node bench.js 
real    0m0.866s
user    0m0.686s
sys     0m0.243s

ws$ time node bench.js 
real    0m0.319s
user    0m0.144s
sys     0m0.204s
```

It's more than 2x faster without `Buffer.from`.

lib/sender.js
ronag: https://github.com/websockets/ws/pull/1998#issuecomment-1005582961

ronag: > Also, I think this crashes if the user tries to write an Array or Object.

Should that be possible? Also I think it already crashes if you do that. How does the PR change the behavior?

lib/sender.js
ronag: Fixed

lib/sender.js
lpinca:
```js
Buffer.byteLength([1,2,3]);
Buffer.byteLength(new String('foo'));
```

throws.

lib/sender.js
lpinca: Now it ignores the `readOnly` info, so it might mutate a user provided `Buffer`

lib/sender.js
ronag: Fixed

lib/sender.js
```diff
@@ -80,12 +80,24 @@ class Sender {
...
+    let dataLength;
+    if (typeof data === 'string') {
+      if (!options.mask || skipMasking) {
+        dataLength = Buffer.byteLength(data);
+      } else {
+        data = toBuffer(data);
+        dataLength = data.length;
+      }
+    } else {
+      dataLength = data.length;
```
lpinca, line 96: `data` can be a `DataView` an `ArrayBuffer`, etc. Writing them to a socket is not supported.

lib/sender.js
```diff
@@ -80,12 +80,24 @@ class Sender {
...
+      merge = options.mask && options.readOnly && !skipMasking;
```
lpinca, line 97: https://github.com/websockets/ws/pull/1998/files#r778324069

lib/sender.js
ronag: If `data` is not a string then it's expected to be a buffer, just like before.

lib/sender.js
ronag: ?

lib/sender.js
ronag: now?

lib/sender.js
ronag: i.e. nothing has changed in terms of behaviour other than we check for string type.

lpinca: Maybe I am misreading but this change

```js
const buf = perMessageDeflate ? toBuffer(data) : data;
```

in `send()` keep data as is if permessage-deflate is disabled.

lib/sender.js
lpinca: Still ignored if `permessage-deflate` is disabled.

lib/sender.js
ronag: You were right. Fixed.

lib/sender.js
ronag: So we only ignore it if `buf` is `string` which is only true if `!perMessageDeflate && typeof data === 'string'`

ronag: I don't see a problem atm.

lib/sender.js
```diff
@@ -303,8 +315,9 @@ class Sender {
    * @public
    */
   send(data, options, cb) {
-    const buf = toBuffer(data);
     const perMessageDeflate = this._extensions[PerMessageDeflate.extensionName];
+    const buf =
+      perMessageDeflate || typeof data !== 'string' ? toBuffer(data) : data;
```
lpinca, line 320: The `buf.length >= perMessageDeflate._threshold` check below might not work as expected when `data` is a string because it does not use the string bytes length.

---

Also, when `data` is a string and permessage-deflate is enabled, `toBuffer()` should be ideally called only when the the message will actually be compressed (`this._compress === true`).

It would be better to make this permessage-deflate agnostic. Removing the `perMessageDeflate` check here and passing the string to `perMessageDeflate.compress()` should work. The problem is that `enqueue()` and `dequeue()` currently use the buffer length to update the `_bufferedBytes` value.

ronag: That branch is only used if `perMessageDeflate` in which case it's converted to a buffer.

ronag: While I agree that we could improve it, I don't think there is an issue here currently. Only that it's a little hard to read.

lpinca:
I think the benchmark is a bit fabricated. It writes 512 MiB synchronously by calling `socket.write()` 1024 * 128 times. There are `1024 * 128` allocations when using the `toBuffer()` variant but a lot less when using the string variant because there is only one allocation for the data buffered in the socket writable queue. Also in the string variant the string is always the same. There is basically no memory usage.

ronag:
> I think the benchmark is a bit fabricated. It writes 512 MiB synchronously by calling socket.write() 1024 * 128 times. There are 1024 * 128 allocations when using the toBuffer() variant but a lot less when using the string variant because there is only one allocation for the data buffered in the socket writable queue. Also in the string variant the string is always the same. There is basically no memory usage.

I mean the point here is all of those buffer allocations are not necessary. The strings are already allocated. This just demonstrates the overhead of creating all of those buffers.

lib/sender.js
lpinca: Correct. Anyway atm this improvement only works when permessage-deflate is disabled. What about `websocket.ping()` and `websocket.pong()`?

ronag:
Even with smaller values e.g. 4 MiB there is improvement:

```
ws$ time node bench

real    0m0.058s
user    0m0.043s
sys     0m0.013s
ws$ time node bench

real    0m0.052s
user    0m0.038s
sys     0m0.012s
ws$ 
```

ronag:
```
╔══════════════╤═════════╤═══════════════╤═══════════╤═════════════════════════╗
║ Slower tests │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ toBuffer     │     100 │ 141.63 op/sec │  ± 1.44 % │                         ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ Fastest test │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ string       │     100 │ 495.38 op/sec │  ± 2.48 % │ + 249.78 %              ║
╚══════════════╧═════════╧═══════════════╧═══════════╧═════════════════════════╝
```

ronag:
Without pre-allocating the string:

```
╔══════════════╤═════════╤══════════════╤═══════════╤═════════════════════════╗
║ Slower tests │ Samples │       Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼──────────────┼───────────┼─────────────────────────╢
║ toBuffer     │      99 │ 49.12 op/sec │  ± 0.47 % │                         ║
╟──────────────┼─────────┼──────────────┼───────────┼─────────────────────────╢
║ Fastest test │ Samples │       Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼──────────────┼───────────┼─────────────────────────╢
║ string       │     100 │ 60.01 op/sec │  ± 1.17 % │ + 22.15 %               ║
╚══════════════╧═════════╧══════════════╧═══════════╧═════════════════════════╝
```

lpinca:
```
╔══════════════╤═════════╤══════════════╤═══════════╤═════════════════════════╗
║ Slower tests │ Samples │       Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼──────────────┼───────────┼─────────────────────────╢
║ toBuffer     │     100 │ 41.60 op/sec │  ± 1.00 % │                         ║
╟──────────────┼─────────┼──────────────┼───────────┼─────────────────────────╢
║ Fastest test │ Samples │       Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼──────────────┼───────────┼─────────────────────────╢
║ string       │     100 │ 42.91 op/sec │  ± 1.66 % │ + 3.15 %                ║
╚══════════════╧═════════╧══════════════╧═══════════╧═════════════════════════╝
```

with `Buffer.from()` instead of `toBuffer()`, that is removing the `Buffer`, `ArrayBuffer`, `DataView` checks from it.

> Hence the benchmark should only show the difference between doing the conversion in "user space" or inside the native writeGeneric

The problem with this is that the difference is only the result of the huge amount of buffered data in the socket writable queue. The difference should not be noticeable if data is regularly flowing, because the number of allocations would basically be 1:1.

ronag:
> with Buffer.from() instead of toBuffer(), that is removing the Buffer, ArrayBuffer, DataView checks from it.

Sorry. I don't understand?

> The problem with this is that the difference is only the result of the huge amount of buffered data in the socket writable queue. The difference should not be noticeable if data is regularly flowing, because the number of allocations would basically be 1:1.

This is with 4MiB of back pressure, which is not unrealistic.

ronag:
I don't quite see why you think the benchmark is not relevant. According to you, how would a "correct" benchmark look?

lpinca:
> Sorry. I don't understand? We're using toBuffer just as it would with ws. We are not using `Buffer.from`

Yes but we could optimize it for the string case instead of dealing with the string only at the end.

ronag:
Ok updated benchmark:

- Backpressure down to 64kiB
- Added toBuffer2 which check for string earlier
```
╔══════════════╤═════════╤════════════════╤═══════════╤═════════════════════════╗
║ Slower tests │ Samples │         Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼────────────────┼───────────┼─────────────────────────╢
║ toBuffer     │    1000 │ 1662.28 op/sec │  ± 2.12 % │                         ║
║ toBuffer2    │    1000 │ 1695.58 op/sec │  ± 1.17 % │ + 2.00 %                ║
╟──────────────┼─────────┼────────────────┼───────────┼─────────────────────────╢
║ Fastest test │ Samples │         Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼────────────────┼───────────┼─────────────────────────╢
║ string       │    1000 │ 2467.67 op/sec │  ± 1.18 % │ + 48.45 %               ║
╚══════════════╧═════════╧════════════════╧═══════════╧═════════════════════════╝
```

lpinca:
It's 15% on my machine but only 3% with `Buffer.from()`, that 12% only for the `toBuffer()` wrapper is odd.

lib/sender.js
ronag: > What about websocket.ping() and websocket.pong()?

We could do those as well. Didn't see them as a hot path.

lpinca:
How to make cronometro wait for the socket to be connected?

freddy123098:
Well I think by let me check and I will let you know.

freddy123098:
This is how. 

1. var data = bar();

function bar() {
  var result;

  socket.emit('message', 'test');
  socket.on('msg', function(message) {
    result = message;
  });

  return result;
}
Here, the result value is getting undefined. How to assign the value retrieved from the socket.on function to result variable and return?

This is my other end that receives the socket.io message event.

and then this 

2. var io = require('socket.io').listen(server);

io.sockets.on('connection', function(socket, username) {
  socket.on('message', function(message) {
    var result = 'test value'
    socket.emit('msg', result);
  });
});

And then 

3. rying to return a object in acknowlegment.

io.sockets.on('connection', function (socket) {
    socket.on('message', function(message, ackCallback) {
        console.log("server received message", message);
        var result = tools.execute(message);  
        console.log(typeof(result)); // result is returned a object
        ackCallback(result);    
    });  
});

And then 

4. var result = {firstName:"John", lastName:"Doe", age:50, eyeColor:"blue"};
ackCallback(result);


This is it.

ronag:
cronometro does a warmup run so waiting for the socket doesn't really make any difference.

lpinca:
I think cronometro is flawed. On my machine it always win by a large margin the variant added first. I ran a similar test with Benchmark.js

```js
const net = require('net');
const Benchmark = require('benchmark');

const suite = new Benchmark.Suite();

const server = net.createServer(function (socket) {
  socket.resume();
});

server.listen(function () {
  const socket = net.connect(server.address().port, '0.0.0.0');

  socket.on('connect', function () {
    suite.add('string', {
      defer: true,
      fn: function (deferred) {
        for (let n = 0; n < 16; ++n) {
          socket.write(makeStr());
        }
        socket.write(makeStr(), function () {
          deferred.resolve();
        });
      }
    });

    suite.add('toBuffer', {
      defer: true,
      fn: function (deferred) {
        for (let n = 0; n < 16; ++n) {
          socket.write(toBuffer(makeStr()));
        }
        socket.write(toBuffer(makeStr()), function () {
          deferred.resolve();
        });
      }
    });

    suite.add('bufferFrom', {
      defer: true,
      fn: function (deferred) {
        for (let n = 0; n < 16; ++n) {
          socket.write(Buffer.from(makeStr()));
        }
        socket.write(Buffer.from(makeStr()), function () {
          deferred.resolve();
        });
      }
    });

    suite.on('cycle', function (event) {
      console.log(event.target.toString());
    });

    suite.on('complete', function () {
      console.log('Fastest is ' + this.filter('fastest').map('name'));
    });

    suite.run({ async: true });
  });
});

function toBuffer(data) {
  toBuffer.readOnly = true;

  if (Buffer.isBuffer(data)) return data;

  let buf;

  if (data instanceof ArrayBuffer) {
    buf = Buffer.from(data);
  } else if (ArrayBuffer.isView(data)) {
    buf = Buffer.from(data.buffer, data.byteOffset, data.byteLength);
  } else {
    buf = Buffer.from(data);
    toBuffer.readOnly = false;
  }

  return buf;
}

function makeStr() {
  return Buffer.allocUnsafe(4096).toString();
}
```

```
string x 2,682 ops/sec ±2.03% (66 runs sampled)
toBuffer x 2,501 ops/sec ±1.49% (78 runs sampled)
bufferFrom x 2,309 ops/sec ±1.34% (80 runs sampled)
Fastest is string
```

lpinca:
Probably because I'm using macOS

```
$ npx envinfo --system

  System:
    OS: macOS 12.1
    CPU: (16) x64 Intel(R) Xeon(R) W-2140B CPU @ 3.20GHz
    Memory: 20.48 GB / 32.00 GB
    Shell: 5.1.12 - /usr/local/bin/bash
```

ronag:
Added system info to my results.

lpinca:
Please run this on you systems:

```js
// server.js, run it on separate process
const http = require('http');

  const key = crypto
    .createHash('sha1')
    .digest('base64');

  socket.resume();

  socket.write(
    [
      'Upgrade: websocket',
      'Connection: Upgrade',
      '\r\n'
    ].join('\r\n')
  );
});

server.listen(8080);
```

```js
// bench.js
const Benchmark = require('benchmark');

const WebSocket1 = require('../../code/ws');
const WebSocket2 = require('../ws');

function makeStr() {
  return Buffer.allocUnsafe(4096).toString();
}

const wscurrent = new WebSocket1('ws://localhost:8080', {
  generateMask() {}
});
const wspathed = new WebSocket2('ws://localhost:8080', {
  generateMask() {}
});

const suite = new Benchmark.Suite();

suite.add('current', {
  defer: true,
  fn: function (deferred) {
    for (let n = 0; n < 16; ++n) {
      wscurrent.send(makeStr());
    }
    wscurrent.send(makeStr(), function () {
      deferred.resolve();
    });
  }
});

suite.add('patched', {
  defer: true,
  fn: function (deferred) {
    for (let n = 0; n < 16; ++n) {
      wspathed.send(makeStr());
    }
    wspathed.send(makeStr(), function () {
      deferred.resolve();
    });
  }
});

suite.on('cycle', function (event) {
  console.log(event.target.toString());
});

suite.on('complete', function () {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
});

let connected = 0;

wscurrent.on('open', run);
wspathed.on('open', run);

function run() {
  if (++connected === 2) {
    suite.run({ async: true });
  }
}
```

ronag:
```
current x 3,494 ops/sec ±1.31% (82 runs sampled)
patched x 4,920 ops/sec ±0.83% (85 runs sampled)
Fastest is patched
```

I ran a profile and most time is spent in `makrStr` the above results is with the following change:

```
const str = Buffer.allocUnsafe(4096).toString();
function makeStr() {
  return str;
}
```

lpinca:
What are the result without the changes to `makeStr()`?

ronag:
Same as your's.

ronag:
```
const strings = [];
for (let n = 0; n < 10000; ++n) {
  strings.push(Buffer.allocUnsafe(4096).toString());
}

function makeStr() {
  return strings.pop();
}
```

```
current x 13,942 ops/sec ±6.34% (73 runs sampled)
patched x 14,793 ops/sec ±2.11% (75 runs sampled)
```

lpinca:
Ok, always sending the same string, does not make sense and invalidates the test.

ronag:
> Ok, always sending the same string does not make sense and invalidates the test.

Added another one. See above.

lpinca:
That's a ~5% improvement.

ronag:
+ reduced gc pressure

ronag:
btw, I kind of did manage to resolve that with the current version by setting `Buffer.poolSize = 128 * 1024`.

lpinca:
Yes, it's a small improvement that has some readability and maintenance drawbacks. We can move this forward but I want to make is general and handle `ping()`, `pong()`, and permessage-deflate. It shouldn't be hard, we only have to precompute the message byte length and pass it along to `sender.enqueue()`, `sender.dequeue()`, and `Sender.frame()` with a dedicate option.

lpinca:
Anyway, thank you for bearing with me.

ronag:
Something like this?

ronag:
Tried cleaning it up a little more with proper commit msgs. A little exhausted on this one ATM. Hope this can be included in some form. Feel free to take over.

freddy123098:
Why did u re-open this issue?

PR #2000, opened 2022-01-06
Reduce buffer allocations

lpinca:
Do not convert strings to `Buffer`s if data does not need to be masked.

Refs: https://github.com/websockets/ws/pull/1998

### DIFF
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -46,8 +46,11 @@ class Sender {
   /**
    * Frames a piece of data according to the HyBi WebSocket protocol.
    *
-   * @param {Buffer} data The data to frame
+   * @param {(Buffer|String)} data The data to frame. The string type is only
+   *     allowed for internal usage.
    * @param {Object} options Options object
+   * @param {Number} [options.byteLength] The length in bytes of the data. Only
+   *     relevant if data is a string. This is only for internal usage.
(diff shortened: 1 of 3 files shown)


lpinca:
@ronag, this is what I had in mind. Anyway

1. The API is convoluted
1. It is harder to read and maintain
1. `Buffer.concat(Sender.frame(data, { opcode: 1 }))` no longer works if `data` is a string.

so I'm not ok with merging if the gain is marginal.

ronag (approved):

lib/sender.js
```diff
@@ -80,12 +83,27 @@ class Sender {
...
       offset = 6;
     }
 
-    let payloadLength = data.length;
+    let dataLength;
+
+    if (typeof data === 'string') {
+      if (!options.mask || skipMasking) {
+        dataLength =
+          options.byteLength !== undefined
```
lpinca, line 94: I wonder if it is better to hide this behind a symbol property (`options[kByteLength]`) so that the API cannot be misused. I'm don't know if symbol properties are still slower.

lpinca:
Current implementation addresses my two main concerns:

1. The API is now simpler.
1. The `data` argument of the `Sender.frame()` function can now be a string but the function always returns a list of `Buffer`s when used externally. Behavior changes when an internal only option(`options[kByteLength]`) is used.

@ronag  what do you think?

PR #2004, opened 2022-01-14
fix: use dataLength

ronag:


ronag:
Minor bug when passing strings, we shouldn't look at the string char length but it's byte length

lib/sender.js
```diff
@@ -135,6 +135,10 @@ class Sender {
 
     if (skipMasking) return [target, data];
 
+    if (typeof data === 'string') {
+      data = Buffer.from(data);
+    }
```
ronag, line 140: This isn't strictly need but makes things safer for bugs introduced by future changes.

lib/sender.js
```diff
@@ -103,10 +103,10 @@ class Sender {
 
     let payloadLength = dataLength;
 
-    if (data.length >= 65536) {
+    if (dataLength >= 65536) {
```
lpinca, line 106: Good catch. Thanks.

lib/sender.js
lpinca: Not very happy with this. I would prefer to use `data.length` below.

lpinca (approved):

lpinca:
Thank you. I will merge this and cut a new release later.

PR #2007, opened 2022-01-31
[WebSocketServer] Allow to extend the original `WebSocket` class

vansergen:
@lpinca What do you think about adding this feature?

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -85,7 +85,7 @@ This class represents a WebSocket server. It extends the `EventEmitter`.
     connections. See description below. (Usage is discouraged: see
     [Issue #337](https://github.com/websockets/ws/issues/377#issuecomment-462152231))
   - `WebSocket` {Function} Specifies the `WebSocket` class to be used. It must
-    be extended from the original `WebSocket`. _Default_: `WebSocket`.
+    be extended from the original `WebSocket`. Defaults to `WebSocket`.
 - `callback` {Function}
 
 Create a new server instance. One and only one of `port`, `server` or `noServer`
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -84,6 +84,8 @@ This class represents a WebSocket server. It extends the `EventEmitter`.
   - `verifyClient` {Function} A function which can be used to validate incoming
     connections. See description below. (Usage is discouraged: see
     [Issue #337](https://github.com/websockets/ws/issues/377#issuecomment-462152231))
+  - `WebSocket` {Function} Specifies the `WebSocket` class to be used. It must
+    be extended from the original `WebSocket`. _Default_: `WebSocket`.
 - `callback` {Function}
 
 Create a new server instance. One and only one of `port`, `server` or `noServer`
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -49,8 +49,8 @@ class WebSocketServer extends EventEmitter {
    * @param {Boolean} [options.skipUTF8Validation=false] Specifies whether or
    *     not to skip UTF-8 validation for text and close messages
    * @param {Function} [options.verifyClient] A hook to reject connections
-   * @param {Function} [options.WebSocket] Specifies the `WebSocket` class to be
-   *     used. Useful for extending the original `WebSocket`
+   * @param {Function} [options.WebSocket=WebSocket] Specifies the `WebSocket`
+   *     class to use. It must be the `WebSocket` class or class that extends it
    * @param {Function} [callback] A listener for the `listening` event
    */
   constructor(options, callback) {
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -49,6 +49,8 @@ class WebSocketServer extends EventEmitter {
    * @param {Boolean} [options.skipUTF8Validation=false] Specifies whether or
    *     not to skip UTF-8 validation for text and close messages
    * @param {Function} [options.verifyClient] A hook to reject connections
+   * @param {Function} [options.WebSocket] Specifies the `WebSocket` class to be
+   *     used. Useful for extending the original `WebSocket`
    * @param {Function} [callback] A listener for the `listening` event
    */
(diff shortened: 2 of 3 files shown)


lpinca:
I'm not very happy with it. Are there big advantages over monkey patching the `WebSocket` class directly?

```js

Object.defineProperty(WebSocket.prototype, 'custom_field', {
  get() {
    return '123';
  }
});
```

vansergen:
> Are there big advantages over monkey patching the WebSocket class directly?

I do believe so. It would give more flexibility in terms of accessing the
original methods while overwriting them

```javascript
class CustomWebSocket extends WebSocket {
  #sent;
  #created_at;
  constructor(...params) {
    super(...params);
    this.#sent = 0;
    this.#created_at = new Date();
  }
  get sent() {
    return this.#sent;
  }
  get created_at() {
    return new Date(this.#created_at);
  }
  send(...params) {
    this.#sent++;
    super.send(...params);
  }
}
```

in the same manner as the `node:http` module does

```javascript
import { IncomingMessage, ServerResponse, createServer } from 'node:http';
class CustomIncomingMessage extends IncomingMessage {}
class CustomServerResponse extends ServerResponse {}
createServer({
  IncomingMessage: CustomIncomingMessage,
  ServerResponse: CustomServerResponse
}).on('request', (request, response) => {
  console.log(request instanceof CustomIncomingMessage);
  console.log(response instanceof CustomServerResponse);
});
```

In this way one can modify `WebSocket` classes per server (without modifying the original class) 

```javascript
import { parse } from 'url';
import { WebSocketServer, WebSocket } from 'ws';

class WebSocket1 extends WebSocket {}
class WebSocket2 extends WebSocket {}

const wss1 = new WebSocketServer({ WebSocket: WebSocket1, noServer: true });
const wss2 = new WebSocketServer({ WebSocket: WebSocket2, noServer: true });

wss1.on('connection', function connection(ws) {
  // console.log(ws instanceof WebSocket1)
});
wss2.on('connection', function connection(ws) {
  // console.log(ws instanceof WebSocket2)
});

  const { pathname } = parse(request.url);
  if (pathname === '/foo') {
    wss1.handleUpgrade(request, socket, head, function done(ws) {
      wss1.emit('connection', ws, request);
    });
    wss2.handleUpgrade(request, socket, head, function done(ws) {
      wss2.emit('connection', ws, request);
    });
  } else {
    socket.destroy();
  }
});
server.listen(8080);
```

lpinca:
Hmm, ok. It might break if the class does not extend `WebSocket` and I do not like to keep adding options but it is reasonable.

test/websocket-server.test.js
```diff
@@ -510,6 +510,41 @@ describe('WebSocketServer', () => {
...
+
+        const ws = new WebSocket(`ws://localhost:${server.address().port}`);
+
+        ws.on('open', () => {
+          ws.on('close', () => {
+            server.close(done);
+          });
+        });
+      });
+    });
```
lpinca, line 547:
```suggestion
    it('honors the `WebSocket` option', (done) => {
      class CustomWebSocket extends WebSocket.WebSocket {
        get foo() {
          return 'foo';
        }
      }

      const wss = new WebSocket.Server(
        {
          port: 0,
          WebSocket: CustomWebSocket
        },
        () => {
          const ws = new WebSocket(`ws://localhost:${wss.address().port}`);

          ws.on('open', ws.close);
        }
      );

      wss.on('connection', (ws) => {
        assert.ok(ws instanceof CustomWebSocket);
        assert.strictEqual(ws.foo, 'foo');
        wss.close(done);
      });
    });
```

Also, please move it under the "options" section.

doc/ws.md
```diff
@@ -84,6 +84,8 @@ This class represents a WebSocket server. It extends the `EventEmitter`.
   - `verifyClient` {Function} A function which can be used to validate incoming
     connections. See description below. (Usage is discouraged: see
     [Issue #337](https://github.com/websockets/ws/issues/377#issuecomment-462152231))
+  - `WebSocket` {Function} Specifies the `WebSocket` class to be used. It must
+    be extended from the original `WebSocket`. _Default_: `WebSocket`.
```
lpinca, line 88:
```suggestion
    be extended from the original `WebSocket`. Defaults to `WebSocket`.
```

For consistency

test/websocket-server.test.js
```diff
@@ -510,6 +510,32 @@ describe('WebSocketServer', () => {
         });
       });
     });
+
+    it('honors the `WebSocket` option', (done) => {
```
lpinca, line 514: Please move the test under the "options" section (line 92).

vansergen: https://github.com/websockets/ws/pull/2007/commits/50226c8a74cf99ebfae56006ccb4a09e0e4ff5f7

lpinca (approved):

lib/websocket-server.js
```diff
@@ -356,7 +359,7 @@ class WebSocketServer extends EventEmitter {
       `Sec-WebSocket-Accept: ${digest}`
     ];
 
-    const ws = new WebSocket(null);
+    const ws = new this.options.WebSocket(null);
```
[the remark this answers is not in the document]
vansergen: I got the following results (using `hrtime.bigint()`):

```javascript
const ws = new WebSocket(null); // Average `9161n` nanoseconds
const ws = new this.options.WebSocket(null); // Average `13524n` nanoseconds
```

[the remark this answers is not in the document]
vansergen: It seems that introducing a new variable
```js 
this._WebSocket = options.WebSocket
``` 
and accessing it with
```javascript
const ws = new this._WebSocket(null);
```
would not give such an impact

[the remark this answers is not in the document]
lpinca: I think this is acceptable, we can eventually add the new property in future.

lpinca:
Thank you.

PR #2013, opened 2022-02-05
Drop sensitive headers when following redirects

lpinca:
Do not forward the `Authorization` and `Cookie` headers if the redirect
host is different from the original host.

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -766,6 +766,45 @@ function initAsClient(websocket, address, protocols, options) {
     opts.path = parts[1];
   }
 
+  if (opts.followRedirects) {
+    if (websocket._redirects === 0) {
+      websocket._originalHost = parsedUrl.host;
+
+      const headers = options && options.headers;
+
+      //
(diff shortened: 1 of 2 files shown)


lpinca:
This is the topic of some recent CVEs like https://nvd.nist.gov/vuln/detail/CVE-2022-0235, and https://nvd.nist.gov/vuln/detail/CVE-2022-0155, but I will not create a security advisory  for `ws` because

1. Redirects are not followed by default.
1. The issue was known when the feature was added (https://github.com/websockets/ws/pull/1490).
1. I do not think this fixes the issue because there might be other headers containing sensitive data, like `X-AUTH-KEY`, `X-API-KEY`, etc.  It might actually break legitimate use cases.
    - Refs: https://bugs.python.org/issue33661
    - Refs: https://github.com/node-fetch/node-fetch/pull/1449#issuecomment-1018669981
    - Refs: https://github.com/node-fetch/node-fetch/pull/1449#issuecomment-1018828475

PR #2016, opened 2022-02-10
Default empty sec-websocket-protocol header to undefined

juanmjacobs:
Hi! How are you doing? 
First of all, thank you for all of your hard work on this great library!
I'm definitely not an expert on web sockets, but I've noticed that some servers respond with an empty `sec-websocket-protocol` header instead of not returning one at all. 
For those servers, the connection ends up failing with `Server sent a subprotocol but none was requested` because the validation expects an `undefined` header as opposed to an empty `""` header. 
This change solves that issue, allowing the connection to succeed.

Please let me know if you have any comments or if I can be of any help!

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -877,7 +877,7 @@ function initAsClient(websocket, address, protocols, options) {
       return;
     }
 
-    const serverProt = res.headers['sec-websocket-protocol'];
+    const serverProt = res.headers['sec-websocket-protocol'] || undefined;
     let protError;
 
     if (serverProt !== undefined) {


juanmjacobs:
Thank you for the info, I totally see the point!
> **and** this header field indicates the use of a subprotocol that was not present in the client's handshake
Just asking from ignorance, does an empty header value actually counts as an "indicated subprotocol"?

juanmjacobs:
>It is not a valid subprotocol, but it technically indicates a subprotocol that was not requested by the client.

Ok, cool, thanks! I've also tried specifying an `""` as a requested subprotocol but that also fails, as an empty string is not a valid subprotocol. So the library makes it impossible to connect to servers with that peculiarity. I think that from a pragmatic standpoint, it could be useful to have this merged even if it kinda stretches the RFC interpretation a little. 

But I also see totally your point! Feel free to close this PR if you think that's for the best. I'll try and make a workaround for my particular issue.

lpinca:
@juanmjacobs `ws@7` should be more lenient and accept an empty `Sec-WebSocket-Protocol` header filed value in the response.

juanmjacobs:
> @juanmjacobs IIRC `ws@7` is more lenient and accepts an empty `Sec-WebSocket-Protocol` header filed value in the response.

I've tried it and that's exactly right! Thanks a lot @lpinca

PR #2025, opened 2022-03-18
Added WebSocketServer.createClient()

apancutt:
While it's possible to customise the `WebSocket` class used by `WebSocketServer` (using the `WebSocket` option), it's not possible to pass custom constructor arguments to it.

This PR isolates the creation of the `WebSocket` instance in an extensible way allowing `WebSocketServer` subclasses to pass custom constructor arguments - based on the `request`, for example.

The same arguments provided to `WebSocketServer.completeUpgrade()` are passed to `WebSocketServer.createClient()` for broader use-cases.

Rationale...

I want to attach metadata to each `WebSocket` instance based on the incoming request URL. Every `WebSocket.open` event should be logged, and the log message should include this metadata.

Problem is, the `WebSocket.open` event is fired before the `WebSocketServer.handleUpgrade()` callback is executed so by then it's too late to attach the metadata. Providing the metadata at instantiation solves it.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -12,6 +12,7 @@
   - [server.address()](#serveraddress)
   - [server.clients](#serverclients)
   - [server.close([callback])](#serverclosecallback)
+  - [server.createClient(extensions, key, protocols, request, socket, head)](#servercreateclientextensions-key-protocols-request-socket-head)
   - [server.handleUpgrade(request, socket, head, callback)](#serverhandleupgraderequest-socket-head-callback)
   - [server.shouldHandle(request)](#servershouldhandlerequest)
 - [Class: WebSocket](#class-websocket)
@@ -225,6 +226,21 @@ connections are closed unless an external HTTP server is used and client
 tracking is disabled. In this case the `'close'` event is emitted in the next
 tick. The optional callback is called when the `'close'` event occurs and
 receives an `Error` if the server is already closed.
+
+### server.createClient(extensions, key, protocols, request, socket, head)
+
+- `extensions` {Object} The accepted extensions.
+- `key` {String} The value of the `Sec-WebSocket-Key` header.
+- `protocols` {Set} The subprotocols.
+- `request` {http.IncomingMessage} The request object.
+- `socket` {(net.Socket|tls.Socket)} The network socket between the server and
+  client.
+- `head` {Buffer} The first packet of the upgraded stream
+
+Creates a new `WebSocket` client instance.
+
+This method can be overridden to provide custom constructor arguments to your
+custom `WebSocket` class (see the [`WebSocket` option](#new-websocketserveroptions-callback)).
 
 ### server.handleUpgrade(request, socket, head, callback)
 
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -239,8 +239,9 @@ receives an `Error` if the server is already closed.
 
 Creates a new `WebSocket` client instance.
 
-This method can be overridden to provide custom constructor arguments to your
-custom `WebSocket` class (see the [`WebSocket` option](#new-websocketserveroptions-callback)).
+An alternative to [the `WebSocket` option](#new-websocketserveroptions-callback),
+this method can be overridden when you need to construct custom `WebSocket`
+instances that require extended constructor arguments.
 
 ### server.handleUpgrade(request, socket, head, callback)
 
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -359,7 +359,7 @@ class WebSocketServer extends EventEmitter {
       `Sec-WebSocket-Accept: ${digest}`
     ];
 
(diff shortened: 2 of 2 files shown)


lpinca:
On the server your `'open'` event is the `'connection'` event.

```js
websocketServer.on('connection', function (ws, req) {
  ws.metadata = req.url;
});
```

lpinca:
Sorry but I'm not ok with this, metadata from the `IncomingMessage` can already be added to a `WebSocket`. The `WebSocket` is created and returned synchronously to the user. There is no need to use a custom constructor.

apancutt:
Forgot to mention, I'm using `noServer` mode, so that event is never fired.

Even in server mode the `WebSocket.open` event is fired after the `WebsocketServer.connection` event. This would mean you'd need to add your opening listener to the server and all other listeners (e.g. `message`, etc.) to the client.

lpinca:
There is no difference, in `noServer` mode `WebSocket` is passed to the `webocketServer.handleUpgrade()` callback. Your `'open'` event is when that callback is called. That callback also takes the `IncomingMessage` as argument.

apancutt:
Not according to my testing...

```js
class MyWebSocket extends WebSocket {

  constructor(...args) {
    super(...args);
    this.on('open', () => {
      console.log('open');
    });
  }

}
```

```js
wss.handleUpgrade(request, socket, head, () => {
  console.log('callback');
});
```

`open` is logged before `callback`

You're right in that I could remove the open listener to the callback instead, but that means the open-_ish_ handler is attached in a different place to all the other event listeners which isn't ideal.

Anyway, logging metadata is only half the problem...

Typically, one would add `message` listeners in the `handleUpgrade()` callback (right?). Problem is, there are a few milliseconds between the handshake completing at the socket level and the `handleUpgrade()` callback firing. During that short window the client is able to send messages that we're not listening to. This is a real-world issue we're facing.

If we attach a `message` listener at WebSocket construction time then the window is closed and there is no way for messages to slip through. This is achievable with the current version of `ws` using custom WebSocket classes, but without the metadata we don't know where they came from.

apancutt:
Maybe an example would help...

```js
class MyWebSocket extends WebSocket {

  constructor(address, protocols, options, metadata) {
    super(address, protocols, options);
    this.metadata = metadata;
    this.on('message', () => {
      console.log('never missed', this.metadata);
    });
  }

}
```

```js
wss.createClient = (extensions, key, protocols, req, socket, head) => {
  const metadata = { url: req.url };
  return new MyWebSocket(null, undefined, undefined, metadata);
};

wss.handleUpgrade(request, socket, head, (client) => {
  client.on('message', () => {
    console.log('missed', client.metadata);
  });
});
```

If the client immediately sends a message after the handshake completes but before the `handleUpgrade()` callback is fired then you will see `never missed` but will not see `missed` in the logs.

In our case it's crucial we don't miss messages, and it's equally crucial that we have access to the metadata.

lpinca:
> You're right in that I could move the open listener to the callback instead, but that means the open-ish handler is attached in a different place to all the other event listeners which isn't ideal.

You have to add the callback anyway.

> Typically, one would add `message` listeners in the `handleUpgrade()` callback (right?). Problem is, there are a few milliseconds between the handshake completing at the socket level and the `handleUpgrade()` callback firing. During that short window the client is able to send messages that we're not listening to. This is a real-world issue we're facing.

> If we attach a `message` listener at WebSocket construction time then the window is closed and there is no way for messages to slip through. This is achievable with the current version of `ws` using custom WebSocket classes, but without the metadata we don't know where they came from.

Wrong, the callback is called is called synchronously right after the `WebSocket` is created. You can't miss messages. This is by design.

As an example, consider the following example, where a client that does not follow the spec (that specifies that no data should be sent before the handshake completes) sends data in the handshake request, before the handshake completes (before the `'open'` event is emitted on the client).

```js
const http = require('http');

  wss.handleUpgrade(request, socket, head, open);
});

function open(ws, request) {
  ws.metadata = request.url;
  ws.on('message', function (message) {
    console.log(ws.metadata);
    console.log(message.toString());
  });
}

server.listen(function () {
  const req = http.request({
    path: '/foo',
    port: server.address().port,
    headers: {
      Connection: 'Upgrade',
      Upgrade: 'websocket',
      'Sec-WebSocket-Key': 'dGhlIHNhbXBsZSBub25jZQ==',
      'Sec-WebSocket-Version': 13
    }
  });

  const list = WebSocket.Sender.frame(Buffer.from('Hello'), {
    fin: true,
    rsv1: false,
    opcode: 0x01,
    mask: true,
    readOnly: false
  });

  req.write(Buffer.concat(list));
  req.end();
});
```

apancutt:
OK, thanks a lot for taking the time to review this. Using your example I can see that messages are never lost and that the problem we're experiencing is likely caused by something else.

PR #2030, opened 2022-04-07
Introduce the `'redirect'` event

lpinca:
Add the ability to remove confidential headers on a per-redirect basis.

Closes #2014


lpinca:
cc: @vansergen

PR #2034, opened 2022-04-12
Make cjs example compatible with esm

nopeless:
@lpinca

I think this is a better example


lpinca:
Please do it also in the other examples ("server-stats" and "express-session-parse").

nopeless:
I went to sleep before I could see
Thanks for caring 👍

PR #2036, opened 2022-04-21
Examples update: Fix HTTP method

protofarer:
Logout endpoint does not meet spec for HTTP DELETE per https://httpwg.org/specs/rfc7231.html#DELETE, use POST instead.


lpinca:
A lot of APIs use the `DELETE` method to remove a session and it makes sense to me. I don't see the need to change this.

PR #2037, opened 2022-04-21
Fix example: http post instead of delete

protofarer:
Logout endpoint does not meet spec for HTTP DELETE per https://httpwg.org/specs/rfc7231.html#DELETE, use POST instead.


lpinca:
See https://github.com/websockets/ws/pull/2036#issuecomment-1104722312.

PR #2038, opened 2022-04-21
Bugfix: missing else wsconnect

protofarer:
The connection logic needs to be moved into the else block to stop the client from immediately reconnecting after pressing the button to disconnect.


lpinca:
This is by design. When the button is clicked, if there is an already open WebSocket connection, the existing one should be closed and a new one opened.

lpinca:

PR #2046, opened 2022-05-20
Introduce the `'wsClientError'` event

### LINKED ISSUE #2045: Emit abort handshake event

### Description

_No response_

### ws version

8.6.0

### Node.js Version

16

### System

_No response_

### Expected result

During the handleUpgrade it would be nice an event to catch the handshake fail

### Actual result

I'm not able to catch upgrade errors, like in the attached screenshot where the client is sending an invalid _sec-websocket-key_

### Attachments

image

lpinca:
Add the ability to inspect the invalid handshake requests and respond
to them with a custom HTTP response.

Closes #2045

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -9,6 +9,7 @@
   - [Event: 'error'](#event-error)
   - [Event: 'headers'](#event-headers)
   - [Event: 'listening'](#event-listening)
+  - [Event: 'wsClientError'](event-wsclienterror)
   - [server.address()](#serveraddress)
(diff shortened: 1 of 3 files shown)


lpinca:
cc: @emanuele-dedonatis

lpinca:
The event is named after the [`'clientError'`](https://nodejs.org/api/http.html#event-clienterror) and [`tlsClientError`](https://nodejs.org/api/tls.html#event-tlsclienterror) events.

PR #2056, opened 2022-06-08
Add the `WS_NO_{BUFFER_UTIL, UTF_8_VALIDATE}` variables

lpinca:
When set to non empty values, the `WS_NO_BUFFER_UTIL` and
`WS_NO_UTF_8_VALIDATE` environment variables, prevent the optional
`bufferutil` and `utf-8-validate` dependencies  from being required,
respectively.

These might be useful to enhance security in systems where a user can
put a package in the package search path of an application of another
user, due to how the Node.js resolver algorithm works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,13 @@ necessarily need to have a C++ compiler installed on your machine.
 - `npm install --save-optional utf-8-validate`: Allows to efficiently check if a
   message contains valid UTF-8.
 
+To not even try to require and use these modules, use the
+[`WS_NO_BUFFER_UTIL`](./doc/ws.md#ws_no_buffer_util) and
+[`WS_NO_UTF_8_VALIDATE`](./doc/ws.md#ws_no_utf_8_validate) environment
+variables. These might be useful to enhance security in systems where a user can
+put a package in the package search path of an application of another user, due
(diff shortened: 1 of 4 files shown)


lpinca:
cc: @arhart

arhart:
@lpinca The code looks good, and I've confirmed that setting `WS_NO_BUFFER_UTIL` and/or `WS_NO_UTF_8_VALIDATE` to non-empty strings prevent the calls to `require()`, as expected.

I liked this sentence from the commit message:
>     These might be useful to enhance security in systems where a user can
>     put a package in the package search path of an application of another
>     user, due to how the Node.js resolver algorithm work.

Could you add something like that to the README.md as well? Perhaps like this:
```
To not even try to require and use these modules, use the
[`WS_NO_BUFFER_UTIL`](./doc/ws.md#ws_no_buffer_util) and
[`WS_NO_UTF_8_VALIDATE`](./doc/ws.md#ws_no_utf_8_validate) environment
variables. These might be useful to enhance security in systems where a user can
put a package in the package search path of an application of another
user, due to how the Node.js resolver algorithm work.
```

lpinca:
Done.

arhart:
Thanks, LGTM

PR #2067, opened 2022-07-15
Made "Simple Server" more complete and at the top

SephReed:
Details added from change:
- the type of url that web sockets are hosted at (eg. ws:/).  This is not indicated elsewhere
- a simple example of how one would connect using native functions that come in browser or node.js

By adding these changes, the example now includes everything someone would need to setup a basic websocket connection

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,32 @@ const ws = new WebSocket('ws://www.host.com/path', {
 
 ## Usage examples
 
+### Simple server
+
+```js
+import { WebSocketServer } from 'ws';
+
+const wss = new WebSocketServer({ port: 8080 });
+
+wss.on('connection', function connection(ws) {
+  ws.on('message', function message(data) {
+    console.log('received: %s', data);
+  });
+
+  ws.send('something');
+});
+
+console.log("Now hosting at ws://localhost:8080")
+```
+
+```html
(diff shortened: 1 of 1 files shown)


lpinca:
Thank you for the PR but I'm not ok with the changes. The example is now no longer consistent with the other examples and can no longer be be copied and pasted and run.

> the type of url that web sockets are hosted at (eg. ws:/). This is not indicated elsewhere

This is shown in the other examples.

> a simple example of how one would connect using native functions that come in browser or node.js

This is also shown in the other examples.

lpinca:

SephReed:
This is very disappointing.  Surely you recognize that one can never view their own documentation with fresh eyes?  I know you're smart enough to see the implications of that.

lpinca:
Yes, changes for the sake of change are not accepted. That isolated `<script>` element does not make any sense to me when there are dedicated examples like https://github.com/websockets/ws#sending-and-receiving-text-data and https://github.com/websockets/ws#sending-binary-data.

For the browser use the [browser documentation](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket). Support for the `EventTarget` interface is documented here https://github.com/websockets/ws/blob/8.10.0/doc/ws.md#websocketaddeventlistenertype-listener-options.

PR #2071, opened 2022-08-03
Add "Server" alias to wrapper.mjs to match index.js exports and types

floodfx:
`index.js` exports the contents of `.lib/websocket-server` under the aliases of both `WebSocketServer` and `Server`.  Unfortunately, `wrapper.mjs` only exports `WebSocketServer` not `Server`.  Only exporting `WebSocketServer` breaks the types (i.e. `@types/ws`) and causes build errors (which is how I caught this).  

This commit adds the `Server` alias to `wrapper.mjs`.

### DIFF
--- a/wrapper.mjs
+++ b/wrapper.mjs
@@ -4,5 +4,7 @@ import Sender from './lib/sender.js';
 import WebSocket from './lib/websocket.js';
 import WebSocketServer from './lib/websocket-server.js';
 
-export { createWebSocketStream, Receiver, Sender, WebSocket, WebSocketServer };
+const Server = WebSocketServer;
+
+export { createWebSocketStream, Server, Receiver, Sender, WebSocket, WebSocketServer };
 export default WebSocket;


lpinca:
I would prefer not do it. Users can/should use `WebSocketServer` which works in both CJS and ESM. I can't remove `WebSocket.Server` because it is a breaking change. It is planned for the next major release.

floodfx:
Seems easy to add and consistent with the types and the index.js. If you are going to remove next major release that makes sense but to why not add until then?  Some users will run into errors in the meantime if you don’t.

lpinca:
> If you are going to remove next major release that makes sense but to why not add until then? Some users will run into errors in the meantime if you don’t.

It will only cause unnecessary breakage for people that will start using it. Use `WebSocketServer` instead. Not adding `Server` to the ESM wrapper was an intentional choice.

lpinca:
> breakage because `Server` is not exported there but published as a type

That is not an issue in the library but in the Typescript type definitions that are not maintained by us.

lpinca:

PR #2075, opened 2022-09-14
Add named pipe support for Windows

jacksteamdev:
This PR applies the suggestions from #1808 to add support to the client for named pipes on Windows.

### Cross Platform Example

This example works on both MacOS and Windows.

```javascript
// paths.js

import * as os from 'os'
import { join } from 'path'

let sockPath = join(os.tmpdir(), `ws.example.sock`)
if (os.platform() === 'win32') {
  sockPath = join('\\\\.\\pipe', sockPath)
  console.log('updated sockPath', sockPath)
}

export const sockUrl = `ws+unix:///${sockPath}`
export { sockPath }
```

```javascript
// server.js

import { existsSync, unlinkSync } from 'fs'
import { createServer } from 'http'
import * as os from 'os'
import { WebSocketServer } from 'ws'
import { sockPath } from './paths.js'

const server = createServer()
const wss = new WebSocketServer({ server })

  console.log('server connected')

  ws.on('message', (data) => {
    console.log('data from client', data.toString())
  })

  ws.send('hello')
})

// server.listen() throws if unix domain socket exists
if (os.platform() !== 'win32' && existsSync(sockPath)) {
  unlinkSync(sockPath)
}

server.listen(sockPath)
```

```js
// client.js

import { WebSocket } from 'ws'
import { sockUrl } from './paths.js'

const ws = new WebSocket(sockUrl)
ws.on('open', () => {
  console.log('client socket open')
  ws.send('hello')
})
```

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -2,68 +2,71 @@
 
 ## Table of Contents
 
-- [Class: WebSocketServer](#class-websocketserver)
-  - [new WebSocketServer(options[, callback])](#new-websocketserveroptions-callback)
-  - [Event: 'close'](#event-close)
-  - [Event: 'connection'](#event-connection)
-  - [Event: 'error'](#event-error)
-  - [Event: 'headers'](#event-headers)
-  - [Event: 'listening'](#event-listening)
-  - [Event: 'wsClientError'](#event-wsclienterror)
-  - [server.address()](#serveraddress)
-  - [server.clients](#serverclients)
-  - [server.close([callback])](#serverclosecallback)
-  - [server.handleUpgrade(request, socket, head, callback)](#serverhandleupgraderequest-socket-head-callback)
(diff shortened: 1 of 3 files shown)


lpinca:
The problem with Windows named pipes is that they create ambiguity when used in URLs. The pipe name string specified by `PipeName` in `\\.pipe\PipeName` can include any character other than a backslash.

For example the pipe name string in `\\.\pipe\foo:/bar?bar=baz` is `foo:/bar?bar=baz` but when added to a URL the `?bar=baz` part is parsed as the URL search component. Similarly the `:` character creates ambiguity because it cannot be used as the actual request path separator. The same applies to `:/`.

The feature can only be added with limitations so I think it does not woth the effort.

jacksteamdev:
Closing in favor of #2079

Issue #1944, opened 2021-09-10
Allow passing in a custom `createConnection` function


connor4312:
### 

- [X] I've searched for any related issues and avoided creating a duplicate issue.

### Description

We have a kind of exotic use case where we want to instantiate a websocket off an existing stream that's not from an HTTP or unix socket source. Currently the `createConnection` method is hardcoded to either create a TLS or standard net connection:

https://github.com/websockets/ws/blob/f871195c55cff39ac51020ea344ec6a2c4ea9f4e/lib/websocket.js#L654

I suggest allowing a custom `createConnection` function to be passed in the websocket options and, if present, it should take precedence over the existing default logic.

I'm happy to put together a PR for this if it's something you'd accept 🙂 

### ws version

8.2.2

### Node.js Version

v16.6.1

### System

  System:
    OS: Windows 10 10.0.19043
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 23.22 GB / 63.89 GB

### Expected result

N/A

### Actual result

N/A

lpinca:
You can use the `agent` option and provide a custom [`Agent`](https://nodejs.org/api/http.html#http_new_agent_options) to do that.

connor4312:
Ah, cool, I didn't realize that that was also an option on the agent. I'll try that out and close this issue if it works.

connor4312:
It turns out that it's not enough to just return the stream, it also needs to be wrapped to look at least a bit like a network socket. Here's the conversion function, with TypeScript typings, in case anyone else stumbles upon the need for this:

```ts
import { Socket } from 'net';
import { Duplex } from 'stream';

export const makeNetSocketFromDuplexStream = (stream: Duplex) => {
	const cast = stream as Socket;
	const patched: { [K in keyof Omit<Socket, keyof Duplex>]: Socket[K] } = {
		bufferSize: 0,
		bytesRead: 0,
		bytesWritten: 0,
		connecting: false,
		localAddress: '127.0.0.1',
		localPort: 1,
		remoteAddress: '127.0.0.1',
		remoteFamily: 'tcp',
		remotePort: 1,
		address: () => ({ address: '127.0.0.1', family: 'tcp', port: 1 }),
		unref: () => cast,
		ref: () => cast,
		connect: (_port: unknown, _host?: unknown, connectionListener?: () => void) => {
			if (connectionListener) {
				setImmediate(connectionListener);
			}
			return cast;
		},
[...]

PR #2088, opened 2022-10-13
Allow passing in a custom createConnection function

tbn-mm:
https://github.com/websockets/ws/issues/1944


lpinca:
See https://github.com/websockets/ws/issues/1944#issuecomment-916650992. There is no need for it.

tbn-mm:
You are right. Sorry for this pull request.

PR #2112, opened 2023-01-23
docs: add error handlers to examples, and improve error event description

dyc3:
- docs: add minimal error handlers to examples
- docs: indicate that failing to handle an error event will crash the process


lpinca:
There were unrelated changes, some missing handlers, and some nits to fix, so I went ahead and pushed https://github.com/websockets/ws/commit/b82682a8d89f7a35b236868e9f0554e953c7fcb3. I did not change the `'error'` event description because crashing on unhandled `'error'` events is standard Node.js behavior.

Thank you anyway.

Issue #1818, opened 2020-11-21
Use Node.js 15 native `EventTarget` object


piranna:
- [x] I've searched for any related issues and avoided creating a duplicate
      issue.

#### Description

Node.js 15 already provides a native implementation of `EventTarget`, so there's no need to use our own implementation. In fact, using both of them at the same time leads to errors.

In my use case, I've created a `Client` class that extends from Node.js native `EventTarget` class, that internally it's using an `ws` instance (and doing some other project specific things), and setting as listeners methods from this `Client` class, with the idea of propagate these errors to the user:

```js
export class Client extends EventTarget
{
  constructor(ws)
  {
    super()

    if(!(ws instanceof WebSocket)) ws = new WebSocket(ws)

    ws.binaryType = 'arraybuffer'

    ws.addEventListener('close', this.#onClose, {once: true})
    ws.addEventListener('error', this.#onError)
    ws.addEventListener('message', this.#onMessage)
    ws.addEventListener('open', this.#onOpen, {once: true})

    this.#ws = ws
  }

  #onClose = this.dispatchEvent.bind(this)
  #onError = this.dispatchEvent.bind(this)
  #onOpen  = this.dispatchEvent.bind(this)
}
```

Problem is, since `ws` is using its own implementation of both `EventTarget` and `Event` classes, when these events gets propagated to the native one, I get the next error:

```js
TypeError [ERR_INVALID_ARG_TYPE]: The "event" argument must be an instance of Event. Received an instance of ErrorEvent
    at new NodeError (node:internal/errors:277:15)
    at EventTarget.dispatchEvent (node:internal/event_target:326:13)
    at WebSocket.onError (/home/piranna/Trabajo/Atos/awrtc_signaling/node_modules/ws/lib/event-target.js:141:16)
    at WebSocket.emit (node:events:329:20)
    at WebSocket.EventEmitter.emit (node:domain:467:12)
    at ClientRequest.<anonymous> (/home/piranna/Trabajo/Atos/awrtc_signaling/node_modules/ws/lib/websocket.js:579:15)
    at ClientRequest.emit (node:events:329:20)
    at ClientRequest.EventEmitter.emit (node:domain:467:12)
    at TLSSocket.socketErrorListener (node:_http_client:478:9)
    at TLSSocket.emit (node:events:329:20) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

[...]

PR #2116, opened 2023-02-01
Export CloseEvent, ErrorEvent, MessageEvent classes

edolix:
I don't know if you guys like the idea to make these classes public but i ended up implementing the CloseEvent class in one of my project (since i had to trigger it manually).

I thought it would be useful for others so that's the reason for the PR. If you think otherwise, feel free to close it! 

Thanks for the great work on this package 💯

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,11 +1,16 @@
 'use strict';
 
 const WebSocket = require('./lib/websocket');
+const { CloseEvent, ErrorEvent, MessageEvent } = require('./lib/event-target');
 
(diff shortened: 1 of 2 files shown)


lpinca:
Thank you Edoardo but I think it is not a good idea to expose them because

1. We are not strictly following the `EventTarget` specification (#1818).
2. Those events should not be dispatched manually.

edolix:
Sounds good @lpinca! 
Thanks for the quick review - i'll close the PR!

PR #2118, opened 2023-02-13
Allow ws to provide a fallback when bundling for the browser

flashd2n:
Added "browser" property to "exports", which mimics the default behaviour already present by: "browser": "browser.js"

Signed-off-by: Kalin Kostov <k.kostov.eu@gmail.com>


package.json
```diff
@@ -18,6 +18,9 @@
   "main": "index.js",
   "exports": {
     ".": {
+      "browser": {
+        "default": "./browser.js"
+      },
```
lpinca, line 23:
```suggestion
      "browser": "./browser.js",
```

package.json
```diff
@@ -61,4 +64,4 @@
     "prettier": "^2.0.5",
     "utf-8-validate": "^6.0.0"
   }
-}
+}
```
lpinca, line 67:
```suggestion
}

```

lpinca (approved):

PR #2122, opened 2023-03-02
[WebSocketServer] Allow to use a custom `IncomingMessage` class

vansergen:
## Motivation

```javascript
import { IncomingMessage } from "node:http";
import { WebSocketServer } from "ws";
class MyIncomingMessage extends IncomingMessage {}
const wss = new WebSocketServer({ port: 8080, IncomingMessage: MyIncomingMessage }).on(
  "connection",
  (_ws, request) => {
    console.log(request instanceof MyIncomingMessage); // true
  }
);
```

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -90,6 +90,9 @@ This class represents a WebSocket server. It extends the `EventEmitter`.
   - `verifyClient` {Function} A function which can be used to validate incoming
     connections. See description below. (Usage is discouraged: see
     [Issue #337](https://github.com/websockets/ws/issues/377#issuecomment-462152231))
+  - `IncomingMessage` {Function} Specifies the `IncomingMessage` class to be
+    used. Useful for extending the original `IncomingMessage`. _Defaults_ to
+    `http.IncomingMessage`
   - `WebSocket` {Function} Specifies the `WebSocket` class to be used. It must
     be extended from the original `WebSocket`. Defaults to `WebSocket`.
 - `callback` {Function}
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -49,6 +49,8 @@ class WebSocketServer extends EventEmitter {
    * @param {Boolean} [options.skipUTF8Validation=false] Specifies whether or
    *     not to skip UTF-8 validation for text and close messages
    * @param {Function} [options.verifyClient] A hook to reject connections
+   * @param {Function} [options.IncomingMessage=IncomingMessage] Specifies the `IncomingMessage`
+   *     class to be used. Useful for extending the original `IncomingMessage`.
    * @param {Function} [options.WebSocket=WebSocket] Specifies the `WebSocket`
    *     class to use. It must be the `WebSocket` class or class that extends it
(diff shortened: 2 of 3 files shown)


lpinca:
This is already supported via the `server` option.

```js
const http = require('http');
const { WebSocketServer } = require('ws');

class MyIncomingMessage extends http.IncomingMessage {}
const server = http.createServer({ IncomingMessage: MyIncomingMessage });

const wss = new WebSocketServer({ server });

wss.on('connection', function (ws, request) {
  console.log(request instanceof MyIncomingMessage); // true
});

server.listen(8080);
```

lpinca:
The `server` and `noServer` options are preferable as they offer more control. The `port` option is just a quick way to start a server without customization and boilerplate code.

vansergen:
> This is already supported via the `server` option.

Actually, that is the reason why I created this PR - to simplify this (when the `server` is required).

> The `server` and `noServer` options are preferable as they offer more control. The `port` option is just a quick way to start a server without customization and boilerplate code.

Totally agree. One could have more flexibility with `server`/`noServer` options, but sometimes 

> without customization and boilerplate code

is exactly what is needed (416 responses by default). @lpinca I know you do not like adding extra options, but this one seems natural to me. We can use an extended `WebSocket` class with the `port` option. Why can't we extend the `IncomingMessage`?

vansergen:
> A custom `WebSocket` class is a `ws` thing. You can't pass it as an option to `http.createServer()`.

Agree, but we use things `backlog` in the `options` to customize an internally created server. Why could we customize it a bit more with `IncomingMessage`?

vansergen:
I mean events like `connection` and `upgrade` are a part of the Public API, right? The `request` variable is a part of it, so it could be nice to have a native option to extend it.

lpinca:
This opens the doors to every other option listed here https://nodejs.org/api/http.html#httpcreateserveroptions-requestlistener and here https://nodejs.org/api/https.html#httpscreateserveroptions-requestlistener. I don't want that when the `server` option exists.

What are the benefits of supporting this option? A total of ~10 lines of boilerplate code removed? Sorry but the cons outweigh the pros.

vansergen:
> This opens the doors to every other option listed here https://nodejs.org/api/http.html#httpcreateserveroptions-requestlistener

I disagree. The server itself(when the `port` options is used) is not exposed as a part of the Public API. So, there is no need to support those options at all, but the `IncomingMessage` is (`connection` event etc.). That's why I suggest to include only this option.

lpinca:
> The server itself(when the `port` options is used) is not exposed as a part of the Public API. So, there is no need to support those options at all, but the `IncomingMessage` is (`connection` event etc.). That's why I suggest to include only this option.

It does not matter. If the `IncomingMessage` option is supported why `ServerResponse`, `insecureHTTPParser`, etc. are not. They are all options of `http.createServer()` and the next time someone like you want to support one of those options I can't say no because there is a precedent. The user argument will be "But the `IncomingMessage` option is supported!".

Again, the `server` option already allows to do this easily so it does not make sense to me.

lpinca:

PR #2123, opened 2023-03-03
[feature] Add option to support late addition of headers

mvduin:
This adds an option for WebSocket clients that makes it possible to add request headers late minute, specifically to support the use-case where a header can't be determined until the socket is connected (TLS channel binding), which means the `req.end()` call needs to be deferred.

I wasn't sure what the most elegant way to support this was so I've tried to keep this simple and minimally invasive, especially since it's for a rather obscure use-case. It just adds a callback hook which, if provided, will be invoked instead of `req.end()`. It's then up to the provided function call `req.end()` when it's done customizing the request.

If this interface looks okay I'll amend this pull request to document the option and add a test.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -293,6 +293,8 @@ This class represents a WebSocket. It extends the `EventEmitter`.
 - `address` {String|url.URL} The URL to which to connect.
 - `protocols` {String|Array} The list of subprotocols.
 - `options` {Object}
+  - `finishRequest` {Function} A function which can be used to customize the
+    headers of each http request before it is sent. See description below.
   - `followRedirects` {Boolean} Whether or not to follow redirects. Defaults to
     `false`.
   - `generateMask` {Function} The function used to generate the masking key. It
@@ -316,12 +318,24 @@ This class represents a WebSocket. It extends the `EventEmitter`.
     Options given do not have any effect if parsed from the URL given with the
     `address` parameter.
 
+Create a new WebSocket instance.
+
 `perMessageDeflate` default value is `true`. When using an object, parameters
 are the same of the server. The only difference is the direction of requests.
 For example, `serverNoContextTakeover` can be used to ask the server to disable
 context takeover.
 
-Create a new WebSocket instance.
+`finishRequest` is called with arguments
+
+- `request` {http.ClientRequest}
+- `websocket` {WebSocket}
+
+for each HTTP GET request (the initial one and any caused by redirects) when it
+is ready to be sent, to allow for last minute customization of the headers. If
+`finishRequest` is set then it has the responsibility to call `request.end()`
+once it is done setting request headers. This is intended for niche use-cases
+where some headers can't be provided in advance e.g. because they depend on the
+underlying socket.
 
 #### IPC connections
 
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -989,7 +989,11 @@ function initAsClient(websocket, address, protocols, options) {
     });
   });
 
-  req.end();
+  if (opts.finishRequest) {
+    opts.finishRequest(req, websocket);
+  } else {
+    req.end();
+  }
 }
 
 /**
--- a/test/websocket.test.js
+++ b/test/websocket.test.js
@@ -3857,6 +3857,35 @@ describe('WebSocket', () => {
         agent
       });
     });
+
+    it('honors the `finishRequest` option', (done) => {
+      const wss = new WebSocket.Server({ port: 0 }, () => {
+        const ws = new WebSocket(`ws://localhost:${wss.address().port}`, {
+          finishRequest(request, websocket) {
+            process.nextTick(() => {
+              assert.strictEqual(request, ws._req);
+              assert.strictEqual(websocket, ws);
+            });
+            request.on('socket', (socket) => {
+              socket.on('connect', () => {
+                request.setHeader('Cookie', 'foo=bar');
+                request.end();
+              });
+            });
+          }
+        });
+
+        ws.on('close', (code) => {
+          assert.strictEqual(code, 1005);
+          wss.close(done);
+        });
+      });
+
+      wss.on('connection', (ws, req) => {
+        assert.strictEqual(req.headers.cookie, 'foo=bar');
+        ws.close();
+      });
+    });
   });
 
   describe('permessage-deflate', () => {


lpinca:
> specifically to support the use-case where a header can't be determined until the socket is connected (TLS channel binding), which means the `req.end()` call needs to be deferred.

Can you show an example using the Node.js `https` module?

mvduin:
Here's a minimalistic toy demo of using the tls-exporter channel binding to securely authenticate a websocket client using a pre-shared key. The authentication is bound to the TLS connection which means that even if an attacker successfully impersonates the server to the client, the credential they receive from the client (the Authorization header) cannot be used by the attacker to authenticate to the real server.

This example of course assumes my patch has been applied to `ws`.

```js
import { WebSocket, WebSocketServer } from 'ws';
import https from 'https';
import fs from 'fs';
import crypto from 'crypto';
import events from 'events';

// CAUTION: this implementation requires TLSv1.3.  older TLS versions can use tls-exporter
// channel binding but that requires additional checks to be implemented to be secure.
function get_tls_exporter_channel_binding( socket ) {
	return Buffer.concat([
		Buffer.from('tls-exporter:'),
		socket.exportKeyingMaterial( 32, 'EXPORTER-Channel-Binding', Buffer.alloc(0) )
	]);
}

// note: this is just a toy example and not a real-world authentication protocol
function _tls_silly_psk_auth_data( socket, psk ) {
	let hmac = crypto.createHmac( 'sha256', psk );
	hmac.update( get_tls_exporter_channel_binding( socket ) );
	return hmac.digest( 'base64url' );
}
async function tls_silly_psk_auth_apply( req, psk ) {
	let socket = req.socket || ( await events.once( req, 'socket' ) )[0];
	if( socket.secureConnecting )
		await events.once( socket, 'secureConnect' );
	let auth = 'X-Silly-PSK-Auth ' + _tls_silly_psk_auth_data( socket, psk );
	req.setHeader( 'Authorization', auth );
}
function tls_silly_psk_auth_verify( req, psk ) {
	let auth = req.headers.authorization;
	if( ! auth )
		return false;
	// FIXME this should probably parse the header properly in case alternate
	// authorizations are provided alongside the one we care about
	[ , auth ] = /^\s*X-Silly-PSK-Auth\s+([\w-]+)\s*$/i.exec( auth ) || [];
	return auth === _tls_silly_psk_auth_data( req.socket, psk );
}


const shared_secret = Buffer.from( '81f98e0e75dbaefa6e62363b293d9629', 'hex' );

const wss = new WebSocketServer({ noServer: true, path: '/' });

const https_server = https.createServer({
	cert: fs.readFileSync('test/fixtures/certificate.pem'),
	key: fs.readFileSync('test/fixtures/key.pem'),
	minVersion: 'TLSv1.3',  // see caution above get_tls_exporter_channel_binding()
});
https_server.on( 'upgrade', ( req, socket, head ) => {
	if( ! tls_silly_psk_auth_verify( req, shared_secret ) ) {
		socket.write(
			'HTTP/1.1 401 Unauthorized\r\n' +
			'WWW-Authenticate: X-Silly-PSK-Auth\r\n' +
			'Content-Length: 0\r\n' +
			'Connection: close\r\n' +
			'\r\n' );
		socket.destroy();
		return;
	}
	wss.handleUpgrade( req, socket, head, ws => {
		console.log( `client connected` );
		ws.on( 'close', ( code, reason ) => {
			reason = reason.toString();
			console.log( `client disconnected`, { code, reason } );
		});
	});
});
https_server.listen( 8123 );
https_server.unref();
await events.once( https_server, 'listening' );
console.log( 'server listening on', https_server.address() );


let ws = new WebSocket( 'wss://localhost:8123/', {
	async finishRequest( req ) {
		await tls_silly_psk_auth_apply( req, shared_secret );
		req.end();
	},
	minVersion: 'TLSv1.3',  // see caution above get_tls_exporter_channel_binding()
	rejectUnauthorized: false,  // ignore bogus server cert
});
await events.once( ws, 'open' );
console.log( 'test successful!' );
ws.close();
```

lpinca:
I meant an example without `ws` to see how you would normally do it for HTTP (no WebSocket) but the above example also shows that.

It is really an obscure use case and I don't like to expose the `http.ClientRequest` object to the user but I have not better idea. Currently the `http.ClientRequest` is only exposed to the listeners of the `'redirect'` event.

> If this interface looks okay I'll amend this pull request to document the option and add a test.

Go ahead. Thank you.

doc/ws.md
```diff
@@ -316,12 +318,23 @@ This class represents a WebSocket. It extends the `EventEmitter`.
...
+`finishRequest` is called with arguments
+
+- `request` {http.ClientRequest}
+- `websocket` {WebSocket}
+
+for each HTTP GET request (the initial one and any caused by redirects) when it
+is ready to be sent. If `finishRequest` is set then it has the responsibility to
+call `request.end()` when it is done customizing the request. This is intended
+for niche use-cases that need to defer finishing the request until its socket is
+available, e.g. to add an authentication header involving TLS channel binding.
```
lpinca, line 337: Please clarify that customization should be limited to inspecting and adding headers. For example, the user should not add data to the request body or call `request.destroy()`.

lpinca (commented):
Please add a test.

mvduin:
I was wondering, would this perhaps be a nicer approach:
```js
  if (typeof opts.setLateHeaders === 'function') {
    (async () => {
      await opts.setLateHeaders(req, websocket);
      if (req === null || req[kAborted]) return;
      req.end();
    })();
  } else {
    req.end();
  }
```
This makes `options.setLateHeaders` an async function that only sets the headers, the responsibility to `end()` the request remains with ws. It also makes the name of the option specifically focussed on setting headers.

mvduin:
Okay, test has been added and documentation updated based on your feedback.

doc/ws.md
```diff
@@ -316,12 +318,24 @@ This class represents a WebSocket. It extends the `EventEmitter`.
...
 `perMessageDeflate` default value is `true`. When using an object, parameters
 are the same of the server. The only difference is the direction of requests.
 For example, `serverNoContextTakeover` can be used to ask the server to disable
 context takeover.
 
-Create a new WebSocket instance.
+`finishRequest` is called with arguments
+
+- `request` {http.ClientRequest}
+- `websocket` {WebSocket}
```
lpinca, line 331: Is this argument needed? The user already has access to the `WebSocket` instance returned by the constructor, no?

mvduin: `finishRequest` is called from the constructor, so the new WebSocket hasn't been returned yet

lpinca: Fair point, but is it needed? What would it be used for?

mvduin: I have no specific use-case, it just seemed polite to provide it to finishRequest just in case it's needed since it gets invoked before the caller has access to the websocket. It's not critical, but it takes no effort to pass it as argument while it would be annoying if it doesn't get passed and you do end up needing it.

lpinca: I think we can add it later if someone needs it.

lpinca: On second thought, it might be useful if someone want to call `websocket.close()` or `websocket.terminate()` synchronously.

mvduin: Or they may just want to keep track of the WebSocket for which they're doing asynchronous work for bookkeeping/debugging purposes. Passing it just seemed like a sensible thing to do.

mvduin: In the end I don't care enough to continue debating it, I'll remove the websocket argument.

lpinca: No, keep it. I quoted my comment, not yours.

lib/websocket.js
```diff
@@ -989,7 +989,11 @@ function initAsClient(websocket, address, protocols, options) {
     });
   });
 
-  req.end();
+  if (typeof opts.finishRequest === 'function') {
```
lpinca, line 992:
```suggestion
  if (typeof opts.finishRequest) {
```

test/websocket.test.js
```diff
@@ -3857,6 +3857,31 @@ describe('WebSocket', () => {
         agent
       });
     });
+
+    it('supports late headers using finishRequest', (done) => {
```
lpinca, line 3861:
```suggestion
    it('honors the `finishRequest` option', (done) => {
```

test/websocket.test.js
```diff
@@ -3857,6 +3857,31 @@ describe('WebSocket', () => {
...
+      const wss = new WebSocket.Server({ port: 0 }, () => {
+        const ws = new WebSocket(`ws://localhost:${wss.address().port}`, {
+          finishRequest(req, ws_arg) {
+            setTimeout(() => {
+              assert.strictEqual(ws_arg, ws);
+              assert.strictEqual(req, ws._req);
+              req.setHeader('Cookie', 'foo=bar');
+              req.end();
+            }, 150);
+          }
```
lpinca, line 3871:
```suggestion
          finishRequest(request, websocket) {
            setImmediate(() => {
              assert.strictEqual(request, ws._req);
              assert.strictEqual(websocket, ws);
              req.setHeader('Cookie', 'foo=bar');
              req.end();
            });
          }
```

mvduin:
I've fixed the lint issues

lpinca (approved):

test/websocket.test.js
mvduin: I used setTimeout to verify that it works "later" when everything that can progress without `req.end()` has had a chance to do so, setImmediate will not be sufficient  for that.

lpinca: It doesn't matter we can also do synchronously. We only have to test that a header added here is received. Headers are buffered and not sent (unless `request.flushHeaders()` is called) until `request.end()` is called.

lib/websocket.js
mvduin: You obviously meant just `if (opts.finishRequest)` without the `typeof`, but yeah. I wrote the test like this because that's what node's http code usually seemed to do (e.g. for `options.createConnection`), but I personally also don't really see a good reason to not just check for truthy.

lpinca: Yes, the suggestion is mainly for consistency with the codebase.

test/websocket.test.js
mvduin: I know that delaying it a bit *shouldn't* matter, but I'd prefer the test to confirm that this is genuinely the case. I could alternatively make the test reflect actual expected use:
```js
    it('honors the `finishRequest` option', (done) => {
      const server = https.createServer({
        cert: fs.readFileSync('test/fixtures/certificate.pem'),
        key: fs.readFileSync('test/fixtures/key.pem')
      });

        assert.strictEqual(req.headers.cookie, 'foo=bar');
        server.close(done);
      });

      server.listen(0, () => {
        const ws = new WebSocket(`wss://localhost:${wss.address().port}`, {
          rejectUnauthorized: false,
          finishRequest(req, ws) {
            assert.strictEqual(req, ws._req);
            req.on('socket', socket => {
              socket.on('secureConnect', () => {
                req.setHeader('Cookie', 'foo=bar');
                req.end();
              });
            });
          }
        });

        ws.on('open', () => {
          ws.close();
        });
      });
    });
```

Also, all other tests abbreviate request to `req` and websocket to `ws`, it would be inconsistent to spell them out here.

lpinca: It would certainly be better than waiting arbitrarily for 150 ms. I don't think the test needs to use TLS, it can use the `'connect'` event of the socket.

> Also, all other tests abbreviate request to `req` and websocket to `ws`, it would be inconsistent to spell them out here.

My suggestion was because I don't like `ws_arg` (snake case) or `wsArg`. `ws` can't be used, so for consistency with the verbose `websocket` I also used the verbose `request`.

mvduin: How about
```js
    it('honors the `finishRequest` option', (done) => {
      const wss = new WebSocket.Server({ port: 0 }, () => {
          finishRequest(req, ws) {
            assert.strictEqual(req, ws._req);
            req.on('socket', socket => {
              socket.on('connect', () => {
                req.setHeader('Cookie', 'foo=bar');
                req.end();
              });
            });
          }
        });

        ws.on('open', () => {
          ws.close();
        });
      });

        assert.strictEqual(req.headers.cookie, 'foo=bar');
        wss.close(done);
      });
    });
```
this doesn't explicitly test that the inner `ws` equals the outer one but if the `ws` argument isn't the websocket then the `assert.strictEqual(req, ws._req);` test will catch that anyway.

lpinca: I'm not very happy with it because that is an implementation detail. Anyway my bad, I should have fixed the nits myself after merging.

mvduin:
I've updated it based on your feedback.

lpinca:
Thank you.

test/websocket.test.js
lpinca: See https://github.com/websockets/ws/commit/23acf8cfaff73fadf89c69be669b3baa29b60233. It is similar to your suggestion with a few more assertions to make sure that the `http.ClientRequest` received by the `finishRequest` function is the same created by the `WebSocket` constructor. In this way with`assert.strictEqual(req, ws._req)` we also make sure that the inner `ws` is the same as the outer `ws` as they share the same request.

There are still no checks for reference equality so it is not perfect but I prefer it to using `process.nextTick()`.

Issue #2048, opened 2022-05-24
It is not possible to pass a custom Sec-WebSocket-Key field in headers


bahamut657:
### Is there an existing issue for this?

- [X] I've searched for any related issues and avoided creating a duplicate issue.

### Description

Passing a custom Sec-WebSocket-Key field is not considered.
It seems to calculate each time a new key even if provided.

### ws version

8.5.0

### Node.js Version

16.15.0

### System

System:
    OS: Linux 5.17 Arch Linux
    CPU: (4) x64 Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
    Memory: 5.69 GB / 11.61 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 16.15.0 - /usr/bin/node
    npm: 8.5.5 - /usr/bin/npm
  Managers:
    pip3: 21.0 - /usr/bin/pip3
    RubyGems: 3.3.8 - /usr/bin/gem
  Utilities:
    CMake: 3.23.1 - /usr/bin/cmake
    Make: 4.3 - /usr/bin/make
    GCC: 11.2.0 - /usr/bin/gcc
    Git: 2.36.1 - /usr/bin/git
    Clang: 13.0.1 - /usr/bin/clang
    Ninja: 1.10.2 - /usr/bin/ninja
    Subversion: 1.14.1 - /usr/bin/svn
    FFmpeg: 5.0 - /usr/bin/ffmpeg
  Servers:
    Apache: 2.4.53 - /usr/bin/apachectl
  Virtualization:
    Docker: 20.10.15 - /usr/bin/docker
    VirtualBox: 6.1.34 - /usr/bin/vboxmanage
  SDKs:
    Android SDK:
      API Levels: 23, 30, 31
      Build Tools: 29.0.1, 29.0.2, 32.0.0
      System Images: android-23 | Intel x86 Atom, android-23 | Google APIs Intel x86 Atom, android-30 | Google APIs Intel x86 Atom
  IDEs:
    Nano: 6.3 - /usr/bin/nano
    Sublime Text: 2 - /usr/bin/subl
    VSCode: 1.63.2 - /usr/bin/code
    Vim: 8.2 - /usr/bin/vim
  Languages:
    Bash: 5.1.16 - /usr/bin/bash
    Go: 1.18.1 - /usr/bin/go
    Perl: 5.34.1 - /usr/bin/perl
    PHP: 8.1.5 - /usr/bin/php
[...]

PR #2125, opened 2023-03-09
Sec-WebSocket-Key is possible to override

gustawdaniel:
I need to use `Sec-Websocket-Key` but can't override random value selected by this module. Thanks to this change I will be able to set my own `Sec-Websocket-Key`, save it and reuse without problems.


lpinca:
See https://github.com/websockets/ws/issues/2048.

gustawdaniel:
Ok. Thanks,

Issue #2124, opened 2023-03-09
create-websocket-stream.test.js fails if the domain module is loaded


mvduin:
### Is there an existing issue for this?

- [X] I've searched for any related issues and avoided creating a duplicate issue.

### Description

If the `domain` module is loaded when`test/create-websocket-stream.test.js` executes, it has some bizarre failures (see below). Other tests do not have this porblem.

I ran into this problem because I have `NODE_OPTIONS="-r /home/matthijs/.node_defaults.js"` in my environment to load a script that tweaks some REPL behaviour, for which I have to require `repl` which then requires `domain`.

While an argument could be made that it's impolite to have an environment that causes `domain` to be loaded into every node script, it's concerning for it to be causing test failures since that may imply actual bugs when `ws` is used in an environment that has `domain` loaded (e.g. whenever you're testing something from the REPL).

### ws version

8.12.1

### Node.js Version

14.21.3, 16.19.1, 18.14.2

### Attachments

Test output:

```
  createWebSocketStream
    ✓ is exposed as a property of the `WebSocket` class
    ✓ returns a `Duplex` stream
    ✓ passes the options object to the `Duplex` constructor
    The returned stream
      ✓ buffers writes if `readyState` is `CONNECTING`
      ✓ errors if a write occurs when `readyState` is `CLOSING`
      ✓ errors if a write occurs when `readyState` is `CLOSED`
      ✓ does not error if `_final()` is called while connecting
      ✓ makes `_final()` a noop if no socket is assigned
      ✓ reemits errors
      ✓ does not swallow errors that may occur while destroying
      1) does not suppress the throwing behavior of 'error' events
      2) is destroyed after 'end' and 'finish' are emitted (1/2)
      3) is destroyed after 'end' and 'finish' are emitted (1/2)
      ✓ is destroyed after 'end' and 'finish' are emitted (2/2)
      ✓ handles backpressure (1/3)
      ✓ handles backpressure (2/3)
      ✓ handles backpressure (3/3)
      ✓ can be destroyed (1/2)
      ✓ can be destroyed (2/2)
      ✓ converts text messages to strings in readable object mode
      ✓ resumes the socket if `readyState` is `CLOSING`


  18 passing (119ms)
  3 failing

  1) createWebSocketStream
       The returned stream
         does not suppress the throwing behavior of 'error' events:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

2 !== 1

      + expected - actual

      -2
      +1
      
[...]

PR #2126, opened 2023-03-09
[fix] make test/create-websocket-stream.test.js compatible with domains

mvduin:
Fix a failure in test/create-websocket-stream.test.js if the node:domain module is loaded (e.g. due to NODE_OPTIONS in environment).

The cause of the failure was that installing an 'uncaughtException' event handler on process will cause node:domain to prepend its own handler for the same event, which confused the test.

Fixes: #2124

### DIFF
--- a/test/create-websocket-stream.test.js
+++ b/test/create-websocket-stream.test.js
@@ -295,11 +295,14 @@ describe('createWebSocketStream', () => {
         ws._socket.write(Buffer.from([0x85, 0x00]));
       });
 
-      assert.strictEqual(process.listenerCount('uncaughtException'), 1);
+      assert.strictEqual(
+        process.listenerCount('uncaughtException'),
+        EventEmitter.usingDomains ? 2 : 1
+      );
 
-      const [listener] = process.listeners('uncaughtException');
+      const listener = process.listeners('uncaughtException').pop();
 
-      process.removeAllListeners('uncaughtException');
+      process.removeListener('uncaughtException', listener);
       process.once('uncaughtException', (err) => {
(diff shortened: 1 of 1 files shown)


test/create-websocket-stream.test.js
```diff
@@ -295,11 +295,14 @@ describe('createWebSocketStream', () => {
...
       });
 
-      assert.strictEqual(process.listenerCount('uncaughtException'), 1);
+      assert.strictEqual(
+        process.listenerCount('uncaughtException'),
+        EventEmitter.usingDomains ? 2 : 1
+      );
 
-      const [listener] = process.listeners('uncaughtException');
+      const listener = process.listeners('uncaughtException').pop();
```
lpinca, line 303: The domain added `'uncaughtException'` listener is always prepended and kept, right?

mvduin: It is prepended when you add the first listener for `'uncaughtException'` and removed when you remove the last listener for it (excluding the one from `domain`)

```
> let foo = () => {}
> let bar = () => {}
> process.listenerCount('uncaughtException')
0
> process.on('uncaughtException', foo); process.listenerCount('uncaughtException')
2
> process.on('uncaughtException', bar); process.listenerCount('uncaughtException')
3
> process.off('uncaughtException', foo); process.listenerCount('uncaughtException')
2
> process.off('uncaughtException', bar); process.listenerCount('uncaughtException')
0
```

lpinca (approved):

lpinca:
Thank you.

PR #2128, opened 2023-03-14
Adding a 'closing' event when the server starts the close handshake

barsvinicius:
This event helps us control the flow of streaming data to the remote websocket server when a close handshake is initiated. The 'close' event is normally only triggered after the connection is already closed, but at this time it is possible that data has been written to the socket, which is no longer available. When the 'closing' event is now emitted, this means that it is time to stop sending any data upstream because a close event will follow soon.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -365,6 +365,11 @@ ws+unix:/absolute/path/to/uds_socket
 
 it defaults to `/`.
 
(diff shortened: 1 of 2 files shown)


barsvinicius:
I am closing this PR because I failed to add the `closeCode` to the event.

PR #2146, opened 2023-05-24
[pkg] Move performance lib to optionalDependencies

nuintun:
Move opt-in for performance lib to optionalDependencies


lpinca:
No, sorry. `optionalDependencies` are installed by default.

PR #2152, opened 2023-07-18
feat: Generate typing information for library

mastrzyz:
I know this is a pure .js library but I'm wondering if we can introduce typescript to generate the typing information only?

Not 100% ready yet PR just want to get opinions before I continue.


lpinca:
Sorry but no, the definitions are externally maintained and I don't want to deal with any issues arising from them.

Issue #1458, opened 2018-10-15
WebSocket over HTTP2


rahbari:
Now that node core has added websocket support over http2 https://github.com/nodejs/node/pull/23284, It's nice to know when it will be added to WS?

lpinca:
See https://github.com/nodejs/node/pull/23284#discussion_r223162692. To answer your question, when someone interested in having the feature, will implement it and open a PR.

ayZagen:
@lpinca We rely on ws and want to upgrade our servers to http2. It would be great to have http2 support on ws.
Thanks

FallingSnow:
I'm using node's http2 framework with the `allowHTTP1: true` option. Then setting a `ws` server to listen on the http2 server just works.

Although I would like to see `ws` support for HTTP2 so it feels less hacky.

josephg:
@FallingSnow does that do websockets over http2 or is that just falling back to websockets over http1.1?

josephg:
@lpinca the big advantage (architecturally) is taking advantage of multiplexing. Right now using websockets sort of requires you to throw away any http-like abstractions (and indeed we often just open a single ws connection to `ws://example.com/ws` or something) and do our own multiplexing.

With ws over h2 it can start making sense again for applications / APIs to add per-URL websocket connections for specific change feeds / use cases. (Since all connections are multiplexed through the single h2 tcp socket anyway).

The other advantage is faster connection times, since if the page is loaded over h2 then we don't need to do another tcp/TLS handshake to open a websocket.

lpinca:
WebSocket over HTTP2 has only one advantage that is as you said the use of a single TCP connection when the origin is the same origin/user/app. In all other cases there are only disadvantages especially on the server and with the Node.js HTTP2 implementation.

Node.js HTTP2 adds a lot of overhead over a plain `{net,tls}.Socket`. There are streams wrapping other streams and the whole HTTP2 machinery. I did not run any benchmark but I'm pretty sure it would perform way worse in terms of both speed and especially memory usage.

See also this discussion https://github.com/aspnet/AspNetCore/issues/7801 and in particular this comment https://github.com/aspnet/AspNetCore/issues/7801#issuecomment-486498006.

FallingSnow:
@josephg I believe the connect comes in as an HTTP1.1 request. There is no actual HTTP2 as far as I'm aware.

borodadada:
[...]

PR #2165, opened 2023-08-30
[fix] ensure handleUpgrade doesn't break if socket is just a stream

pimterry:
If you attempt to handle a websocket over a connection that is a stream but not a `net.Socket`, right now WS fails to do so. With this very small change (very similar to the [equivalent handling](https://github.com/nodejs/node/blob/fcf5de008a59229fe65a7b8e4e45fb4cfef4de81/lib/net.js#L2144-L2147) in Node itself) it's possible to call `handleUpgrade` with a socket argument that's actually just a duplex stream, and then successfully handle websocket traffic that way.

This can be useful in some cases: in my specific case this is required to proxy WebSockets over non-sockets, such as HTTP/2 streams (note that this is not the same as real WS-over-HTTP/2 support - I receive a proxy CONNECT sent on an HTTP/2 stream, and then the client does an HTTP/1.1 upgrade on that proxied stream, resulting in running an HTTP/1.1 websocket over an HTTP/2 stream instead of a raw socket).

With this change, that works flawlessly for me, proxying real websocket traffic from Chrome in an HTTP/2 proxy.

This is also a small step towards https://github.com/websockets/ws/issues/1458, since the same change will also eventually be required to add full HTTP/2 support, where the entire underlying connection is HTTP/2 throughout, with no proxying involved.

(Arguably the same could be done for the `socket.setTimeout` above too, but in my case that doesn't seem to be required, and I wanted to avoid changing anything else unnecessarily).

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -224,7 +224,10 @@ class WebSocket extends EventEmitter {
     receiver.on('pong', receiverOnPong);
 
     socket.setTimeout(0);
-    socket.setNoDelay();
+    if (socket.setNoDelay) {
+      // May not be available if 'socket' is actually a stream
+      socket.setNoDelay();
+    }
 
     if (head.length > 0) socket.unshift(head);
 
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -223,9 +223,11 @@ class WebSocket extends EventEmitter {
     receiver.on('ping', receiverOnPing);
     receiver.on('pong', receiverOnPong);
 
-    socket.setTimeout(0);
+    // These methods may not be available if `socket` is actually just a stream:
+    if (socket.setTimeout) {
+      socket.setTimeout(0);
+    }
     if (socket.setNoDelay) {
-      // May not be available if 'socket' is actually a stream
       socket.setNoDelay();
     }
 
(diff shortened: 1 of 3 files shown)


lib/websocket.js
```diff
@@ -224,7 +224,10 @@ class WebSocket extends EventEmitter {
     receiver.on('pong', receiverOnPong);
 
     socket.setTimeout(0);
-    socket.setNoDelay();
+    if (socket.setNoDelay) {
+      // May not be available if 'socket' is actually a stream
```
lpinca, line 228: Nits:

```suggestion
      //
      // May not be available if `socket` is actually a stream.
      //
```

lpinca (approved):

lpinca:
Why the same is not needed for `socket.setTimeout()`? I also wonder if it is possible to add a test.

pimterry:
For setTimeout, my reasoning is in the description:

> (Arguably the same could be done for the socket.setTimeout above too, but in my case that doesn't seem to be required, and I wanted to avoid changing anything else unnecessarily).

Happy to change it though, it works for me either way!

I've just pushed a commit to cover setTimeout too, apply your nit suggestion, and add a separate test to cover this (which fails without this change, but now passes & works correctly).

lpinca:
Thank you.

PR #2178, opened 2023-11-30
Support for the oppurtunity to set frame headers before connection upgrade on the serverside

WarstekHUN:
# The idea
I wanted to implement a cookie based authentication system to [my WS library](https://zilaws.com/), however I could not find a solution to do so via the `Set-Cookie` header.

# Usage

My implementation requires some extra work to do when upgrading a connection, but lets you set any headers you want.

If I can already do that with the `handleUpgrade`'s header param, then I'm sorry, but I couldn't find out how It would affect frame or what value I should give it to.

```js
import crypto from 'crypto';

const server = createServer((req, res) => {
    const customHeaders = !(req.headers.cookie && req.headers.cookie.includes('myCookie')) ? {
        'Set-Cookie': `myCookie=${crypto.randomUUID()};expires=${new Date(Date.now() + 1000 * 60 * 60 * 24 * 60).toUTCString()};SameSite=None;`
    } : {};
    
    wss.handleUpgrade(req, req.socket, '', (client, req) => {
        wss.emit("connection", client, req);
    }, customHeaders);

}).listen(6678, () => {
    console.log("Listening...");
});

const wss = new WebSocketServer({ noServer: true });

wss.on("connection", (socket, req) => {
    //Logs out the cookies the client already has in their browser
    console.log(cookie.parse(req.headers.cookie ?? ""));
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,39 @@ server.on('upgrade', function upgrade(request, socket, head) {
 server.listen(8080);
 ```
 
+### Modifying upgrade headers if needed
+
+```js
+import { WebSocketServer } from 'ws';
+import { createServer } from 'http';
+import crypto from 'crypto';
+
+const server = createServer((req, res) => {
+  //Checks if there is already a cookie called 'myCookie' in the clients browser. If there's not, sets the 'Set-Cookie' header with a new randomly generated one.
+  const customHeaders = !(
+    req.headers.cookie && req.headers.cookie.includes('myCookie')
+  )
+    ? {
+        'Set-Cookie': `myCookie=${crypto.randomUUID()};expires=${new Date(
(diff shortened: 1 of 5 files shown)


lpinca:
You can use the [`'headers'`](https://github.com/websockets/ws/blob/8.14.2/doc/ws.md#event-headers) event to do this.

WarstekHUN:
@lpinca How exactly? Neither the `headers` or `request` parameter is passed by reference. I've tried setting them but it seemed that it didn't effect the actual response header.

lpinca:
The listener of the `'headers'` event takes 2 parameters. The first is an array with the response headers, the second is the request (`http.IncomingMessage`). You can push the `Cookie` header in the array.

PR #2187, opened 2023-12-24
emit raw tcp socket error to ws client

morpig:
currently, error event for ws client is triggered from pre-upgrade connection (econnrefused, enotfound, etc)

any tcp errors during active ws connection are not forwarded. only 1006 ws close is sent to client.

this PR will emit error event when active tcp socket throws an error.

merry christmas!

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -1317,7 +1317,7 @@ function socketOnEnd() {
  *
  * @private
  */
-function socketOnError() {
+function socketOnError(err) {
   const websocket = this[kWebSocket];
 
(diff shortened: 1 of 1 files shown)


lpinca:
Thank you for the PR but socket errors are not re-emitted by design.

lpinca:
See https://github.com/websockets/ws/commit/a4050db3ad981db1c49daf5ba756ec4f4f71bc44 and https://github.com/websockets/ws/blob/527ec97264cf063bd9c75f33e6a085559fb7d1da/test/websocket.test.js#L629-L652.

PR #2198, opened 2024-02-03
Update ws.md

infomidhun:


lpinca:
Thanks for the PR but the change does not seem correct.

infomidhun:
Hi
Thanks for correcting me too.

Thanks and regards

Midhun Kallayil Madhavan

Issue #2188, opened 2024-01-03
options.autoPong


LiST-GIT:
### Is there an existing issue for this?

- [X] I've searched for any related issues and avoided creating a duplicate issue.

### Description

```js
// websocket.js line 69
if (address !== null) {
      this._bufferedAmount = 0;
      this._isServer = false;
      this._redirects = 0;

      if (protocols === undefined) {
        protocols = [];
      } else if (!Array.isArray(protocols)) {
        if (typeof protocols === 'object' && protocols !== null) {
          options = protocols;
          protocols = [];
        } else {
          protocols = [protocols];
        }
      }

      initAsClient(this, address, protocols, options);
    } else {
      this._autoPong = options.autoPong;
                    // ^ options?.autoPong
      this._isServer = true;
    }
  }
```

```js
new WebSocket(null) // Performing it this way will result in an error.
```

### ws version

8.16.0

### Node.js Version

v20.5.0

lpinca:
That's expected.

```
new WebSocket(null)
```

is not a public API. Use `new WebSocket(null, undefined, {})`.

LiST-GIT:
```ts
new WebSocket(address[, protocols][, options])
```
In the declaration, both protocols and options are optional. However, if I use``` new WebSocket(null, undefined, {}) ```to avoid errors, it could cause compatibility issues. I have encountered this problem. If this is expected behavior, then I will close the issue.

lpinca:
`address` can't be `null`. It's a special value `ws` uses to create server-side instances, but that is not an official API.

LiST-GIT:
Okay, I got it. Thank you for your answer.

PR #2199, opened 2024-02-05
options Undefined error fixed

ackava:
The new version broke http2 proxy example, as `new WebSocket(null)` fails with error `unable to read autoPong from undefined`. This simple check will prevent this error.


lpinca:
See https://github.com/websockets/ws/issues/2188.

PR #2205, opened 2024-03-09
Update websocket.js

t0chk:
Fix TypeError: Cannot read properties of undefined (reading 'autoPong') error

if i make an inheriting class like
```
class AntaresWebsocket extends WebSocket {
  constructor() {
    super(null);
  }
```

and then 

```
    this.wss = new WebSocketServer({
      WebSocket: AntaresWebsocket,
    });
```

take TypeError after client connection

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -84,7 +84,7 @@ class WebSocket extends EventEmitter {
 
       initAsClient(this, address, protocols, options);
     } else {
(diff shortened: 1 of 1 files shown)


lpinca:
See https://github.com/websockets/ws/pull/2199 and https://github.com/websockets/ws/issues/2188.

PR #2218, opened 2024-04-11
Emit at most one event per event loop iteration

### LINKED ISSUE #2216: Messages are dispatched while microtask queue is not empty

### Description

I have previously reported https://github.com/websockets/ws/issues/2159 which was fixed but it looks like the solution might not have been complete. 

With the latest version of ws the microtasks scheduled in the event callback are handled first but the microtasks created by those microtasks are not. This is different from the WebSocket client behavior in the browser.

Server code:

```js

const wss = new WebSocketServer({ port: 8080 });

  ws.send('something1');
  ws.send('something2');
});
```

Client code:

```
import { WebSocket } from 'ws';

const ws = new WebSocket('ws://127.0.0.1:8080');

ws.addEventListener('message', function message(event) {
    console.log('received: %s', event.data);
    new Promise(resolve => {
      resolve()
    }).then(() => {
      console.log('microtask', event.data.toString())
      return Promise.resolve().then(() => {
        console.log('microtask-nested', event.data.toString());
      });
    })
});
```

### ws version

8.16.0

### Node.js Version

v20.10.0

### System

  System:
    OS: macOS 14.4.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 12.88 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh

### Expected result

```
received: something1
microtask something1
microtask-nested something1
received: something2
microtask something2
microtask-nested something2
```

### Actual result

```
received: something1
microtask something1
received: something2
microtask-nested something1
microtask something2
microtask-nested something2
```

### Attachments

_No response_

lpinca:
Fixes #2216

### DIFF
--- a/lib/receiver.js
+++ b/lib/receiver.js
@@ -13,13 +13,6 @@ const { concat, toArrayBuffer, unmask } = require('./buffer-util');
 const { isValidStatusCode, isValidUTF8 } = require('./validation');
 
 const FastBuffer = Buffer[Symbol.species];
-const promise = Promise.resolve();
-
-//
-// `queueMicrotask()` is not available in Node.js < 11.
-//
-const queueTask =
-  typeof queueMicrotask === 'function' ? queueMicrotask : queueMicrotaskShim;
 
 const GET_INFO = 0;
 const GET_PAYLOAD_LENGTH_16 = 1;
@@ -567,17 +560,12 @@ class Receiver extends Writable {
         data = fragments;
       }
 
-      //
-      // If the state is `INFLATING`, it means that the frame data was
-      // decompressed asynchronously, so there is no need to defer the event
-      // as it will be emitted asynchronously anyway.
-      //
-      if (this._state === INFLATING || this._allowSynchronousEvents) {
+      if (this._allowSynchronousEvents) {
         this.emit('message', data, true);
         this._state = GET_INFO;
       } else {
         this._state = DEFER_EVENT;
-        queueTask(() => {
+        setImmediate(() => {
           this.emit('message', data, true);
           this._state = GET_INFO;
           this.startLoop(cb);
@@ -604,7 +592,7 @@ class Receiver extends Writable {
         this._state = GET_INFO;
       } else {
         this._state = DEFER_EVENT;
-        queueTask(() => {
+        setImmediate(() => {
           this.emit('message', buf, false);
           this._state = GET_INFO;
           this.startLoop(cb);
@@ -675,7 +663,7 @@ class Receiver extends Writable {
       this._state = GET_INFO;
     } else {
       this._state = DEFER_EVENT;
-      queueTask(() => {
+      setImmediate(() => {
         this.emit(this._opcode === 0x09 ? 'ping' : 'pong', data);
         this._state = GET_INFO;
         this.startLoop(cb);
@@ -711,32 +699,3 @@ class Receiver extends Writable {
 }
 
 module.exports = Receiver;
-
-/**
- * A shim for `queueMicrotask()`.
- *
- * @param {Function} cb Callback
- */
-function queueMicrotaskShim(cb) {
-  promise.then(cb).catch(throwErrorNextTick);
-}
-
-/**
- * Throws an error.
- *
- * @param {Error} err The error to throw
- * @private
- */
-function throwError(err) {
-  throw err;
-}
-
-/**
- * Throws an error in the next tick.
- *
- * @param {Error} err The error to throw
- * @private
- */
-function throwErrorNextTick(err) {
-  process.nextTick(throwError, err);
-}
--- a/test/receiver.test.js
+++ b/test/receiver.test.js
@@ -1085,17 +1085,21 @@ describe('Receiver', () => {
     receiver.write(Buffer.from([0x88, 0x03, 0x03, 0xe8, 0xf8]));
   });
 
-  it('emits at most one event per microtask', (done) => {
+  it('emits at most one event per event loop iteration', (done) => {
     const actual = [];
     const expected = [
       '1',
-      'microtask 1',
+      '- 1',
+      '-- 1',
       '2',
-      'microtask 2',
+      '- 2',
+      '-- 2',
       '3',
-      'microtask 3',
+      '- 3',
+      '-- 3',
       '4',
-      'microtask 4'
+      '- 4',
+      '-- 4'
     ];
 
     function listener(data) {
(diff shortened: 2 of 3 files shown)


lpinca:
This is more in line with the expectations

```
luigi@imac:bench (master)$ pwd
/Users/luigi/code/ws/bench
luigi@imac:bench (master)$ node parser.benchmark.js 
ping frame (5 bytes payload) x 1,968,413 ops/sec ±0.50% (87 runs sampled)
ping frame (no payload) x 2,595,798 ops/sec ±0.19% (90 runs sampled)
text frame (20 bytes payload) x 1,500,100 ops/sec ±0.66% (85 runs sampled)
binary frame (125 bytes payload) x 1,499,442 ops/sec ±0.54% (82 runs sampled)
binary frame (65535 bytes payload) x 165,904 ops/sec ±0.26% (90 runs sampled)
binary frame (200 KiB payload) x 58,271 ops/sec ±0.29% (90 runs sampled)
binary frame (1 MiB payload) x 11,380 ops/sec ±0.27% (89 runs sampled)
```

```
luigi@imac:bench (master)$ git checkout fix/issue-2216 
Switched to branch 'fix/issue-2216'
Your branch is up to date with 'origin/fix/issue-2216'.
luigi@imac:bench (fix/issue-2216)$ node parser.benchmark.js 
ping frame (5 bytes payload) x 63,608 ops/sec ±0.41% (82 runs sampled)
ping frame (no payload) x 64,907 ops/sec ±0.30% (86 runs sampled)
text frame (20 bytes payload) x 63,806 ops/sec ±0.31% (89 runs sampled)
binary frame (125 bytes payload) x 63,565 ops/sec ±0.30% (87 runs sampled)
binary frame (65535 bytes payload) x 47,378 ops/sec ±0.41% (90 runs sampled)
binary frame (200 KiB payload) x 30,985 ops/sec ±0.56% (87 runs sampled)
binary frame (1 MiB payload) x 9,406 ops/sec ±0.38% (88 runs sampled)
```

The echo benchmark is a special case.

lpinca:
```js
// server.js

const { createHash } = require('crypto');
const { createServer } = require('http');
const { Readable } = require('stream');
const { Sender } = require('ws');

const buf = Buffer.alloc(125);
const frame = Buffer.concat(Sender.frame(buf, { fin: true, opcode: 2 }));

const readable = new Readable({
  read() {
    this.push(frame);
  }
});

  const key = createHash('sha1')
    .digest('base64');

  socket.write(
    [
      'Upgrade: websocket',
      'Connection: Upgrade',
      '\r\n'
    ].join('\r\n')
  );

  readable.pipe(socket);
});

  console.log('Server listening on [::]:8080');
});
```

```js
// client.js

const prettierBytes = require('prettier-bytes');
const speedometer = require('speedometer');

const speed = speedometer();

const ws = new WebSocket('ws://localhost:8080');

ws.on('message', function (data) {
  speed(data.length);
});

setInterval(function () {
  console.log(prettierBytes(speed()) + '/s');
}, 2000);
```

```
# websockets/ws#master
node client.js
56 MB/s
62 MB/s
63 MB/s
63 MB/s
63 MB/s
63 MB/s
63 MB/s
63 MB/s
```

```
# websockets/ws#fix/issue-2216
node client.js
59 MB/s
66 MB/s
67 MB/s
66 MB/s
66 MB/s
67 MB/s
66 MB/s
65 MB/s
```

OrKoN:
On my macbook pro m1 I see different results:

```
# websockets/ws#master
node client.js
29 MB/s
33 MB/s
33 MB/s
31 MB/s
31 MB/s
32 MB/s
```

```
# websockets/ws#fix/issue-2216
node client.js
8.1 MB/s
9.0 MB/s
9.1 MB/s
9.1 MB/s
9.2 MB/s
```

lpinca:
Yes, I also saw similar results on macOS yesterday. I can't make this the new default if the same happens also on Linux.

OrKoN:
I have also tried the following client benchmark:

```
const WebSocket = require('ws'); // omit in the browser

const ws = new WebSocket('ws://localhost:8080');

console.time("ws");

let i = 0;
ws.on('message', function (data) { // addEventListener in browser
  i++;
  if (i === 1000000) {
    console.timeEnd("ws");
    ws.close();
  }
});
```

```
# websockets/ws#master
node client.cjs
ws: 12.662s
```

```
#websockets/ws#fix/issue-2216
node client.cjs
ws: 3.670s
```

```
#chrome-canary
ws: 101s
```

lpinca:
I'm not sure how to proceed. This adds a very big performance penalty but the behavior on master is broken. We can make the whole "one event per tick" optional but that means reverting 93e3552e95ba5ad656c30b94f6be96afe22d4805 and adding a new option for it and that is semver-major.

OrKoN:
We are not able to replicate the regression on a Linux bare metal machine. I think the benchmark is a bit flawed since it exercises the network stack and it is probably affected by the OS scheduler, network stack etc. For example, in the browser the network messages probably go throw the network service resulting in some IPC roundtrips. I guess to measure the impact correctly we would need to mock the network stack and dispatch client events in some other way. 

Also, I think I mentioned this but this sort of logic for event dispatch is only-speced for the client and the server side does not need it as there is no need for interoperability.

lpinca:
> We are not able to replicate the regression on a Linux bare metal machine. I think the benchmark is a bit flawed since it exercises the network stack and it is probably affected by the OS scheduler, network stack etc. For example, in the browser the network messages probably go throw the network service resulting in some IPC roundtrips. I guess to measure the impact correctly we would need to mock the network stack and dispatch client events in some other way.

It should exercises the network stack, otherwise a reliable benchmark is this one https://github.com/websockets/ws/pull/2218#issuecomment-2050331136 🥲 

> Also, I think I mentioned this but this sort of logic for event dispatch is only-speced for the client and the server side does not need it as there is no need for interoperability.

Yes but there is only one `WebSocket` class in `ws`. It is shared by the client and the server.

lpinca:
Many small messages (30 bytes) in the same data chunk

```js
// server.js

const { createHash } = require('crypto');
const { createServer } = require('http');
const { Readable } = require('stream');
const { Sender } = require('ws');

const buf = Buffer.alloc(30);
const frame = Buffer.concat(Sender.frame(buf, { fin: true, opcode: 2 }));
const data = Buffer.concat(new Array(512).fill(frame));

const readable = new Readable({
  read() {
    this.push(data);
  }
});

  const key = createHash('sha1')
    .digest('base64');

  socket.write(
    [
      'Upgrade: websocket',
      'Connection: Upgrade',
      '\r\n'
    ].join('\r\n')
  );

  readable.pipe(socket);
});

  console.log('Server listening on [::]:8080');
});
```

```js
// client.js

let messages = 0;

const ws = new WebSocket('ws://127.0.0.1:8080');

ws.on('open', function () {
  console.time('ws');
});

ws.on('message', function () {
  if (++messages === 10_000_000) {
    console.timeEnd('ws');

    ws.terminate();
  }
});
```

```
# WSL 2, websockets/ws#master
node client.js
ws: 3.461s
```

```
# WSL 2, websockets/ws#fix/issue-2216
node client.js
ws: 11.426s
```

lpinca:
@OrKoN any feedback/suggestions? FWIW, I noticed that both Bun and Deno leak memory when running the benchmark above (https://github.com/websockets/ws/pull/2218#issuecomment-2056889841) without a limit on the number of messages on the client.

OrKoN:
@lpinca sorry, for the delay, requiring an option to opt-in into more spec-aligned behaviour would work for us, thanks!

PR #2228, opened 2024-06-11
doc: which events are proxied through `createWebSocketStream`

coolaj86:
The documentation doesn't specify how the `ws` module stream maps to which type of Node Duplex Stream.

I assume that it behaves like a TCP Duplex Stream: <https://nodejs.org/api/net.html#class-netsocket>.

I also assume that the ping and pong and keepalive timeouts are proxied in a meaningful way.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -625,7 +625,14 @@ The URL of the WebSocket server. Server clients don't have this attribute.
   constructor.
 
 Returns a `Duplex` stream that allows to use the Node.js streams API on top of a
-given `WebSocket`.
+given `WebSocket` which emits the following events:
+
+- `end`
+- `close`
+- `error`
+- `data`
+- `drain`
+- `timeout`
 
 ## Environment variables
 
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -630,6 +630,7 @@ given `WebSocket` which emits the following events:
 - `end`
 - `close`
 - `error`
+- `readable`
 - `data`
 - `drain`
 - `timeout`


lpinca:
> I assume that it behaves like a TCP Duplex Stream

Yes, https://nodejs.org/api/stream.html#class-streamduplex specifically.

> I also assume that the ping and pong and keepalive timeouts are proxied in a meaningful way.

You can't send and receive pings/pongs from the wrapper `Duplex`, but they are correctly handled by the wrapped `WebSocket`.

Do we have to list the events? They are the standard `Duplex` events.

coolaj86:
> Is it necessary to list the events? They are the standard Duplex events

I don't think that there are a standard set of events. I think it's all implementation specific - such as the use of 'end' vs 'close' and such. I think Duplexes are basically two-way EventEmitters, and can handle any documented event.

If it's listed, it's clear what this implementation handles. If it's not listed, someone has to guess or read the code.

lpinca:
> I don't think that there are a standard set of events. I think it's all implementation specific - such as the use of 'end' vs 'close' and such. I think Duplexes are basically two-way EventEmitters, and can handle any arbitrary (hopefully documented) event.

Yes there is standard set of events. See https://nodejs.org/api/stream.html. For a `Duplex` the set is events of `Readable` ∪ events of `Writable`. `'end'` and `'close'` have a different meaning and they are used accordingly. There no implementation specific events or they would have already be listed.

> If it's not listed, someone has to guess or read the code.

They just have to read Node.js documentation where the events are listed.

lpinca:
I'm closing this as per https://github.com/websockets/ws/pull/2228#issuecomment-2161283387 and https://github.com/websockets/ws/pull/2228#issuecomment-2196964912. Thank you anyway.

PR #2229, opened 2024-06-11
Add support for `Blob`

### LINKED ISSUE #2206: Support sending Blob

### Description

Currently, `ws` doesn't accept `Blob` as an argument to the `.send()` function:

```js
wss.on('connection', (ws) => ws.send(new Blob(['hello']))
```

I believe this is mainly for historic reasons because there used to be no `Blob` in Node.js. There is now, and it would be great to align the sending (and receiving also) with the WHATWG WebSocket standard. 

### ws version

8.16.0

### Node.js Version

v20.11.0

### System

Irrelevant.

### Expected result

`ws` supports sending `Blob` from the server.

### Actual result

- Type error on the unknown input type to `.send()`;
- A runtime `TypeError`:

```
TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Blob
```

### Attachments

_No response_

lpinca:
Closes #2206

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -14,6 +14,8 @@ const { URL } = require('url');
 const PerMessageDeflate = require('./permessage-deflate');
 const Receiver = require('./receiver');
 const Sender = require('./sender');
+const { isBlob } = require('./validation');
+
 const {
   BINARY_TYPES,
   EMPTY_BUFFER,
@@ -58,6 +60,7 @@ class WebSocket extends EventEmitter {
     this._closeFrameSent = false;
     this._closeMessage = EMPTY_BUFFER;
     this._closeTimer = null;
+    this._errorEmitted = false;
     this._extensions = {};
     this._paused = false;
     this._protocol = '';
@@ -90,9 +93,8 @@ class WebSocket extends EventEmitter {
   }
 
   /**
-   * This deviates from the WHATWG interface since ws doesn't support the
-   * required default "blob" type (instead we define a custom "nodebuffer"
-   * type).
+   * For historical reasons, the custom "nodebuffer" type is used by the default
+   * instead of "blob".
    *
    * @type {String}
    */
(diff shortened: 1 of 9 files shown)


lpinca:
@kettanaito I'm still not very happy with this because, other than compatibility with the WHAWG spec, it is pretty much useless. Anyway, here it is. I also didn't run any benchmarks to see the impact on performance.

lib/websocket.js
```diff
@@ -1030,6 +1031,11 @@ function initAsClient(websocket, address, protocols, options) {
  */
 function emitErrorAndClose(websocket, err) {
   websocket._readyState = WebSocket.CLOSING;
+  //
+  // The following assignment is practically useless and is done only for
```
kettanaito, line 1035: Which summarizes the entire change pretty flawlessly. I am still grateful for this consistency! Thank you!

kettanaito (approved):

dharishbabu007 (approved):

lpinca:
@titanism are you are setting [`websocket.binaryType`](https://github.com/websockets/ws/blob/master/doc/ws.md#websocketbinarytype) to `'blob'`?

5twelve (approved):

lib/websocket.js
```diff
@@ -1188,7 +1194,10 @@ function receiverOnError(err) {
     websocket.close(err[kStatusCode]);
   }
 
-  websocket.emit('error', err);
+  if (!websocket._errorEmitted) {
+    websocket._errorEmitted = true;
+    websocket.emit('error', err);
```
nikwen, line 1199: This seems worthwhile to pull out into a small function:

```js
function emitError(err) {
  websocket._errorEmitted = true;
  websocket.emit('error', err);
}
```

Seems like these lines should always be called together. Pulling them out into a function will make it less likely that someone will forget to do so in the future.

nikwen (commented):
Was just looking at the commit to make sure that upgrading `ws` won't break my app. Looks great overall. Just had a small suggestion. :)

matrix-compute (approved):

Issue #2241, opened 2024-07-06
[Feature]: Improve Project's JSDoc Types


ITenthusiasm:
**EDIT**: This will now only be an improvement to the project's JSDoc types. (See discussion below.)

### Is there an existing issue for this?

- [X] I've searched for any related issues and avoided creating a duplicate issue.

### Description

From a developer standpoint, it's a little inconvenient that developers have to install `ws` and `@types/ws` separately. From a maintainability standpoint, it's more work to manage both the library's code and the library's types separately. Doing so also opens up the door for types to accidentally get out of sync.

I've noticed that this library is so important that even [members of the TypeScript team](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57234) have contributed to the package's type definitions at certain points in time. (Though I don't know how motivated they are to do this in the present.) Having the types and the library code in a central place would provide a great DX.

For clarity, I saw #1623 and #1762. This is _not_ a request to rewrite the project to TypeScript. Instead, it's a request to _reuse_ the JSDoc approach that the library is already leveraging -- just like [Rich Harris suggests](https://www.youtube.com/watch?v=s234IO3RcZE) (and has suggested in many [other videos](https://www.youtube.com/live/MJHO6FSioPI?t=95s)). Some of the existing JSDocs might need a _little_ improvement, but I'm not assuming this would require any kind of overhaul.

**_I'm happy to contribute to this if the team is open to the idea._**

### The Approach

If this is something that's of interest, then the approach should be pretty simple. It's pretty much just creating a `tsconfig` that can check JSDocs. For example:

```json
{
  "compilerOptions": {
    "target": "ES2020",
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "strict": true,
    "allowJs": true,
    "checkJs": true,
    "declaration": true,
    "emitDeclarationOnly": true,
    "forceConsistentCasingInFileNames": true
  },
  "include": ["./lib/**/*.js"]
}
```

Then add a new step to the package _release_ process. (**There is no build step.**) Basically, this step would tell TypeScript to generate the `.d.ts` files from the JS files (`tsc`), and it would include those generated files in the released package. That's pretty much it. Since you won't be managing the `.d.ts` files yourself, they should probably be added to the `.gitignore`.

[...]

PR #2242, opened 2024-07-13
Improve JSDoc Types (Part 1)

ITenthusiasm:
## Description

This is an initial amount of work for improving the JSDocs in the codebase (as mentioned in #2241).

---

I figured it would make more sense to apply the JSDoc improvements incrementally. This has a few benefits:
- PRs end up being smaller to review (compared to if everything was done at once)
- Maintainers can more easily determine if #2241 is worth pursuing (in case they want to cancel everything or only accept fragments of the work)
- Contributors can make progress more asynchronously (in case there are gaps in free time)

Note: If the preference is only to merge everything at once, then this PR can be kept in a Draft State until it is "sufficiently complete".

## Additional Notes

When using JSDocs, TypeScript is fine with using `String` or `string`, `*` or `any`, and the like interchangeably. Not as idiomatic, but I'm not focused on that at the moment.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,5 +1,7 @@
+/* eslint no-unused-vars: ["error", { "varsIgnorePattern": "^WebSocket$" }] */
 'use strict';
 
+const WebSocket = require('./websocket');
 const { Duplex } = require('stream');
 
 /**
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -551,7 +551,7 @@ class Sender {
   /**
    * Sends a frame.
    *
-   * @param {Buffer[]} list The frame to send
+   * @param {(Buffer | String)[]} list The frame to send
    * @param {Function} [cb] Callback
    * @private
    */


lib/event-target.js
```diff
@@ -123,7 +123,7 @@ class ErrorEvent extends Event {
   }
 
   /**
-   * @type {*}
+   * @type {Required<NonNullable<ConstructorParameters<typeof ErrorEvent>[1]>>["error"]}
```
ITenthusiasm, line 126: This can just be the direct type, but it would mean that `error` would not automatically be kept in sync with the constructor argument. Same for the `data` property in `MessageEvent`.

lib/buffer-util.js
```diff
@@ -103,6 +105,8 @@ function toBuffer(data) {
   return buf;
 }
 
+toBuffer.readOnly = false;
```
ITenthusiasm, line 108: Don't necessarily need to keep this, but it does make life a little easier.

lib/event-target.js
ITenthusiasm: I guess another option is to try to type the `error` property itself, and then make the constructor parameter `ErrorEvent["error"] | null` or something like that. (Again, same for `MessageEvent`.)

lpinca:
Before you go any further, JSDoc comment should be kept as valid JSDoc, no TypeScript specific syntax.

ITenthusiasm:
Noted. I'll try to see what I can do.

lib/event-target.js
```diff
@@ -149,7 +153,7 @@ class MessageEvent extends Event {
    * @param {String} type The name of the event
    * @param {Object} [options] A dictionary object that allows for setting
    *     attributes via object members of the same name
-   * @param {*} [options.data=null] The message content
+   * @param {ArrayBuffer|Blob|Buffer|Buffer[]} [options.data=null] The message content
```
lpinca, line 156: It can actually be anything. When used in isolation

```js
const event = new MessageEvent('message', { data: NaN });
```

is perfectly valid. The same applies to `ErrorEvent`.

ITenthusiasm:
For clarity, it seems like [JSDoc supports Google Closure Compiler type expressions](https://jsdoc.app/tags-type). Are we also interested in having those (e.g., generics via `@template`) here? Or no?

(I'm just trying to figure out what to remove/filter out, e.g., `ConstructorParameters`.)

lpinca:
I don't think `@template` exists in JSDoc.

lib/event-target.js
ITenthusiasm: Just to be sure, is the expectation/desire for this to be used in isolation?

lpinca: It is not, but it is a standalone module that can be used in isolation, so functions/features are documented as such.

lib/buffer-util.js
ITenthusiasm: Outdated, so resolving.

lib/event-target.js
ITenthusiasm: Got it. I will remove the commit, then.

ITenthusiasm:
Okay. The PR is a little more anticlimactic, but I _think_ that all of the TypeScript syntax should be gone now.

lib/event-target.js
ITenthusiasm: Resolving. See https://github.com/websockets/ws/pull/2242#discussion_r1676920333

lib/stream.js
```diff
@@ -41,8 +41,8 @@ function duplexOnError(err) {
 /**
  * Wraps a `WebSocket` in a duplex stream.
  *
- * @param {WebSocket} ws The `WebSocket` to wrap
- * @param {Object} [options] The options for the `Duplex` constructor
+ * @param {import("./websocket")} ws The `WebSocket` to wrap
+ * @param {import("node:stream").DuplexOptions} [options] The options for the `Duplex` constructor
```
lpinca, line 45: I think this is TypeScript-specific.

ITenthusiasm: Strange. I thought that this was. normal in JSDoc, but I can remove this as well.

ITenthusiasm:
The update to `sendFrame` is all that's left. Pretty simple 👍🏾

lpinca (approved):

lib/stream.js
lpinca: For the `import("./websocket")` case we can do this:

```diff
diff --git a/lib/stream.js b/lib/stream.js
index 230734b..1eaf7b6 100644
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,5 +1,8 @@
+/* eslint no-unused-vars: ["error", { "varsIgnorePattern": "^WebSocket$" }] */
+
 'use strict';
 
+const WebSocket = require('./websocket');
 const { Duplex } = require('stream');
 
 /**

```

ITenthusiasm: Added.

ITenthusiasm:
Made the suggested change for `createWebSocketStream`. Can remove if desired.

PR #2244, opened 2024-07-23
Add a space between subprotocols in Sec-WebSocket-Protocol header

orgads:
Some servers fail to parse this value when passed without a space[1].

Both websocket package[2] and Node undici[3] have a space.

[1] https://github.com/deepgram/deepgram-js-sdk/issues/321
[2] https://github.com/theturtle32/WebSocket-Node/blob/d87afb7ef2/lib/WebSocketClient.js#L200
[3] https://github.com/nodejs/undici/blob/53350135c8/lib/web/fetch/headers.js#L176

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -774,7 +774,7 @@ function initAsClient(websocket, address, protocols, options) {
       protocolSet.add(protocol);
     }
 
-    opts.headers['Sec-WebSocket-Protocol'] = protocols.join(',');
+    opts.headers['Sec-WebSocket-Protocol'] = protocols.join(', ');
   }
   if (opts.origin) {
     if (opts.protocolVersion < 13) {


lpinca:
> Some servers fail to parse this value when passed without a space

It should be fixed on the server.

```
The request MAY include a header field with the name
|Sec-WebSocket-Protocol|.  If present, this value indicates one
or more comma-separated subprotocol the client wishes to speak,
ordered by preference.  The elements that comprise this value
MUST be non-empty strings with characters in the range U+0021 to
U+007E not including separator characters as defined in
[RFC2616] and MUST all be unique strings.  The ABNF for the
value of this header field is 1#token, where the definitions of
constructs and rules are as given in [RFC2616].
```

```
#rule
   A construct "#" is defined, similar to "*", for defining lists of
   elements. The full form is "<n>#<m>element" indicating at least
   <n> and at most <m> elements, each separated by one or more commas
   (",") and OPTIONAL linear white space (LWS). This makes the usual
   form of lists very easy; a rule such as
      ( *LWS element *( *LWS "," *LWS element ))
   can be shown as
      1#element
   Wherever this construct is used, null elements are allowed, but do
   not contribute to the count of elements present. That is,
   "(element), , (element) " is permitted, but counts as only two
   elements. Therefore, where at least one element is required, at
   least one non-null element MUST be present. Default values are 0
   and infinity so that "#element" allows any number, including zero;
   "1#element" requires at least one; and "1#2element" allows one or
   two.
```

The space is optional.

orgads:
I agree the server is bad, but since it is common to have a space, other widespread libraries do it on the client side, and there are servers that misbehave, do you see a reason not to add it?

lpinca:
do you see a reason not to add it?

Broken servers will remain broken.

lukeocodes:
> > do you see a reason not to add it?
> 
> Broken servers will remain broken.

This is a rather poor answer. You acknowledge either way works, then you affectively acknowledge it should be a setting in the library. 

Never mind that built-in examples in browsers and the W3C specs use spacing.

lpinca:
My point is that if all clients have the same behavior and the issue is not fixed on the server, the issue will be never be properly fixed.

lukeocodes:
> My point is that if all clients have the same behavior and the issue is not fixed on the server, the issue will be never be properly fixed. What would you do for the next client that does not add a space?

In earlier versions, `ws` conformed with the default for other clients, as we have literally never had another client-related issue. 

In v8, you appear to have modified the string you send from the one you had previously. I consider this a regression between major releases. Until I see a pattern of other clients (browsers, native, etc) start dropping the space, you're the outlier. You're not flexible as the spec is, you're just fixed in one solution, and opposed to other clients in the wild.

I'm sure you'll agree, in a high-speed service that handles many millions of requests, having to parse a space out of a string is an overhead servers shouldn't _need_ to absorb, and libraries should be flexible and configurable.

https://github.com/websockets/ws/commit/0aecf0c95b41eaeeec9e265d5f1cb6c96e2f4d08#diff-44c2c0628f273ea1ab6dbb7c1fa81a1765d8adf9fbdd083e2a3b93d8d1cb5f61L65

orgads:
@lpinca and @lukeocodes Thanks for your prompt responses and cooperation. I'm closing this PR.

PR #2250, opened 2024-08-20
Update README.md

metric152:
updated the readme to show how to retrieve the string data from the message response

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,8 +172,8 @@ ws.on('open', function open() {
   ws.send('something');
 });
 
-ws.on('message', function message(data) {
-  console.log('received: %s', data);
+ws.on('message', function message(data, isBinary) {
+  console.log('received: %s', isBinary ? data : data.toString());
 });
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -172,8 +172,8 @@ ws.on('open', function open() {
   ws.send('something');
 });
 
-ws.on('message', function message(data) {
-  console.log('received: %s', data);
+ws.on('message', function message(data, isBinary) {
+  console.log('received: %s', isBinary ? data : data.toString());
```
lpinca, line 176: This is not needed `data.toString()` is called automatically.

lpinca (changes requested):

README.md
metric152: When I tried it it returned “Buffer <“ and a bunch of numbers. Maybe it’s the version of node I’m using : 16. I had to use “toString” to access the data.

lpinca: It's not the Node.js version. See the string substitution (`%s`) in the message.

https://runkit.com/5ea585d2c57dcf0013d0bdc2/66c4fb24dc86270008e51a66

OKEAMAH (approved):
Good

lpinca:
Thank you anyway.

PR #2262, opened 2024-11-06
First message lost after connectiong to the server, race condition.

vedun:
Hi. I encountered a problem. When the server sends the first message after establishing a connection, the client does not always receive it. As far as I understand, this happens because after creating an instance of the WebSocket object, the process of connecting to the server immediately begins and does not always have time to subscribe to the message event. I created a small fix for this situation while maintaining backward compatibility. I would like you to look at it and comment and if everything is okay, accept it.
P.S.
I haven't figured out yet how to create a race condition in the test.

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -325,6 +325,9 @@ This class represents a WebSocket. It extends the `EventEmitter`.
   - `skipUTF8Validation` {Boolean} Specifies whether or not to skip UTF-8
     validation for text and close messages. Defaults to `false`. Set to `true`
     only if the server is trusted.
+  - `connectImmediately` {Boolean} If set to `false` prevent connecting to the
+    server in constructor. Later you should use `connect` method to establish
+    the connection. Defaults set to `true`.
   - Any other option allowed in [`http.request()`][] or [`https.request()`][].
     Options given do not have any effect if parsed from the URL given with the
     `address` parameter.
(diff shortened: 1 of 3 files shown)


lpinca:
> When the server sends the first message after establishing a connection, the client does not always receive it. As far as I understand, this happens because after creating an instance of the WebSocket object, the process of connecting to the server immediately begins and does not always have time to subscribe to the message event.

The `'message'` event depends on the `'data'` event of the underlying `net.Socket` and that event is emitted _after_ (not in the same tick) the WebSocket instance is created, so it not possible to miss `'data'`/`'message'` events.

vedun:
> The `'message'` event depends on the `'data'` event of the underlying `net.Socket` and that event is emitted _after_ (not in the same tick) the WebSocket instance is created, so it not possible to miss `'data'`/`'message'` events.

It is posible if betveen subsribe to "message" and creation of the WebSocket object we have some async operations (like connecting to DB and other services). WebSocket is part of a larger system and the initialization order in it cannot always be changed, unfortunately.

Now I use another library for the client where I can postpone the connection establishment. If you think this design approach is bad then you can close this PR. I thought maybe this would be useful to someone else.

lpinca:
> It is posible if betveen subsribe to "message" and creation of the WebSocket object we have some async operations (like connecting to DB and other services). WebSocket is part of a larger system and the initialization order in it cannot always be changed, unfortunately.

It is not a `ws` issue. Adding yet another non standard method is not something I want. Thank you anyway.

lpinca:
FWIW, if a system that uses a WebSocket implementation with the WHATWG API prevents you from immediately adding a listener for the `'message'` event, that system is poorly designed. If for some reason this is unavoidable, that system should buffer messages and return them to the users when they have the opportunity to listen to the messages themselves.

PR #2280, opened 2025-03-31
fix(Sender): Clear RSV2/RSV3 bits on outgoing WebSocket frames

romanouille:
This PR fixes a protocol violation where outgoing frames could include RSV2 or RSV3 bits set, even though no extension negotiates them.

According to RFC 6455 §5.2, RSV2 and RSV3 must be 0 unless an extension specifically uses them.

This change ensures the bits are cleared before frames are sent, restoring compatibility with strict WebSocket clients.

Also includes a warning if such bits are requested by mistake.

### DIFF
--- a/lib/sender.js
+++ b/lib/sender.js
@@ -140,7 +140,7 @@ class Sender {
 
     target[0] = options.fin ? options.opcode | 0x80 : options.opcode;
     if (options.rsv1) target[0] |= 0x40;
-
+    target[0] &= ~0x30;
     target[1] = payloadLength;
 
     if (payloadLength === 126) {


lib/sender.js
Junjun13910, line 1: Hello everyone

Junjun13910: 7ababb6ca78c30e328835a938a764edefa82027c

Junjun13910 (commented):
7ababb6ca78c30e328835a938a764edefa82027c

lib/sender.js
Junjun13910, line 1: `target

Junjun13910 (approved):

lpinca:
Why? Those bits are never set unless you use un invalid opcode in `Sender.frame()` which is a valid use case for testing.

lpinca (changes requested):
https://github.com/websockets/ws/pull/2280#issuecomment-2770212995

lpinca:
> I’d like to highlight why this PR is not only valid, but essential to ensure strict compliance with the WebSocket protocol (RFC 6455), particularly section 5.2, which governs the interpretation and transmission of control bits in the frame header.

> According to the spec:

> “If the negotiated extension defines the meaning of RSV1, RSV2, or RSV3, then the bits MUST be set appropriately. Otherwise, they MUST be 0.”

Yes, the RSV2 and RSV3 are never set via the public API. There is strict compliance.

> A developer may incorrectly or unintentionally pass rsv2: true in a call to a lower-level internal function (e.g. via ws._sender.sendFrame).

That is not a public API, you can also use `ws._socket.write()` and send garbage.

lpinca:
> When rsv2: true is set manually, the ws module currently emits a byte with the RSV2 bit set in the frame's first byte (e.g., 0b00100001 instead of 0b00000001 for a simple text frame). After applying the patch from this PR, the rsv2 and rsv3 bits are forcibly cleared, and the output byte conforms to spec regardless of what upstream code attempted.

What is `rsv2: true`?

Junjun13910:
Hello goo morning. Thank you for update just wanted  to let everyone know that interesting the money. The only thing I want  to get is the money I deposit in crypto before. And I provide everything they need. Till  Mr  GREG DAWSON  playing at me they get the money from crypto.com put it back in base. I open account in coinbase then coinbase said they takecare of it  till coinbase said they can’t do it anymore.  Both coinbase and crypto.com so they are not safe. How  did they transfer? And they hold my account. Just wanted to know MR greg dawson  that im not interested in that money. I grader working 2 job to survive my family. Than the money come from people get scum. Money is just there around. We just need to work hard to get. Whine we pass away we can’t bring all money anyway. we are all going there.  That money was there you can have it to make you more happy.  I wish you all the best and you happiness 

thank you

lpinca:
The point is that if a developer uses `Sender.frame()` directly to create frames, it is their responsibility to use it correctly. It is their responsibility to

1. not set the RSV1 bit if the permessage-deflate extension was not negotiated
1. not set the RSV1 bit if the opcode is 0
1. not set the opcode to 0 if the frame is not a fragment
1. not set the opcode to 1 or 2 if the frame is not the first fragment
1. set the FIN bit if the frame is a control frame
1. not set the RSV1 bit in a control frame
1. use a **valid** opcode. Using an invalid opcode is currently the only way to set the RSV2 and RSV3 bits via `Sender.frame()`. Clearing the RSV2 and RSV3 bits does not automatically make the opcode valid.
1. etc.

PR #2290, opened 2025-06-27
[feature] Introduce the `httpHandler`, `ssl`, `key`, and `cert` options and `WebSocketServer#listen()` method

richytong:
This adds support for the `httpHandler`, `ssl`, `key`, and `cert` options and `WebSocketServer#listen()` method to the `WebSocketServer` constructor.

With `httpHandler`, users can overwrite the existing default http handler that only responds with `426 Upgrade Required`. This is useful for implementing health checks that require a 200 status code.

```js
const httpHandler = (request, response) => {
  response.writeHead(200, {
    'Content-Type': 'text/plain'
  });
  response.end('OK');
};

const wss = new WebSocket.Server({ httpHandler, port: 8080 });
```

With the `ssl`, `key`, and `cert` options, users can create an HTTPS server over the default HTTP server. This allows for a simple API to create an HTTPS server using `ws`.

```js
  ssl: true,
  cert: fs.readFileSync('/path/to/my/certificate.pem'),
  key: fs.readFileSync('/path/to/my/fixtures/key.pem'),
  port: 443
});
```

With the `WebSocketServer#listen()` method, users can create the WebSocket server and have it listen at a later time. This API is more conformant to NodeJS [http.Server](https://nodejs.org/api/http.html#serverlisten) and [https.Server](https://nodejs.org/api/https.html#serverlisten).

```js
const httpHandler = (request, response) => {
  response.writeHead(200, {
    'Content-Type': 'text/plain'
  });
  response.end('OK');
};

const wss = new WebSocket.Server({ httpHandler });

wss.listen(8080)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,32 @@ wss.on('connection', function connection(ws) {
 server.listen(8080);
 ```
 
+### Internal HTTP/S Server
+
+```js
+import { readFileSync } from 'fs';
+import { WebSocketServer } from 'ws';
+
+const wss = new WebSocketServer({
+  ssl: true,
+  cert: readFileSync('/path/to/cert.pem'),
+  key: readFileSync('/path/to/key.pem')
+});
+
+wss.on('connection', function connection(ws) {
+  ws.on('error', console.error);
+
+  ws.on('message', function message(data) {
+    console.log('received: %s', data);
+  });
+
+  ws.send('something');
+});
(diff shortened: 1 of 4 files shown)


lpinca:
Thanks but no. Something like this was already proposed and rejected in the past. Use an external server for flexibility.

PR #2291, opened 2025-06-28
feat: add `sec-websocket-version` to response headers when request `sec-websocket-version` is invalid

samuel871211:
While test locally with postman, send invalid `sec-websocket-version`, I found that the server doesn't include `sec-websocket-version: 13,8` in response headers
no-sec-websocket-version-header-present

1. Based on MDN Document
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Version

```
If the server doesn't support the version, or any header in the handshake is not understood or has an incorrect value, the server should send a response with status [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400) and immediately close the socket. It should also include Sec-WebSocket-Version in the 400 response, listing the versions that it does support.
```

2. Based on RFC6455
https://datatracker.ietf.org/doc/html/rfc6455#section-4.4

```
The following example demonstrates version negotiation described above:

      GET /chat HTTP/1.1
      Host: server.example.com
      Upgrade: websocket
      Connection: Upgrade
      ...
      Sec-WebSocket-Version: 25

   The response from the server might look as follows:

      HTTP/1.1 400 Bad Request
      ...
      Sec-WebSocket-Version: 13, 8, 7

   Note that the last response from the server might also look like:

      HTTP/1.1 400 Bad Request
      ...
      Sec-WebSocket-Version: 13
      Sec-WebSocket-Version: 8, 7

   The client now repeats the handshake that conforms to version 13:

      GET /chat HTTP/1.1
      Host: server.example.com
      Upgrade: websocket
      Connection: Upgrade
      ...
      Sec-WebSocket-Version: 13
```

This PR add `sec-websocket-version` to response headers when request `sec-websocket-version` is invalid.

- [x] npm run lint
- [x] npm run test
- [x] npm run integration

### DIFF
--- a/lib/websocket-server.js
+++ b/lib/websocket-server.js
@@ -11,7 +11,11 @@ const extension = require('./extension');
 const PerMessageDeflate = require('./permessage-deflate');
 const subprotocol = require('./subprotocol');
 const WebSocket = require('./websocket');
-const { GUID, kWebSocket } = require('./constants');
+const {
+  GUID,
+  kWebSocket,
+  WEBSOCKET_SUPPORT_VERSIONS_DESC
+} = require('./constants');
 
 const keyRegex = /^[+/0-9A-Za-z]{22}==$/;
 
@@ -256,9 +260,11 @@ class WebSocketServer extends EventEmitter {
       return;
     }
 
-    if (version !== 8 && version !== 13) {
+    if (!WEBSOCKET_SUPPORT_VERSIONS_DESC.includes(version)) {
       const message = 'Missing or invalid Sec-WebSocket-Version header';
-      abortHandshakeOrEmitwsClientError(this, req, socket, 400, message);
+      abortHandshakeOrEmitwsClientError(this, req, socket, 400, message, {
+        'Sec-WebSocket-Version': WEBSOCKET_SUPPORT_VERSIONS_DESC.toString()
+      });
       return;
     }
 
@@ -526,15 +532,23 @@ function abortHandshake(socket, code, message, headers) {
  * @param {Duplex} socket The socket of the upgrade request
  * @param {Number} code The HTTP response status code
  * @param {String} message The HTTP response body
+ * @param {Object} [headers] The HTTP response headers
  * @private
  */
-function abortHandshakeOrEmitwsClientError(server, req, socket, code, message) {
+function abortHandshakeOrEmitwsClientError(
+  server,
+  req,
+  socket,
+  code,
+  message,
+  headers
+) {
   if (server.listenerCount('wsClientError')) {
     const err = new Error(message);
     Error.captureStackTrace(err, abortHandshakeOrEmitwsClientError);
 
     server.emit('wsClientError', err, socket, req);
   } else {
-    abortHandshake(socket, code, message);
+    abortHandshake(socket, code, message, headers);
   }
 }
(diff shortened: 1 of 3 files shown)


lpinca (approved):

lpinca:
FWIW, given that

```
The request MUST include a header field with the name
|Sec-WebSocket-Version|.  The value of this header field MUST be
13.
```

I think it makes sense to drop support for version 8, but it is a breaking change.

lpinca:
Thank you.

lpinca:
Maybe, but it is not a priority. Supporting version 8 does not cost anything.

```
git grep '=== 8'
lib/websocket-server.js:          req.headers[`${version === 8 ? 'sec-websocket-origin' : 'origin'}`],
```

lpinca:
The only valid reason is strict compliance with RFC 6455, but it does not worth a major version bump per se. It should be added to the list of breaking changes for the next major version.

Edgarruiz8585:
> While test locally with postman, send invalid `sec-websocket-version`, I found that the server doesn't include `sec-websocket-version: 13,8` in response headers
> no-sec-websocket-version-header-present
> 
> 1. Based on MDN Document
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Version
> 
> ```
> If the server doesn't support the version, or any header in the handshake is not understood or has an incorrect value, the server should send a response with status [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400) and immediately close the socket. It should also include Sec-WebSocket-Version in the 400 response, listing the versions that it does support.
> ```
> 
> 2. Based on RFC6455
> https://datatracker.ietf.org/doc/html/rfc6455#section-4.4
> 
> ```
> The following example demonstrates version negotiation described above:
> 
>       GET /chat HTTP/1.1
>       Host: server.example.com
>       Upgrade: websocket
>       Connection: Upgrade
>       ...
>       Sec-WebSocket-Version: 25
> 
>    The response from the server might look as follows:
> 
>       HTTP/1.1 400 Bad Request
>       ...
>       Sec-WebSocket-Version: 13, 8, 7
> 
>    Note that the last response from the server might also look like:
> 
>       HTTP/1.1 400 Bad Request
>       ...
>       Sec-WebSocket-Version: 13
>       Sec-WebSocket-Version: 8, 7
> 
>    The client now repeats the handshake that conforms to version 13:
> 
>       GET /chat HTTP/1.1
>       Host: server.https://GitHub.com/Edgarruiz8585 8080
>       Upgrade: websocket
>       Connection: Upgrade
>       ...
>       Sec-WebSocket-Version: 13
> ```
> 
> This PR add `sec-websocket-version` to response headers when request `sec-websocket-version` is invalid.
> Into Edgarruiz8585 
> - [x] npm run lint
> - [x] npm run test
> - [x] npm run integration

lib/websocket-server.js
```diff
@@ -256,9 +256,11 @@ class WebSocketServer extends EventEmitter {
       return;
     }
```
Edgarruiz8585, line 257: Edgar Manuel Ruiz Arias mr Ed

PR #2299, opened 2025-09-28
Update wrapper.mjs

lemanschik:
Fixed: esm warapper which exported nested default

Why this is Correct:
When an ESM file imports a CJS file that uses module.exports = SomeValue;, Node.js (and most transpilers/bundlers) treats that exported value as the default export.
To capture that value and give it a specific named binding (like createWebSocketStream) within the current ESM scope, you must use the explicit named import syntax: import { default as <Name> } from 'module';.
This ensures that the main exports of the underlying CJS files are correctly mapped into the local environment for re-exporting.

```js
export { createWebSocketStream, Receiver, Sender, WebSocket, WebSocketServer };
export default { createWebSocketStream, Receiver, Sender, WebSocket, WebSocketServer };
```

Why this is Correct:
The module is explicitly listing the imported bindings and exporting them as named exports. This is the standard, modern way to expose library functions and classes, ensuring consumers must use named imports: import { WebSocket } from 'ws';.

the ESM entry point (as determined by the exports map), must explicitly provide a binding named default to satisfy the syntax import WS from 'ws'.

### DIFF
--- a/wrapper.mjs
+++ b/wrapper.mjs
@@ -1,8 +1,12 @@
-import createWebSocketStream from './lib/stream.js';
-import Receiver from './lib/receiver.js';
-import Sender from './lib/sender.js';
-import WebSocket from './lib/websocket.js';
-import WebSocketServer from './lib/websocket-server.js';
+// Our SRC is CJS so we need to import default named
+import { default as createWebSocketStream } from './lib/stream.js';
+import { default as Receiver } from './lib/receiver.js';
+import { default as Sender } from './lib/sender.js';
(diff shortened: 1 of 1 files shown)


lemanschik:
@lpinca what do you mean by that ? do you want to verify the error?

npx node@24 my-file.js

```js
import ws from 'ws'
console.log(Object.keys(ws))
```

lemanschik:
@lpinca ok did verify nodejs indeed added now SntacticDefault Import but bundlers like rollup would fail this fixes that

PR #2308, opened 2025-12-16
Add override for default closeTimeout.

anthony-michael-bennett:
Setting the option closeTimeout will override the default milliseconds for terminating unresponsive connections.


anthony-michael-bennett:
Sorry i only had a test on the client. Thank you for the updated test for client and server.  That works for me.

Issue #2331, opened 2026-07-06
Default maxFragments still allows memory-exhaustion DoS (incomplete GHSA-96hv-2xvq-fx4p fix)


geo-chen:
### Is there an existing issue for this?

- [x] I've searched for any related issues and avoided creating a duplicate issue.

### Description

reported via email on 4 June 2026 - no response:

The GHSA-96hv-2xvq-fx4p fix added maxFragments/maxBufferedChunks, but the defaults are high enough that a default-configured server can still be OOM-crashed by one unauthenticated peer.

In lib/receiver.js:509-531, the guard counts fragments and only fires at the limit. An attacker sends a text frame with FIN=0 plus maxFragments-2 one-byte continuation frames and never sends FIN. The message never completes, the fragment array stays just under maxFragments (default 131072) forever, _totalPayloadLength stays ~128 KB (far below maxPayload), and each 1-byte fragment is a separate Buffer view with ~100 bytes of overhead.

Validated on ws 8.21.0 (default options): 131000 fragments (~0.87 MB wire) pin ~13.5 MB of server heap (about 15x), guard never fires. A few dozen such connections (~40 MB total traffic) OOM-kill a default-heap Node process.

Suggested fix: lower the default maxFragments/maxBufferedChunks substantially, and/or add a per-connection (and ideally process-wide) cap on bytes pinned by incomplete messages rather than only a fragment count, so the per-fragment object overhead cannot be amplified.

lpinca:
1. The value is customizable.
2. Lowering the default value does not change much if the number of connections remains unlimited.
3. You can do the same with incomplete non-fragmented messages spread over multiple connections.

rafael-nogueras:
@lpinca  Hello! Although this fix (including all the back-ports) addresses [CVE-2026-62389](https://github.com/advisories/GHSA-73jw-fp74-p77x), the CVE fix record reports only 8.21.1 as the fix version.

What is the process to get the back-ported versions (specifically I am interested in 7.5.13) to also be considered fixes for the CVE? Thanks!

lpinca:
See https://github.com/websockets/ws/releases/tag/7.5.12, anyway I disagree with https://github.com/advisories/GHSA-73jw-fp74-p77x for the reasons stated above. It's a low effort duplicate of https://github.com/advisories/GHSA-96hv-2xvq-fx4p. I'm not sure who published that CVE. It's a shame no one reviewed it.

rafael-nogueras:
[...]

PR #2332, opened 2026-07-11
[fix] Fixing Default maxFragments still allows memory exhaustion Dos

codepawpaw:
Issue: https://github.com/websockets/ws/issues/2331

**Problem:**

The GHSA-96hv-2xvq-fx4p fix added the count-based guards maxFragments and maxBufferedChunks, but their defaults are high enough that a default-configured server can still be OOM-crashed by a single unauthenticated peer.

Both guards count parts, and the only byte-based guard (maxPayload) counts payload bytes only. Nothing accounts for the per-Buffer-view object overhead (~128 B each). An attacker can exploit this:

Send a text frame with FIN=0, then maxFragments - 2 one-byte continuation frames, and never send FIN.
The message never completes, so the fragment array is retained indefinitely.
_totalPayloadLength stays ~128 KB (far below maxPayload of 100 MB) and the fragment count stays just under maxFragments (131072), so no guard ever fires.
Reproduced on the current tree (default server options): ~0.87 MB of wire traffic pins ~13–16 MB of heap (≈15× amplification). A few dozen such connections OOM-kill a default-heap Node process.

**Fix**
Two complementary changes:

1. New maxBufferedBytes option (structural fix). Charges a fixed per-part overhead (BUFFER_OVERHEAD = 128) on top of payload bytes, so the limit bounds the memory actually pinned rather than only the payload size. Enforced on all three retention paths — buffered input chunks (_write), uncompressed fragments (getData), and compressed fragments (decompress). Exceeding it emits WS_ERR_TOO_MANY_BUFFERED_PARTS (close code 1008).

2. Lowered the default count limits (defense-in-depth)
maxBufferedBytes defaults above maxPayload (100 MiB) so that a legitimate max-size message is never rejected; the tighter count defaults are what cap per-connection retained memory in the default configuration

New behavior is opt-in at the Receiver level (maxBufferedBytes defaults to 0 = disabled there). The lowered maxFragments/maxBufferedChunks defaults are a behavior change only for peers sending an unusually large number of fragments/chunks in a single message, all limits remain configurable (set to 0 to disable).

### DIFF
--- a/doc/ws.md
+++ b/doc/ws.md
@@ -87,11 +87,17 @@ This class represents a WebSocket server. It extends the `EventEmitter`.
   - `handleProtocols` {Function} A function which can be used to handle the
     WebSocket subprotocols. See description below.
   - `host` {String} The hostname where to bind the server.
+  - `maxBufferedBytes` {Number} The maximum estimated amount of memory, in
+    bytes, that may be pinned by buffered data chunks or the fragments of an
+    in-progress message. It is measured as the sum of the payload bytes plus a
+    fixed per-part overhead, so it bounds the memory actually retained rather
+    than only the payload size. The value is coerced to a 32-bit signed integer.
+    Defaults to 134217728 (128 MiB). Set to 0 to disable the limit.
   - `maxBufferedChunks` {Number} The maximum number of buffered data chunks. The
-    value is coerced to a 32-bit signed integer. Defaults to 1048576. Set to 0
-    to disable the limit.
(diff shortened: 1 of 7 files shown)


lpinca:
Thanks but this is not wanted. The overhead can change at any time without notice and the default value is customizable. A conservative value was chosen to not break applications.

> can still be OOM-crashed by a single unauthenticated peer

I doubt this is true. You can do it with multiple connection, but in that case the fix is limiting the number of connections.

